PR #5, opened 2016-05-28
add loading of bindings

fredrikaverpil:
This makes it possible to `import Qt` straight from Qt.py:

```
~/code/Qt
❯ python
Python 2.7.11 (default, Jan 22 2016, 08:29:18)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
>>> from Qt import QtWidgets
>>> print QtWidgets
<module 'PyQt5.QtWidgets' from '/usr/local/lib/python2.7/site-packages/PyQt5/QtWidgets.so'>
```

Please note: PR now targets the `master` branch.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1,2 +1,63 @@
-# Importing this file means importing the most desireable
-# bindings for Qt relative the running version of Python
+"""Importing this file means importing the most desireable bindings for
+Qt relative the running version of Python.
+"""
+
+import sys
+
+
+def load_pyqt5():
+    import PyQt5.Qt
+    sys.modules["Qt"] = PyQt5
+    PyQt5.QtCore.Signal = PyQt5.QtCore.pyqtSignal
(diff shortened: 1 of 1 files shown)


mottosso:
Brilliant!

mottosso:
Next I think we should have a look at how to clean up this code.

At the moment, I'm not a fan of:

1. `init()` being declared and immediately called, I might as well not be a function.
2. Commented out lines of code
3. Importing the `.Qt` submodule of `PyQt` means "import all" which might give a performance hit, when one might only end up using `QtWidgets` (?)
4. The growing `try/except` block; flat is better.

fredrikaverpil:
I couldn't agree more. How do you want to do this, should we create issues
for one and each of those?

mottosso:
Sure, let's do that.

Will you do the honors?

fredrikaverpil:
Sure, np.

Issue #9, opened 2016-05-28
The growing try/except block; flat is better


fredrikaverpil:
Agreed.

This is what we have now:

```python
"""Support Qt 4 and 5, PyQt and PySide"""
try:
    load_pyside2()
except ImportError:
    try:
        load_pyqt5()
    except ImportError:
        try:
            load_pyside()
        except ImportError:
            try:
                load_pyqt4()
            except:
                sys.stderr.write("Qt: Could not find "
                                 "appropriate bindings for Qt\n")
```

mottosso:
I'd like to make it more clear in which order things are imported, that's the main thing. At the moment, it's difficult to tell because it just looks bad.

fredrikaverpil:
I wanted to check whether the __binding__ attribute had been created in order to determine whether the next module should attempt to load or not, but then I noticed that `sys` is of `NoneType`:
https://github.com/mottosso/Qt/issues/15

PR #13, opened 2016-05-28
Removed init() and restructured code

fredrikaverpil:
What do you think of this kind of approach?

With this, we get a function (`load_wrapper`) which could handle errors and you can re-order `QT_BINDING_LOAD_ORDER` to change the load order.

Fixes https://github.com/mottosso/Qt/issues/6 and https://github.com/mottosso/Qt/issues/9

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -4,6 +4,8 @@ Qt relative the running version of Python.
 
 import sys
 
+QT_BINDING_LOAD_ORDER = ['PySide2', 'PyQt5', 'PySide', 'PyQt4']
+
 
 def load_pyqt5():
     """Load PyQt5 for Qt5 bindings"""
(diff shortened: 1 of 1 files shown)


mottosso:
Looks good, but I wonder whether it really makes it easier to see what is happening.

fredrikaverpil:
Would you rather see something like this?

```python
loaded = False
for binding in QT_BINDING_LOAD_ORDER:
    if not loaded:
        function_name = 'load_' + binding.lower() + '()'
        try:
            exec function_name
            loaded = True
        except ImportError:
            pass
if not loaded:
    sys.stderr.write("Qt: Could not find appropriate bindings for Qt\n")
```

mottosso:
I'm not a fan of:

1. Needless state, `loaded = False`
2. Guessing function names, `function_name = 'load_' + binding.lower() + '()'`
3. `exec`

(1) makes it difficult to read; having to continue reading before understanding what `loaded` does. (2) Means that the name of a function is stored twice; both in the function, and once in the list of orders. Complicates maintenance. Finally (3) is just no.

The current approach (#9) isn't pretty, but it is easy to understand and maintain. We should hold on to those properties.

mottosso:
Also bear in mind that every variable you declare at module-level is accessible via import.

```python
from Qt import loaded
# what is loaded?
```

fredrikaverpil:
> every variable you declare at module-level is accessible via import.

Ah, that's very true. I didn't think about that.

PR #16, opened 2016-05-28
Removed init() and restructured code

fredrikaverpil:
Not sure if I feel there's a great improvement here over the original... but at least something to discuss.


fredrikaverpil:
I'm revising this.

Issue #10, opened 2016-05-28
Error when no Qt binding is installed


fredrikaverpil:
There should be some kind of error when there's no Qt binding available for loading.

fredrikaverpil:
Oh, crap. It's already in there.

mottosso:
Yeah, it's a message to stderr, but it's worth discussing what kind of error it should produce.

In ordinary circumstances, an unavailable module would throw an `ImportError`. Maybe this should too?

fredrikaverpil:
Possibly, yes. We should be more specific on what we tried to do and what caused an error.

fredrikaverpil:
I think we should use `raise ImportError` instead:

    raise ImportError("Could not find appropriate bindings for Qt")

```python
>>> try:
...     import Qt
... except ImportError:
...     print 'oh no'
...
oh no
```

PR #17, opened 2016-05-29
Cosmetics and #10

mottosso:
Here's an alternative to #10 and #9.


fredrikaverpil:
Cool, that gets rid of those issues.

fredrikaverpil:
Just one question, don't we want to `import sys as _sys`?

Issue #22, opened 2016-05-30
Override preferred binding


mottosso:
### Goal

The available Qt bindings - PySide, PySide2, PyQt4 and PyQt5 - are automatically detected. In case two or more bindings are available and one if preferred, currently there is no way of specifying which to pick.

Enable some mechanism with which to specify up-front which bindings to use.

### Motivation

Primarily intended for use in testing, where tests must be able to specify bindings with which to tests against, but could potentially be useful in an enterprise environment where multiple bindings must be present on the `PYTHONPATH`.

### Implementation

Query the value of `QT_PREFERRED_BINDING` and throw an error is this binding doesn't exist.

fredrikaverpil:
Yes, very useful. I've downloaded Qt.py instead of using the pip installed one just because of this.

PR #23, opened 2016-05-30
Implement #22

mottosso:
```bash
$ set QT_PREFERRED_BINDING=PySide
$ python -c "from Qt import __binding__;print(__binding__)"
PySide
$ set QT_PREFERRED_BINDING=PyQt4
$ python -c "from Qt import __binding__;print(__binding__)"
PyQt4
$ set QT_PREFERRED_BINDING=NotExists
$ python -c "from Qt import __binding__;print(__binding__)"
ImportError: Preferred Qt binding "NotExists" not available
```

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -80,9 +80,12 @@ def _init():
 
     Please note: the entire Qt module is replaced with this code:
         sys.modules["Qt"] = binding()
-    This means no functions or variables can be called after this has
-    executed.
+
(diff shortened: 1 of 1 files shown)


mottosso:
Merging this now because tests are failing due to the missing PySide commit.

- https://travis-ci.org/pyblish/pyblish-lite/builds/133873249

But let's still discuss the preferred binding feature here.

PR #30, opened 2016-05-30
Add __wrapper_version__

mottosso:
Shouldn't use `__version__`, because it might be confused with the `__version__` of the binding.

```bash
$ python -c "import Qt;print(Qt);print(Qt.__version__)"
<module 'PySide' from 'C:\Python27\lib\site-packages\PySide\__init__.pyc'>
1.2.4
```

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -21,6 +21,8 @@ Usage:
 import os
 import sys
 
+__version__ = "0.2.3"
+
 
 def _pyqt5():
     import PyQt5.Qt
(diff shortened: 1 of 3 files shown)


fredrikaverpil:
Very nice!

> `__version__ = "0.2.3"`

So, basically whenever we update Qt.py, we also bump the minor version by 1?
And within minutes (?) we can `pip install -U Qy.py` to get the new release. Very, very nice.

Issue #24, opened 2016-05-30
Wrapper for compiling ui files


fredrikaverpil:
Do we also wish to provide a facility to support differences between PySide/PyQt?

```python
from Qt import QtCore
from Qt import QtWidgets
from Qt import __binding__
if __binding__.startswith('PySide'):
    from Qt import QtUiTools
    import pysideuic
elif __binding__.startswith('PyQt'):
    from Qt import uic
    import sip
```

I'm using the above frequently now, but it could have looked like this with the help of the `Qt` module:

```python
from Qt import QtCore
from Qt import QtWidgets
from Qt import uic
from Qt import generator
```

In this case, perhaps `uic` and `generator` should be prefixed to stress they are custom wrappers.

mottosso:
Ah, interesting, I've never used these myself. What do they do?

fredrikaverpil:
They are used to load .ui files.

These are the most common modules to use when loading a .ui file:
- PyQt4.uic
- PySide.QtUiTools.QUiLoader

However, there's an issue with PySide's QUiLoader which makes it behave differently than PyQt's uic. When you've loaded the UI using QUiLoader, you don't access the objects through the expected namespace (if you consider PyQt to be "god"). I believe that could have been why `pysideuic` was developed, as it fixes that – but I don't for a fact this was the reasoning on why pysideuic was made.

Here's an example: https://gist.github.com/fredrikaverpil/a70f38c8b04c49aaf73c1d22846f924a

So far, I can tell that this same behaviour is true for PyQt5/PySide2.

It's important to note that functions of PyQt4.uic and pysideuic are not the same.

I'm not sure myself if this is a good idea, and I might have a better idea at the end of the week.

mottosso:
> Here's an example

Isn't there an easier way of doing that? It seems awefully complex.

With PyQt4 there is this.

- http://pyqt.sourceforge.net/Docs/PyQt4/designer.html#PyQt4.uic.loadUi

If we're wrapping any of that, I would expect something along these lines.

```python
from Qt import loadUi
```

Where `loadUi` is our own version and behaves like PySide2.

fredrikaverpil:
> It seems awefully complex.

Agreed. But there's no real equivalent of `PyQt4.uic.loadUi` in PySide. Instead, you have to use  `PySide.QtUiTools.QUiLoader`. Which is why I've gone through all that trouble.

[...]

PR #33, opened 2016-05-31
Provide Qt.load_ui

fredrikaverpil:
Fix https://github.com/mottosso/Qt.py/issues/24

    from Qt import load_ui

- The `load_ui` function is made-up and doesn't exist directly under either PyQt or PySide.
- I was unable to "alias" `QtUiTools.QUiLoader().load` directly, which is why I had to create a function.

Example usage in [this gist](https://gist.github.com/fredrikaverpil/2bcd40900a44ae0a89dbc60980804fdf). You can run the code from the gist in a shell or in Maya. I'm adding Nuke support into the gist now too...

`Qt.load_ui` becomes really useful, actually. I initially thought this wouldn't be the case. I'm going to rewrite a lot of our in-house code to support this for easier maintainability between Maya, Nuke, standalone and PySide, PyQt4, PySide2, PyQt5. The Qt.py module has surpassed my expectations already! 😄 

Version bump to 0.2.4 included in this PR.

When this functionality is in Qt.py, I'll write a simple example to the README and link to the gist or possibly create a wiki entry. @mottosso would you prefer wiki or gist?

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -21,7 +21,7 @@ Usage:
 import os
 import sys
 
-__version__ = "0.2.4"
+__version__ = "0.2.5"
 
 
 def _pyqt5():
@@ -133,6 +133,12 @@ def _init():
     preferred = os.getenv("QT_PREFERRED_BINDING")
 
     if preferred:
+
+        # Debug mode, used in installer
+        if preferred == "None":
+            sys.modules[__name__].__wrapper_version__ = __version__
+            return
+
         available = {
             "PySide2": _pyside2,
             "PySide": _pyside,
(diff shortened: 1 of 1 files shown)


mottosso:
About the version bump, I discovered that the installer doesn't work unless you had an available binding on your system. I just pushed a fix for this now under the same version increment. Once we're ready to merge, just up it to the next increment and we're good.

fredrikaverpil:
Okay, I think we are good to go then.

mottosso:
Oh! I actually wasn't finished commenting on this.

I'm not sure I think this is a good idea, mainly polluting the functions with this much code. Would it be possible to move the complexity to functions at module-level, and simply refer to them within each installer-function?

My problem overall is that this is a highly specific problem that now ends up polluting all things. Ideally I'd see this put to the side somewhere so it doesn't obstruct an understanding of the fundamentals.

PR #36, opened 2016-05-31
Provide Qt.load_ui (retry)

fredrikaverpil:
Fix https://github.com/mottosso/Qt.py/issues/24

    from Qt import load_ui

- The `load_ui` function is made-up and doesn't exist directly under either PyQt or PySide.
- I was unable to "alias" `QtUiTools.QUiLoader().load` directly, which is why I had to create a function.

Example usage in [this gist](https://gist.github.com/fredrikaverpil/2bcd40900a44ae0a89dbc60980804fdf). You can run the code from the gist in a shell or in Maya. I'm adding Nuke support too.

`Qt.load_ui` becomes really useful, actually. I initially thought this wouldn't be the case. I'm going to rewrite a lot of our in-house code to support this for easier maintainability between Maya, Nuke, standalone and PySide, PyQt4, PySide2, PyQt5. The Qt.py module has surpassed my expectations already! 😄 

Version bump to 0.2.5 included in this PR.

When this functionality is in Qt.py, I'll write a simple example to the README and link to the gist or possibly create a wiki entry. @mottosso would you prefer wiki or gist?

### Discussion
I'm defining the same `load_ui()` function twice. It feels like it could reside in a place where we put such global functions rather than placing it inside each "load function".

- Function placement (global vs local to the loading function)
- How should such global functions be named?
  - Prefixed?
    - `_load_ui`
    - `___load_ui`
    - `_pyside_load_ui`
    - `__pyside_load_ui`

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -21,16 +21,19 @@ Usage:
 import os
 import sys
 
-__version__ = "0.2.4"
+__version__ = "0.2.5"
 
 
 def _pyqt5():
     import PyQt5.Qt
+    from PyQt5 import uic
 
     # Remap
     PyQt5.QtCore.Signal = PyQt5.QtCore.pyqtSignal
     PyQt5.QtCore.Slot = PyQt5.QtCore.pyqtSlot
     PyQt5.QtCore.Property = PyQt5.QtCore.pyqtProperty
+
+    PyQt5.load_ui = uic.loadUi
 
     # Add
     PyQt5.__wrapper_version__ = __version__
@@ -43,6 +46,7 @@ def _pyqt5():
 
 def _pyqt4():
     import PyQt4.Qt
+    from PyQt4 import uic
 
     # Remap
     PyQt4.QtWidgets = PyQt4.QtGui
@@ -50,6 +54,8 @@ def _pyqt4():
     PyQt4.QtCore.Signal = PyQt4.QtCore.pyqtSignal
     PyQt4.QtCore.Slot = PyQt4.QtCore.pyqtSlot
     PyQt4.QtCore.Property = PyQt4.QtCore.pyqtProperty
+
+    PyQt4.load_ui = uic.loadUi
 
     # Add
     PyQt4.__wrapper_version__ = __version__
@@ -62,6 +68,7 @@ def _pyqt4():
 
 def _pyside2():
     import PySide2
+    from PySide2 import QtUiTools
 
     # Add
     PySide2.__wrapper_version__ = __version__
@@ -69,17 +76,39 @@ def _pyside2():
     PySide2.__binding_version__ = PySide2.__version__
     PySide2.__qt_version__ = PySide2.QtCore.qVersion()
 
+    # Remap
+    def load_ui(ui_filepath, *args, **kwargs):
+        """Wrap QtUiTools.QUiLoader().load()
+        for compatibility against PyQt5.uic.loadUi()
+
+        Args:
+            ui_filepath (str): The filepath to the .ui file
+        """
+        return QtUiTools.QUiLoader().load(ui_filepath)
+    PySide2.load_ui = load_ui
+
     return PySide2
 
 
 def _pyside():
     import PySide
     import PySide.QtGui
+    from PySide import QtUiTools
 
     # Remap
     PySide.QtWidgets = PySide.QtGui
 
     PySide.QtCore.QSortFilterProxyModel = PySide.QtGui.QSortFilterProxyModel
+
+    def load_ui(ui_filepath, *args, **kwargs):
+        """Wrap QtUiTools.QUiLoader().load()
+        for compatibility against PyQt4.uic.loadUi()
+
+        Args:
+            ui_filepath (str): The filepath to the .ui file
+        """
+        return QtUiTools.QUiLoader().load(ui_filepath)
+    PySide.load_ui = load_ui
 
     # Add
     PySide.__wrapper_version__ = __version__
@@ -104,7 +133,7 @@ def _init():
     preferred = os.getenv("QT_PREFERRED_BINDING")
 
     if preferred:
-        
+
         # Debug mode, used in installer
         if preferred == "None":
             sys.modules[__name__].__wrapper_version__ = __version__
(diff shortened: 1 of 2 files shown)


mottosso:
Thanks for this @fredrikaverpil and sorry for the back and forth! Very good of you to include context and thoughts. That's what these pull requests are really good for.

> @mottosso would you prefer wiki or gist?

An entry in the README would be wonderful. It's such a small project that I think it's safe to keep all documentation on the same page.

> I'm defining the same load_ui() function twice. 

I felt the same way. I think this is an excellent opportunity to *not* prefix with an underscore, as this will be a public-facing function. (Then the fact that the module is actually being replaced and that the global variables are going out of scope is implementation detail).

I would define it either at the top or bottom, and then install it like you are now. e.g. `PyQt5.load_ui = load_ui`

That way, the installer functions are kept clean, and the functionality can be reused.

> __pyside_load_ui

This is a good question. Should it reference any particular binding, because it behaves differently across them?

I think "no". Because they behave so differently, I think it's better we define a single function that behaves identically across bindings.

If you can specify the signature, precondition, invariant and output(s) of this function, then it won't matter which binding the user ends up with and the .ui file will be treated the same.

For example.

```python
def load_ui(fname):
  """Read a Qt Designer file and return a subclassable object

  This function bla bla, and is used for bla bla.

  Arguments:
    fname (str): Absolute path to .ui file.

  Usage:
    >> Window = load_ui("/path/to/myfile.ui")
    >> class MyWindow(Window):
    ..   def __init__(self, parent=None):
    ..     # and so on..

  Returns:
    QtWidgets.QWidget: Subclass this object and override its members
  """

  implementation_here()
```

fredrikaverpil:
> I would define it either at the top or bottom

Hm. Since we do a complete replacement of the `Qt` module, I can't see how the function can live at the bottom of the file. What am I missing here?

I was thinking I'd have to implement before the loading functions. No?

mottosso:
The order of the functions won't matter, as the `_init()` function is called after the entire file has been read.

fredrikaverpil:
I did place the function towards the bottom, but my linter doesn't seem too happy.

- Added `# Installations` under `# Remap` and `# Add`.
- I've prefixed my new function with `installation`, marking the fact that this function is not available from the Qt module directly. Instead it tells us it installs something.
- Regarding the docstring it becomes a little bit of a mix now, where the usage refers to how to use `Qt.load_ui` and the rest refers to how to use the `install_load_ui` function. Not sure if this is confusing?

mottosso:
> Added # Install under # Remap and # Add.

Maybe it's just me, but I would have considered this "adding" something and would have put it under add.

> I've prefixed my new function with install, stressing the fact that this function is not available from the Qt module directly. Instead it tells us it installs something into the Qt module.

Again, probably just personal preference as it probably doesn't matter, but I find the install function overly complex and might have gone with something along the lines of..

```python
def pyside_load_ui():
  do_pyside_stuff()

def pyqt_load_ui():
  do_pyqt_stuff()
```

And in the binding function..

```python
def _pyqt5():
  ...
  # Add
  PyQt5.load_ui = pyqt_load_ui
```

Either way, I think this is much more readable!

One question; is it true that both `uic.loadUi` and `QtUiTools.QUiLoader().load` does the *exact* same thing? As in, if I were to run my code on a PySide system, would it still run without modifications on a PyQt5 system? That's important.

> Regarding the docstring it becomes a little bit of a mix now, where the usage refers to how the end-user would use Qt.load_ui() and the rest refers to how Qt is using install_load_ui() internally. Not sure if this is confusing?

That's fine, it's the truth. It's how you use this function, even if is it called by another name.

mottosso:
About the docstring, I generally go with Google Napoleon.

- https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

When it comes to the final `"""` and a newline, I tend to go with the one for `ExampleError`. That is, a newline between the docstring and start of code.

Also pay attention to the `>>>`. Three of those and it will be picked up by doctest. Just make sure to format it like it would appear in the interpreter. That is..

```python
>>> # Notice the dots..
>>> class MyClass(object):
...   pass
...
>>> MyClass()
```

fredrikaverpil:
> I would have considered this "adding"

Well, I don't mind. We could do that. Should I still call the function `install_load_ui()` or would you rather see `add_` as prefix instead of `install_`? (or read on regarding two separate functions)

> `pyside_load_ui()` and `pyqt_load_ui()`

We could do that, but I guess I'd still have to give them the module argument which means we'd end up with two quite similar functions. Meaning;

    def _pyqt5():
    from PyQt5 import uic
    ...
    # Add
    PyQt5.load_ui = pyqt_load_ui(module=uic)

Or did you have something else in mind?
You decide. I don't mind either way.


> is it true that both `uic.loadUi` and `QtUiTools.QUiLoader().load` do the *exact* same thing?

No (but read on).

> As in, if I were to run my code on a PySide system, would it still run without modifications on a PyQt5 system?

Yes!

They're *not* entirely identical, and when you need to mess about with anything other than the filename to the .ui file, you shouldn't use the custom `Qt.load_ui()`. This is information is useful to include in the docstring/README by the way. However, it does work really well when I load .ui files in PySide/PySide2/PyQt4/PyQt5. Works in Maya and in Nuke too. And all of that using the exact same code and the exact same `Qt.py`.

In case you wish to check out the docs:
- [`PySide.QtUiTools.QUiLoader().load`](https://srinikom.github.io/pyside-docs/PySide/QtUiTools/QUiLoader.html#PySide.QtUiTools.PySide.QtUiTools.QUiLoader.load)
- [`PyQt4.uic`](http://pyqt.sourceforge.net/Docs/PyQt4/designer.html#module-PyQt4.uic) (scroll down to `uic.loadUi`)

mottosso:
> We could do that, but I guess I'd still have to give them the module argument which means we'd end up with two quite similar functions

Aah, good point.

> They're not entirely identical, and when you need to mess about with anything other than the filename to the .ui file, you shouldn't use the custom Qt.load_ui().

Ah, ok. Sounds fair.

> You decide. I don't mind either way.

I'm good with this. The only thing would be the docstring, could you take a look at that?

Also, if you're adding to the README, I would that here in the same PR so it's nice and contained.

fredrikaverpil:
> The only thing would be the docstring, could you take a look at that?

Updated it!

> README

Added a simple example to the README. I'm working on a much more complex example, which I will share at a later time which shows how to load a main UI and then load additional UIs into it.

mottosso:
Nice, merge ready!

fredrikaverpil:
Cool!

PR #41, opened 2016-06-01
sip API v2

### LINKED ISSUE #37: PyQt4 import should try and set sip api to v2
It would be good, for compatibility, for the PyQt4 import logic to also attempt to set the sip api version to 2 for at least QString and QVariant (setapi()). This will ensure that PyQt4 bindings behave the same way as PySide, when the helper library is hiding away the import details.

Caveat: It is possible for the setting of sip api v2 to fail, if another library was imported first, which uses PyQt4 and already locked sip into v1.

fredrikaverpil:
Fix #37

@justinfx @mottosso Let's discuss this.

I get this when the API was already set to v2:

```python
python
>>> import Qt
Qt Warning: API 'QString' has already been set to version 1
Qt Warning: API 'QVariant' has already been set to version 1
```

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -53,12 +53,12 @@ def _pyqt4():
         sip.setapi("QTextStream", 2)
         sip.setapi("QTime", 2)
         sip.setapi("QUrl", 2)
-    except AttributeError as error:
-        raise ImportError(error)
+    except AttributeError:
+        raise ImportError
         # PyQt4 < v4.6
-    except ValueError as error:
+    except ValueError:
         # API version already set to v1
-        raise ImportError(error)
+        raise ImportError
 
     import PyQt4.Qt
 
--- a/Qt.py
+++ b/Qt.py
@@ -46,6 +46,7 @@ def _pyqt5():
 def _pyqt4():
     import PyQt4.Qt
     from PyQt4 import uic
+    import sip
 
     # Remap
     PyQt4.QtWidgets = PyQt4.QtGui
@@ -59,7 +60,7 @@ def _pyqt4():
     PyQt4.__binding_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.__qt_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.load_ui = install_load_ui(module=uic)
-    set_sip_api_v2()
+    set_sip_api_v2(module=sip)
 
     return PyQt4
 
@@ -97,8 +98,13 @@ def _pyside():
     return PySide
 
 
-def set_sip_api_v2():
-    """Attempt to set sip API to v2"""
+def set_sip_api_v2(module):
+    """Attempt to set sip API to v2
+
+        Args:
+            module (module): The sip module
+
+    """
 
     def setapi(name, version):
         """Set the given API `version` to the object `name`.
@@ -117,10 +123,8 @@ def set_sip_api_v2():
         except ValueError as error:
             sys.stdout.write('Qt Warning: ' + str(error) + '\n')    
 
-    import sip
     setapi(name='QString', version=2)
     setapi(name='QVariant', version=2)
-
 
 
 def install_load_ui(module):
--- a/Qt.py
+++ b/Qt.py
@@ -59,6 +59,7 @@ def _pyqt4():
     PyQt4.__binding_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.__qt_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.load_ui = install_load_ui(module=uic)
+    set_sip_api_v2()
 
     return PyQt4
 
@@ -94,6 +95,32 @@ def _pyside():
     PySide.load_ui = install_load_ui(module=QtUiTools)
 
     return PySide
+
+
+def set_sip_api_v2():
+    """Attempt to set sip API to v2"""
+
+    def setapi(name, version):
+        """Set the given API `version` to the object `name`.
+
+        If the function fails because the API had already been set to
+        e.g. version 1, it will grab the ValueError and print it.
+
+        Args:
+            name (str): The name of the class to set
+            version (int): The version of the API to set
+
+        """
+
+        try:
+            sip.setapi(name, version)
+        except ValueError as error:
+            sys.stdout.write('Qt Warning: ' + str(error) + '\n')    
+
+    import sip
+    setapi(name='QString', version=2)
+    setapi(name='QVariant', version=2)
+
 
 
 def install_load_ui(module):
--- a/Qt.py
+++ b/Qt.py
@@ -69,15 +69,14 @@ def _pyqt4():
             sip.setapi('QTextStream', 2)
             sip.setapi('QTime', 2)
             sip.setapi('QUrl', 2)
-            PyQt4.__sip_api_version__ = 2
         except AttributeError as error:
             # PyQt < v4.6
-            PyQt4.__sip_api_version__ = sip.getapi('QString')
+            sys.stdout.write('Qt.py Warning: ' + str(error) + '\n')
         except ValueError as error:
-            # Version 1 already set
-            PyQt4.__sip_api_version__ = sip.getapi('QString')
+            # Version (1?) already set
+            sys.stdout.write('Qt.py Warning: ' + str(error) + '\n')
     except ImportError as error:
-        sys.stdout.write('Qt Warning: ' + str(error) + '\n')
+        sys.stdout.write('Qt.py Warning: ' + str(error) + '\n')
 
     return PyQt4
 
--- a/Qt.py
+++ b/Qt.py
@@ -53,11 +53,12 @@ def _pyqt4():
         sip.setapi("QTextStream", 2)
         sip.setapi("QTime", 2)
         sip.setapi("QUrl", 2)
-    except AttributeError:
-        raise AttributeError("Qt.py error: PyQt version < 4.6 not supported")
+    except AttributeError as error:
+        raise ImportError(error)
+        # PyQt4 < v4.6
     except ValueError as error:
         # API version already set to v1
-        raise ValueError("Qt.py error: " + str(error))
+        raise ImportError(error)
 
     import PyQt4.Qt
 
--- a/Qt.py
+++ b/Qt.py
@@ -21,7 +21,7 @@ Usage:
 import os
 import sys
 
-__version__ = "0.2.7"
+__version__ = "0.3.0"
 
 
 def _pyqt5():
--- a/Qt.py
+++ b/Qt.py
@@ -21,7 +21,7 @@ Usage:
 import os
 import sys
 
-__version__ = "0.2.6"
+__version__ = "0.2.7"
 
 
 def _pyqt5():
@@ -44,24 +44,20 @@ def _pyqt5():
 
 def _pyqt4():
     # Attempt to set sip API v2 (must be done prior to importing PyQt4)
+    import sip
     try:
-        import sip
-        try:
-            sip.setapi('QString', 2)
-            sip.setapi('QVariant', 2)
-            sip.setapi('QDate', 2)
-            sip.setapi('QDateTime', 2)
-            sip.setapi('QTextStream', 2)
-            sip.setapi('QTime', 2)
-            sip.setapi('QUrl', 2)
-        except AttributeError as error:
-            # PyQt < v4.6
-            sys.stdout.write('Qt.py Warning: ' + str(error) + '\n')
-        except ValueError as error:
-            # Version (1?) already set
-            sys.stdout.write('Qt.py Warning: ' + str(error) + '\n')
-    except ImportError as error:
-        sys.stdout.write('Qt.py Warning: ' + str(error) + '\n')
+        sip.setapi("QString", 2)
+        sip.setapi("QVariant", 2)
+        sip.setapi("QDate", 2)
+        sip.setapi("QDateTime", 2)
+        sip.setapi("QTextStream", 2)
+        sip.setapi("QTime", 2)
+        sip.setapi("QUrl", 2)
+    except AttributeError:
+        raise AttributeError("Qt.py error: PyQt version < 4.6 not supported")
+    except ValueError as error:
+        # API version already set to v1
+        raise ValueError("Qt.py error: " + str(error))
 
     import PyQt4.Qt
 
--- a/Qt.py
+++ b/Qt.py
@@ -60,7 +60,7 @@ def _pyqt4():
     PyQt4.__binding_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.__qt_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.load_ui = install_load_ui(module=uic)
-    set_sip_api_v2(module=sip)
+    set_sip_api_v2(sip_module=sip)
 
     return PyQt4
 
@@ -98,11 +98,11 @@ def _pyside():
     return PySide
 
 
-def set_sip_api_v2(module):
-    """Attempt to set sip API to v2
+def set_sip_api_v2(sip_module):
+    """Attempt to set sip API to v2.
 
         Args:
-            module (module): The sip module
+            sip_module (module): The sip module
 
     """
 
@@ -123,6 +123,7 @@ def set_sip_api_v2(module):
         except ValueError as error:
             sys.stdout.write('Qt Warning: ' + str(error) + '\n')    
 
+    sip = sip_module
     setapi(name='QString', version=2)
     setapi(name='QVariant', version=2)
 
--- a/Qt.py
+++ b/Qt.py
@@ -58,6 +58,27 @@ def _pyqt4():
     PyQt4.__qt_version__ = PyQt4.QtCore.QT_VERSION_STR
     PyQt4.load_ui = pyqt4_load_ui
 
+    # Attempt to set sip API v2
+    try:
+        import sip
+        try:
+            sip.setapi('QString', 2)
+            sip.setapi('QVariant', 2)
+            sip.setapi('QDate', 2)
+            sip.setapi('QDateTime', 2)
+            sip.setapi('QTextStream', 2)
+            sip.setapi('QTime', 2)
+            sip.setapi('QUrl', 2)
+            PyQt4.__sip_api_version__ = 2
+        except AttributeError as error:
+            # PyQt < v4.6
+            PyQt4.__sip_api_version__ = sip.getapi('QString')
+        except ValueError as error:
+            # Version 1 already set
+            PyQt4.__sip_api_version__ = sip.getapi('QString')
+    except ImportError as error:
+        sys.stdout.write('Qt Warning: ' + str(error) + '\n')
+
     return PyQt4
 
 
--- a/Qt.py
+++ b/Qt.py
@@ -43,22 +43,7 @@ def _pyqt5():
 
 
 def _pyqt4():
-    import PyQt4.Qt
-
-    # Remap
-    PyQt4.QtWidgets = PyQt4.QtGui
-    PyQt4.QtCore.Signal = PyQt4.QtCore.pyqtSignal
-    PyQt4.QtCore.Slot = PyQt4.QtCore.pyqtSlot
-    PyQt4.QtCore.Property = PyQt4.QtCore.pyqtProperty
-
-    # Add
-    PyQt4.__wrapper_version__ = __version__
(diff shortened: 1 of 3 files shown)


Qt.py
```diff
@@ -96,6 +98,36 @@ def _pyside():
...
+
+        """
+
+        try:
+            sip.setapi(name, version)
+        except ValueError as error:
+            sys.stdout.write('Qt Warning: ' + str(error) + '\n')    
+
+    sip = sip_module
+    setapi(name='QString', version=2)
```
mottosso: I think it's safe to assume that if one fails, they both fail.

Qt.py
```diff
@@ -96,6 +98,36 @@ def _pyside():
     return PySide
 
 
+def set_sip_api_v2(sip_module):
```
mottosso: I don't think this is necessary. A simple try/except inside the `_pyside4` function should be clear enough.

fredrikaverpil:
> I think it's safe to assume that if one fails, they both fail.

Well, you *can* set them individually, which is why I didn't just want to throw a general error about not being able to enable API v2. Right now, you see which ones are not successfully set to v2. If we bundle them both under the same try/except, you won't be able to see that fine-grained information.

But I'm fine with that. I'll make a change.

fredrikaverpil:
Now I get this:

```python
python
>>> import Qt
Qt Warning: API 'QString' has already been set to version 1
```

Qt.py
```diff
@@ -60,6 +61,12 @@ def _pyqt4():
     PyQt4.__qt_version__ = PyQt4.QtCore.PYQT_VERSION_STR
     PyQt4.load_ui = install_load_ui(module=uic)
 
+    try:
+        sip.setapi('QString', 2)
+        sip.setapi('QVariant', 2)
+    except ValueError as error:
+        sys.stdout.write('Qt Warning: ' + str(error) + '\n')
```
mottosso:
```python
sys.stdout.write('Qt Warning: ' + str(error) + '\n')`
```

This could be something like..

```python
sys.stdout.write('Qt Warning: Using sip API v1')
```

fredrikaverpil:
> `sys.stdout.write('Qt Warning: Using sip API v1')`

Yes. I can change that. Although it would only be true for `QString` and not necessarily for `QVariant`.

mottosso:
I need a couple of test cases to see where this will actually matter. Right now, we're preparing for something we haven't seen break yet. This is how over-engineering is born.

Here's what I think. Let's either find usecases, or wait until we encounter them ourselves before implementing this feature.

fredrikaverpil:
> Let's either find usecases, or wait until we encounter them ourselves before implementing this feature.

Sure. I'll leave the PR pending. Good thing I created a branch for this ;)

mottosso:
> Good thing I created a branch for this ;)

Indeed. :) I always discover too late that I needed a branch.

justinfx:
A test case would be to import a module that imports PyQt4 directly and
doesn't set the sip api. Then import something that uses Qt.py

This should be pretty likely in a shared environment like Maya where you
don't have 100% ownership of all the tools that play together.

mottosso:
> A test case would be to import a module that imports PyQt4 directly and
doesn't set the sip api. Then import something that uses Qt.py 

Thanks. Now all we need is a usecase. When and why would one want to use v2 in the first place?

justinfx:
On Thu, 2 Jun 2016 7:52 AM Marcus Ottosson <notifications@github.com> wrote:

> A test case would be to import a module that imports PyQt4 directly and
> doesn't set the sip api. Then import something that uses Qt.py
>
> Thanks. Now all we need is a usecase. When and why would one want to use
> v2 in the first place?
>

When one has already written or wants to write their python Qt code in the
more pythonic style that PySide offers by default. It is more natural for,
say,  a model to return native python types from calls to data() as opposed
to wrapper types like QVariant which exist because C++ is a statically
typed language.

Another use case is where it isn't directly intended to use sip api 2. One
may have written a library using PySide and would expect it to run
seamlessly on a system that only has PyQt4. So indirectly they want/need
sip api 2 if they are relying on a helper module to make it all work.

mottosso:
> It is more natural for, say,  a model to return native python types from calls to data() as opposed
to wrapper types like QVariant which exist because C++ is a statically typed language.

Would it be possible to whip up a small example of that?

justinfx:
@mottosso , sure, here you go:

```python
# A function within a library that wants to be able to
# get some data from a model and format it a certain way.
def getFirstLabel():
    model = QtGui.QStringListModel(['a','b','c'])
    firstLabel = model.index(0,0).data()
    return "Label=" + firstLabel

# PySide
# 
from PySide import QtGui
getFirstLabel()
# u'Label=a'

# PyQt4 w/ sip api v2 
# 
import sip
sip.setapi("QString", 2)
sip.setapi("QVariant", 2)
from PyQt4 import QtGui
getFirstLabel()
# u'Label=a'

# PyQt4 w/ sip api v1
# 
from PyQt4 import QtGui
getFirstLabel()
# TypeError: cannot concatenate 'str' and 'QVariant' objects
```

mottosso:
> synced with upstream

It's not a biggie, but there is a cleaner way of handling this.

Have a look at "rebase" and "fast-forward". Rebasing means putting your changes on-top of the latest master branch, as opposed to merge with adds changes ontop of your changes and things gets messy.

The end result is an unmodified master branch, with only your changes added on-top.

fredrikaverpil:
> Have a look at "rebase" and "fast-forward"

Thanks, will do that!

mottosso:
> @mottosso , sure, here you go:

Thanks @justinfx, appreciate that!

fredrikaverpil:
> Have a look at "rebase" and "fast-forward".

Great, thanks!

mottosso:
It did not, you can tell by GitHub's red messages. :)

Usually, when are a few commits behind but aren't changing anything that newer commits are changing, such as the `.travis.yml`, there isn't any need to update your repository before merging. Merging can happen anyway.

In this case, I think it could have been safe.

Otherwise, make sure that after you've rebased *your* commits are on-top of the latest commits found in *this* repository.

A manual way of rebasing it would be to checkout master from this fork and then cherry-pick your new commits onto it.

fredrikaverpil:
> It did not, you can tell by GitHub's red messages. :)

Yes... man, rebasing seems difficult (?) I can't get it to work...

fredrikaverpil:
Sorry, I don't have time to get this right now it seems. I prioritised getting the change up for review instead. Need to look into rebasing at a later time.

Unless we wish to put this change in its own function, I believe this should be good enough to merge now. We can go back later and refine/restructure in my opinion.

fredrikaverpil:
Sorry, I don't have time to get this rebasing right. I just keep ending up in the same error loop... I prioritised getting the change up for review/discussion instead. Need to look into rebasing at a later time.

Judging from the example from @justinfx (thanks for that by the way!), we should set sip API v2 by default and encourage sip API v2 use. However, I feel that at least at this point it must be up to the end-user to make sure API v1 has not already been set. We should provide a way to query which API is currently set though.

Please let me know how you feel about...
- 20 rows of code added to  `_load_pyqt4`. Should this setting of API v2 be put in its own function?
- Should this setting of v2 also be performed in `_load_pyqt5`?
- I'm setting everything to API v2 now, not just `QString` and `QVariant`. Seems more consistent.
- No other errors right now except `ImportError`. Do we want a visible (`sys.stdout.write`) warning when v2 cannot be set? An alternative is to instead document the `Qt.__sip_api_version__` in the README.
- I wonder if we should set `Qt.__sip_api_version__` for PyQt5, PySide and PySide2 as well, so that in your code you can create conditions based on it? This raises a design guideline actually (I'm sure we're going to stumble upon this question again). An alternative is the end-user would have to first check `Qt.__binding__` instead. @mottosso @justinfx  what do you think? Example values:
  - PyQt4: `1` or `2`
  - PyQt5: ?
  - PySide = `0`
  - PySide2 = `0`
- Any other additions to this initial implementation?

A test needs to be added as well as a README entry.

fredrikaverpil:
Sorry, I don't have time to get this rebasing right. I just keep ending up in the same error loop... I prioritised getting the change up for review/discussion instead. Need to look into rebasing at a later time.

Judging from the example from @justinfx (thanks for that by the way!), we should set sip API v2 by default and encourage sip API v2 use. However, I feel that at least at this point it must be up to the end-user to make sure API v1 has not already been set. We should provide a way to query which API is currently set though.

Adding a bit of complexity to the example by @justinfx; notice how PySide2 uses `QtGui` and PyQt5 uses `QtCore`.

```python
# A function within a library that wants to be able to
# get some data from a model and format it a certain way.
def getFirstLabel_PyQt5():
    model = QtCore.QStringListModel(['a','b','c'])
    firstLabel = model.index(0,0).data()
    return "Label=" + firstLabel

def getFirstLabel_PySide2():
    model = QtGui.QStringListModel(['a','b','c'])
    firstLabel = model.index(0,0).data()
    return "Label=" + firstLabel

# PyQt5 w/ sip api v2 
# 
import sip
sip.setapi("QString", 2)
sip.setapi("QVariant", 2)
from PyQt5 import QtCore
print getFirstLabel_Qt5()
# u'Label=a'

# PySide2
# 
from PySide2 import QtGui
print getFirstLabel_PySide2()
# u'Label=a'
```


Please let me know how you feel about...
- 20 rows of code added to  `_load_pyqt4`. Should this setting of API v2 be put in its own function?
- Should this setting of v2 also be performed in `_load_pyqt5`?
- I'm setting everything to API v2 now, not just `QString` and `QVariant`. Seems more consistent.
- No other errors right now except `ImportError`. Do we want a visible (`sys.stdout.write`) warning when v2 cannot be set? An alternative is to instead document the `Qt.__sip_api_version__` in the README.
- I wonder if we should set `Qt.__sip_api_version__` for PyQt5, PySide and PySide2 as well, so that in your code you can create conditions based on it? This raises a design guideline actually (I'm sure we're going to stumble upon this question again). An alternative is the end-user would have to first check `Qt.__binding__` instead. @mottosso @justinfx  what do you think? Example values:
  - PyQt4: `1` or `2`
  - PyQt5: seems you need to first `sip.setapi('QString', 2')` or similar before you can `sip.getapi('QString')`?
  - PySide = `0`
  - PySide2 = `0`
- Any other additions to this initial implementation?

A test needs to be added as well as a README entry.

fredrikaverpil:
When we've discussed my previous post, I'd like to discuss this:

Below, the previously posted example by @justinfx with PyQt5/PySide2 additions:

```python
# A function within a library that wants to be able to
# get some data from a model and format it a certain way.
def getFirstLabel():
    model = QtGui.QStringListModel(['a','b','c'])
    firstLabel = model.index(0,0).data()
    return "Label=" + firstLabel

def getFirstLabel_PyQt5():
    model = QtCore.QStringListModel(['a','b','c'])
    firstLabel = model.index(0,0).data()
    return "Label=" + firstLabel

# PySide
# 
from PySide import QtGui
getFirstLabel()
# u'Label=a'

# PyQt4 w/ sip api v2 
# 
import sip
sip.setapi("QString", 2)
sip.setapi("QVariant", 2)
from PyQt4 import QtGui
getFirstLabel()
# u'Label=a'

# PyQt4 w/ sip api v1
# 
from PyQt4 import QtGui
getFirstLabel()
# TypeError: cannot concatenate 'str' and 'QVariant' objects

# PySide2
# 
from PySide2 import QtGui
print getFirstLabel()
# u'Label=a'

# PyQt5 w/ sip api v2 
# 
import sip
sip.setapi("QString", 2)
sip.setapi("QVariant", 2)
from PyQt5 import QtCore
print getFirstLabel_Qt5()
# u'Label=a'

# PyQt5 w/ sip api v2?
# 
from PyQt5 import QtCore
getFirstLabel_PyQt5()
# u'Label=a'

```

Notice how they all use `QtGui.QStringListModel` but PyQt5 uses `QtCore.QStringListModel` 😢 

It seems PyQt5 doesn't need to import sip and already has v2 set.

mottosso:
> Notice how they all use QtGui.QStringListModel but PyQt5 uses QtCore.QStringListModel

That is terrible. :S I hope it's a temporary bug. QStringListModel is officially in QtCore, as the official docs.

- http://doc.qt.io/qt-5/qstringlistmodel.html

fredrikaverpil:
> QStringListModel is officially in QtCore, as the official docs.

That means it's the PySide2 devs who got it wrong, right?

fredrikaverpil:
> A test case would be to import a module that imports PyQt4 directly and
doesn't set the sip api

@justinfx I can't figure out how to get the sip API version unless you explicitly set it yourself first. Do you know how to fetch the API version?

This works:

```python
>>> import sip
>>> sip.setapi('QString', 2)
>>> sip.getapi(QString')
2
```

...but this doesn't work:

```python
>>> import sip
>>> sip.getapi(QString')
ValueError: unknown API 'QString'
```

fredrikaverpil:
@justinfx scratch that, figured it out!

```python
>>> import PyQt4
>>> import sip
>>> from PyQt4 import QtCore
>>> sip.getapi('QString')
1
```

fredrikaverpil:
A bit of an update:

- Added tests for PyQt4 (should have API v1) and Qt.py with preferred PyQt4 binding (should have API v2)
- Removed `Qt.__sip_api_version__`
- Added warnings

However, it seems something has already set API v1 for Qt.py when using preferred PyQt4 binding.

fredrikaverpil:
> The first test sets the API, the second test tries setting it again. 

...aaah.

fredrikaverpil:
But. But. I was under the impression that each test got run isolated from each other?

If not, it feels like we should clean up after each test?

mottosso:
> But. But. I was under the impression that each test got run isolated from each other?

They are run in the same interpreter process, one function after the other.

> If not, it feels like we should clean up after each test?

Your feeling is right, hence `@with_setup(clean)` :)

If you are modifying state other than what's already in `clean()`, such as wanting `sip` to disappear, feel free to add it.

fredrikaverpil:
Hm, actually, this happens locally too.

```python
>>> import Qt
Qt.py Warning: API 'QString' has already been set to version 1
```

mottosso:
:) Stop for a minute and think things through. This isn't a race.

Have a read through tests.py as well, pay special attention to how `clean()` works and notice the use of `sys.modules.pop("Qt", None)`. The `None` keyword means it won't throw an error if it fails. No need for needless helper method `pop_module()`.

fredrikaverpil:
Hehe. I've been looking at this sporadically today and was using Travis to perform my testing which probably looks like a lot of erratic retries... :)

I'm using a container now, so testing goes a lot quicker. Thanks for the guide! 👍 

Found the issue in Qt.py now, and it's because `import PyQt4.Qt` actually sets the API version to 1. So I need to set API v2 first thing in `_load_pyqt4`. My tests still fail though, so I'll have to look at why that's happening.

Qt.py
```diff
@@ -43,6 +43,26 @@ def _pyqt5():
...
+        try:
+            sip.setapi('QString', 2)
+            sip.setapi('QVariant', 2)
+            sip.setapi('QDate', 2)
+            sip.setapi('QDateTime', 2)
+            sip.setapi('QTextStream', 2)
+            sip.setapi('QTime', 2)
+            sip.setapi('QUrl', 2)
+        except AttributeError as error:
+            # PyQt < v4.6
```
mottosso: How about we only support < 4.6? Otherwise, we'll have to support both sip v1 and v2.

fredrikaverpil:
> How about we only support > 4.6? Otherwise, we'll have to support both sip v1 and v2.

Definitively. That was my intention. I figured it would be useful to provide a warning though, when you're on < 4.6. Do you wish to see some change to that code?

justinfx:
Do you think it should only be a warning if there is an old sip/PyQt4 available? In my version of this helper I throw an ImportError about it. I don't know if its right or wrong... but my take on it is if someone has written a tool in PySide and then its getting run on some other machine that only has an older version of PyQt4, wouldn't it potentially have runtime crashes anyways? I suppose there is a chance that they don't hit these areas though.

fredrikaverpil:
> Do you think it should only be a warning if there is an old sip/PyQt4 available? In my version of this helper I throw an ImportError about it.

Since we're just going to support v1, I'm changing that into raising an error.

mottosso:
Yeah, I think we should raise an import error.

Qt.py
```diff
@@ -43,6 +43,26 @@ def _pyqt5():
 
 
 def _pyqt4():
+    # Attempt to set sip API v2 (must be done prior to importing PyQt4)
+    try:
+        import sip
```
mottosso: I'm not sure there's a need to try/except for `sip`. I think `sip` must be available for PyQt4 to work.

```bash
$ sudo mv /usr/lib/python2.7/dist-packages/sip.so /usr/lib/python2.7/dist-packages/_sip.so
$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> from PyQt4 import QtGui
ImportError: No module named sip
```

fredrikaverpil:
Ok, guys. Let me know what you think.

I'm raising an error if:
- PyQt4 < v4.6
- API v1 has already been set

...and Python itself will raise an ImportError if sip is missing.

tests.py
```diff
@@ -86,3 +88,15 @@ def test_coexistence():
 
         # But does not delete the original
         assert PySide.QtGui.QStringListModel
+
+
+@with_setup(clean)
+def test_sip_api_qtpy():
+    """Qt.py with preferred binding PyQt4 should have sip version 2"""
+
+    os.environ["QT_PREFERRED_BINDING"] = "PyQt4"
```
mottosso: Nitpick, what do you think about the contextmanager, and not declaring that extra variable? The variable doesn't add any new information I think.

```python
with pyqt4():
  import Qt
  import sip
  assert sip.getapi("QString") == 2, ("PyQt4 API version should be 2, "
                                      "instead is %s" % api_version)
```

Qt.py
```diff
@@ -43,6 +43,22 @@ def _pyqt5():
...
+    import sip
+    try:
+        sip.setapi("QString", 2)
+        sip.setapi("QVariant", 2)
+        sip.setapi("QDate", 2)
+        sip.setapi("QDateTime", 2)
+        sip.setapi("QTextStream", 2)
+        sip.setapi("QTime", 2)
+        sip.setapi("QUrl", 2)
+    except AttributeError:
```
mottosso: If these functions raise anything but `ImportError`, it will throw the actual exception to the user, when he *imports* the module. This I think might be unexpected.

Rather, I think they should all throw an import error, without message, since it is discarded [here](https://github.com/fredrikaverpil/Qt.py/blob/07d9720b7f3672c27fd51f7a0d8621609c5f5660/Qt.py#L220) anyway.

If we want to throw a message, I would print it to `sys.stderr.write()`.

mottosso:
Other than that, LGTM!

fredrikaverpil:
> Nitpick, what do you think about the contextmanager, and not declaring that extra variable?
> ...
> Rather, I think they should all throw an import error, without message

Have a look now so we're thinking the same thing.
Should we bump the Qt.py version before merging?

mottosso:
> Should we bump the Qt.py version before merging?

Bumping is fine at any time before a release. You can do it now if you want.

Since we're adding a feature, I'd make it a minor update as opposed to patch. That is, `0.3.0`.

Qt.py
```diff
@@ -43,6 +43,23 @@ def _pyqt5():
...
+    import sip
+    try:
+        sip.setapi("QString", 2)
+        sip.setapi("QVariant", 2)
+        sip.setapi("QDate", 2)
+        sip.setapi("QDateTime", 2)
+        sip.setapi("QTextStream", 2)
+        sip.setapi("QTime", 2)
+        sip.setapi("QUrl", 2)
+    except AttributeError as error:
```
mottosso: This works, but leaves the impression that someone down the line will catch this error and do something with it.

This is more clear I think.

```python
except AttributeError:
  # PyQt4 < v4.6
  raise ImportError

except ValueError:
  # API version already set to v1
  raise ImportError
```

README.md
```diff
@@ -168,6 +168,11 @@ sys.exit(app.exec_())
 
 Please note, for maximum compatibility, only pass the argument of the filename to the `load_ui` function.
 
+**sip API v2**
+
+If you're using PyQt4, `sip` attempts to set its API to version 2 for the following:
+- `QString`
+- `QVariant`
```
mottosso: Best to add all api's we're setting, less we're bound to return here at some point after an issue has been filed about it.

mottosso:
This is what I call a PR! Great work @fredrikaverpil!!

mottosso:
Merge away!

fredrikaverpil:
Let's nail this sucker!

PR #38, opened 2016-06-01
Initial setup for travis

mottosso:

PR #42, opened 2016-06-01
Add support for Travis CI

mottosso:
Keeping this open for subsequent commits while I stretch the legs of Travis.

(Follow-up to #38 where I got lost in git)

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,8 @@ install:
   - python -c "import PySide;print(PySide)"
   - python -c "import PyQt4;print(PyQt4)"
 
-# Enable virtual framebuffer.
+# Enable virtual framebuffer, this way we can instantiate
+# a QApplication and create widgets.
 # https://docs.travis-ci.com/user/gui-and-headless-browsers/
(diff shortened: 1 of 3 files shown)


mottosso:
Ok, this is fine now.

@fredrikaverpil Travis caught something about the `load_ui` function, which was that the class it was instantiating, `QUiLoader`, required a graphics card and window manager to be available. Suggesting it silently setup an QApplication instance in the background. We really don't want that happening implicitly.

So I did some quick surgery to remedy this, let me know if I broke something.

Speaking of which, we should now produce tests for this functionality so that it doesn't break again.

Did you have Docker? I could setup a small "developer guide" to setup a Travis-like environment locally so you can run the tests. Otherwise, you can just pull request and they will get run.

fredrikaverpil:
Nice!

> let me know if I broke something

Looks like everything's working fine, merge away!

> Did you have Docker? I could setup a small "developer guide"

Yes, I do. I'd appreciate a simple dev guide in case you have the time.

PR #43, opened 2016-06-02
Fix #40

### LINKED ISSUE #40: QT_VERSION_STR missing
`Qt.__qt_version__` is currently showing the PyQt version, when it should be showing the Qt version.

- [code](https://github.com/mottosso/Qt.py/blob/36cfe9d077bfe57f3e0ae8c3d73afaf578438011/Qt.py#L40)

fredrikaverpil:


mottosso:
Perfect. Could you add a test for this too? :)

Something that assumes to know which versions of PySide was compiled with, and which version of PySide is installed, and then compares that against the values in these variables?

fredrikaverpil:
> Perfect. Could you add a test for this too?

Will do.

mottosso:
Let's reduce the number of PRs. :)

PR #46, opened 2016-06-02
Travis with Ubuntu 14.04

mottosso:
Testing a few things here.


mottosso:
Bam, now we have testing under both Python 2 and 3

fredrikaverpil:
Excellent!

Issue #51, opened 2016-06-03
Enforce portability?


mottosso:
Should Qt.py enforce/encourage portability? That is, should Qt.py try and *prevent* a user from using functionality found only in a specific binding, so that any code written with it is *guaranteed* to run on any other binding?

It's more or less a road in the fork between being a helper library or a complete framework.

**Scenario A - Yes portability**

```python
>>> from Qt import QtCore
>>> signal = QtCore.pyqtSignal()
AttributeError: 'module' object has no attribute 'pyqtSignal'
```

**Scenario B - No portability**

```python
>>> from Qt import QtCore
>>> signal = QtCore.pyqtSignal()
>>> signal2 = QtCore.Signal()
>>> from PyQt4 import QtCore
>>> signal3 = QtCore.pyqtSignal()
```

### Implementation

We could achieve this by simply removing the redirected members.

```python
PyQt4.QtCore.Signal = PyQt4.QtCore.pyqtSignal
del(PyQt4.QtCore.pyqtSignal)
```

However the problem is that this affects the PyQt4 module globally; meaning that if the user is running another application using PyQt4, it would break.

We could work around this by, rather than redirecting Qt.py to its nearest binding..

```python
>>> import Qt
>>> print(Qt)
<module 'PyQt4' from '/usr/lib/python2.7/dist-packages/PyQt4/__init__.pyc'>
```

..we re-define each member..

```python
>>> from Qt import QtCore
>>> print(QtCore)
<module 'Qt.QtCore' from '/usr/lib/python2.7/dist-packages/Qt.pyc'>
```

But the problem with that is (1) we would have to redefine every possible member of Qt which might be difficult and (2) because of (1) we would lose the current simplicity of Qt.py.

### Discussion

If we choose to enforce portability, we'll have to consider the consequences and find an elegant way in which to implement it. If we chose not to, we should consistently ensure that Qt.py can be used alongside any particular binding.

An alternative approach might be to instead have a separate tool check for used features only found in any one binding?

```python
# psuedo
# validate_portability.py
import PySide
import PyQt4

unique = set(all_pyside).intersect(set(all_pyqt4))

with open("mycode.py") as f:
  used_members = parse(f)
  for member in used_members:
    if member in unique:
      warn("%s is not portable!" % member)
```

fredrikaverpil:
Hm. This is a tough one. What do you think?

[...]

PR #52, opened 2016-06-03
Enforce coexistence

mottosso:
Here's a version of scenario B from #51.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -63,6 +63,10 @@ def _pyqt4():
 
 def _pyside2():
     import PySide2
+    from PySide2 import QtGui, QtCore
+
+    # Remap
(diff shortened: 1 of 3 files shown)


fredrikaverpil:
I like this.

How would you feel about encouraging a certain coding guideline?
I would personally love to see something like this in the README:

```python
from Qt import QtCore

if Qt.__binding__.startswith('PyQt'):
    signal = QtCore.pyqtSignal()
elif Qt.__binding__.startswith('PySide'):
    signal = QtCore.Signal()
```python

mottosso:
> How would you feel about encouraging a certain coding guideline?

That's a good idea! The docs is a good place for this.

PR #58, opened 2016-06-04
Dockerfile

fredrikaverpil:
I set up a `Dockerfile` and a `docker-compose.yml` so that I don't have to rebuild the container each time I start it... and I figured I might as well share it.

However, this makes the test environment setup specified in three places:
- `.travis.yml`
- `README.md`
- `docker/Dockerfile`

We could simplify the current README instructions by removing the `apt-get install` part and instead point users in this direction, which would download a pre-built image (based on the `Dockerfile`):

    docker run -ti --rm -v /local/path/to/Qt.py:/root/Qt.py qtpy/qtpydev:1.0

If anyone's interested in inspecting the setup, you could just check out either `.travis.yml` or `Dockerfile`.

@mottosso in case you think this is nice and accept the PR, I'll add you as collaborator to the [dockerhub repository](https://hub.docker.com/r/qtpy/qtpydev/).


mottosso:
That's a good thought. I wonder if we could simplify things a bit.

mottosso:
With just the Dockerfile, we could have this.

```bash
$ docker build -t mottosso/qt.py https://github.com/mottosso/Qt.py.git # One-time build
$ docker run --rm -v $(pwd):/Qt.py mottosso/qt.py nosetests tests.py --verbose
```

We could also use this for our Travis tests, which might help the many specifications of dependencies.

- https://docs.travis-ci.com/user/docker/

mottosso:
As in, no docker-compose, no additional Docker hub project to maintain, no additional dependency on supervisord, and a smaller guide in the README.

fredrikaverpil:
Love it. Please review changes.

README.md
```diff
@@ -244,4 +235,36 @@ nosetests --verbose
 # OK
 ```
 
+You can also use the provided **docker-compose** setup:
```
mottosso: This can be stripped now.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:14.04
+RUN apt-get update && apt-get install -y \
+    python-qt4 \
+    python-pyside \
+    python3-pyqt4 \
```
mottosso: We can probably strip the Python 3 libraries, as we're only running the tests with Python 2 anyway, and it's probably unnecessary to run both locally anyway.

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+    python3-pyside \
+    python-pip
+RUN pip install nose
+ENTRYPOINT nosetests --verbose /Qt.py/tests.py
```
mottosso: That's nice. Even shorter than what I had in mind. :)

```bash
WORKDIR /Qt.py
```

In which case we would still need to specify `nose` and things via the command-line.

README.md
```diff
@@ -214,25 +214,16 @@ This project uses Travis for continuous integration and Travis uses Ubuntu 14.04
 Assuming you have Docker already setup.
 
 ```bash
-# Explanation of flags
-# -ti 	interactive session
-# --rm 	delete the image on exit
+# Build image (see Dockerfile for specifics)
+# Re-run this command to pull latest version of image
+docker build --pull --force-rm -t mottosso/qtpy https://github.com/mottosso/Qt.py.git
```
mottosso: Why `--pull` and `--force-rm`?

fredrikaverpil:
> `WORKDIR /Qt.py`
> In which case we would still need to specify nose and things via the command-line.

Sorry, not following you here... what do you mean? :)

> Why `--pull` and `--force-rm`?

Hm, for some reason, I thought that `--pull` would force a re-build. But now I see it was `--no-cache` I was really looking for. I'll use `--no-cache` instead, so if you re-run that very same command you're also getting the latest and greatest.

About `--force-rm`, I was under the impression that would remove intermittent containers which are only used during the build and not during run (saving disk space)... no?

mottosso:
> Sorry, not following you here... what do you mean? :)

There is another directive called `WORKDIR` which sets the working directory for any commands  you run on the container.

```bash
FROM ubuntu:14.04

RUN apt-get update && apt-get install -y \
    python-qt4 \
    python-pyside \
    python3-pyqt4 \
    python3-pyside \
    python-pip
RUN pip install nose

WORKDIR /Qt.py
```

As in my first example, with the workdir set to `/Qt.py`, the command would be.

```bash
$ docker run --rm -v $(pwd):/Qt.py mottosso/qt.py nosetests tests.py --verbose
```

As opposed to..

```bash
$ docker run --rm -v $(pwd):/Qt.py mottosso/qt.py /Qt.py/nosetests tests.py --verbose
```

But `ENTRYPOINT` reduces typing even further, so it's a win.

> I'll use --no-cache instead

I haven't used that before and haven't had issues with a vanilla build. Maybe let's go plain until a problem arises? Otherwise I'm afraid of over-engineering and cargo cult programming.

> About --force-rm, I was under the impression that would remove intermediate containers which are only used during the build and not during run (saving disk space)... no?

It looks like it is some form of optimisation, yes. But I'd keep it out of the beginner's docs to make the already long command less scary.

mottosso:
That's mergable to me!

fredrikaverpil:
Thanks for clarifying that last bit. :)

Really nice improvements suggestions on this one. Thanks! :D

PR #59, opened 2016-06-04
Test setup.py and Dockerfile

mottosso:


mottosso:
With building and running the Dockerfile, we're *almost* at a point where we can ditch having Travis run our tests and simply rely on `docker run`.

The only remaining thing I haven't wrapped my head around, is how we run the Dockerfile once per Python distribution.. Maybe the answer is to have it do that itself, from within the Dockerfile? :O

The benefit of using the Dockerfile only, is that we get reproducible results locally, but primarily that we can install and mess about without restriction server-side. (Like avoiding that silly [whitelist problem](https://github.com/travis-ci/apt-package-whitelist/issues/433) with `python-pyside`).

fredrikaverpil:
LGTM!

mottosso:
Giving this a try, now tests are run entirely in Docker on both Python 2 and 3.

mottosso:
Reverting this.

Some downsides, nothing too awful, but enough to prefer the current setup.

1. We lose out on parallelism because both containers are built and run within the same Travis instance.
2. We could work around it, by using a build matrix, but then we lose out on the `on:` parameter for deploying to PyPI
3. It's likely (2) can be worked around too.
4. This is not enough of a reason, but for completeness, time. Building a container takes longer, around 2x.
5. Multiple Dockerfiles. Not entirely avoidable, but probably not a big concern either. See the commit for details on how the files are written and used.

fredrikaverpil:
Did you merge this despite the downsides?

mottosso:
The downsides was related to before I reverted. To having multiple docker files. I merged at where you said "LGTM". :)

PR #60, opened 2016-06-05
Implement nosepipe

### LINKED ISSUE #56: tests: clean function
In tests.py, I find the clean function a bit confusing...

Some points:
- Python [does not support unloading modules](http://bugs.python.org/issue9072)
- Nose has this [`--with-isolation`](http://nose.readthedocs.io/en/latest/plugins/isolate.html) option, which does its best to isolate each test function.
- Currently, `sys.modules.pop()` is being used to "clean out" modules, but it doesn't actually clean out modules imported by the module you're popping.

If we just disregard that Python can't "unimport" modules and see how far we can push it... and skip `--with-isolation` since we might not want to isolate all functions –– shouldn't this be a more reliable cleaning function?

```python
import os
import sys
import imp
import contextlib

from nose.tools import (
    with_setup,
    assert_raises,
)

_modules = sys.modules.copy()

def clean():
    """Provide clean working environment"""
    os.environ.pop("QT_PREFERRED_BINDING", None)
    modules_to_delete = list(set(sys.modules) - set(_modules))
    for module in modules_to_delete:
        sys.modules.pop(module)  # or "del sys.modules[module]"
```

Right now, even if you pop `Qt`, the modules imported by `Qt` aren't popped.

Also, with nose, we can create a `tests folder` (rather than the `tests.py` file), and inside it put several python scripts prefixed with `test_` and nose will run all of them.

@mottosso what do you think about this?

fredrikaverpil:
Run all tests isolated from each other.
If we wish to only see some tests run in its own subprocess, we could implement that at a later time.

Fix #56

This made it possible to also test PyQt4 for sip API v1.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -80,6 +80,16 @@ def test_coexistence():
         assert PySide.QtGui.QStringListModel
 
 
+def test_sip_api_pyqt4():
+    """PyQt4 should have sip version 1"""
+
+    from PyQt4 import QtCore
+    import sip
+    assert sip.getapi("QString") == 1, ("PyQt4 API version should be 1, "
+                                        "instead is %s"
+                                        % sip.getapi("QString"))
+
+
 def test_sip_api_qtpy():
     """Qt.py with preferred binding PyQt4 should have sip version 2"""
 
@@ -89,13 +99,3 @@ def test_sip_api_qtpy():
         assert sip.getapi("QString") == 2, ("PyQt4 API version should be 2, "
                                             "instead is %s"
                                             % sip.getapi("QString"))
-
-
-def test_sip_api_pyqt4():
-    """PyQt4 should have sip version 1"""
-
-    from PyQt4 import QtCore
-    import sip
-    assert sip.getapi("QString") == 1, ("PyQt4 API version should be 1, "
-                                        "instead is %s"
-                                        % sip.getapi("QString"))
--- a/tests.py
+++ b/tests.py
@@ -81,7 +81,7 @@ def test_coexistence():
 
 
 def test_sip_api_pyqt4():
-    """PyQt4 default sip API version"""
+    """PyQt4 default sip API version (Python 2.x = 1, Python 3.x = 2)"""
 
     from PyQt4 import QtCore
     import sip
@@ -107,7 +107,8 @@ def test_sip_api_qtpy():
                                             % sip.getapi("QString"))
 
 def test_sip_api_already_set():
-    """Qt.py should cause ImportError when sip API v1 was already set"""
+    """Qt.py should cause ImportError when sip API v1 was already set
+    (Python 2.x only)"""
 
     with pyqt4():
         def import_qt():
--- a/tests.py
+++ b/tests.py
@@ -10,12 +10,11 @@ import sys
 import imp
 import contextlib
 
+from nose import SkipTest
 from nose.tools import (
     with_setup,
     assert_raises,
 )
-
-from nose.plugins.skip import SkipTest
 
 
 @contextlib.contextmanager
@@ -125,4 +124,4 @@ def test_sip_api_already_set():
             sip.setapi("QString", 1)
             assert_raises(ImportError, import_qt)
     else:
-        raise SkipTest("Test skipped when Python version not equal to 2.x")
+        raise SkipTest("Skipped because Python version is not 2.x")
--- a/tests.py
+++ b/tests.py
@@ -89,13 +89,3 @@ def test_sip_api_qtpy():
         assert sip.getapi("QString") == 2, ("PyQt4 API version should be 2, "
                                             "instead is %s"
                                             % sip.getapi("QString"))
-
-
-def test_sip_api_pyqt4():
-    """PyQt4 should have sip version 1"""
-
-    from PyQt4 import QtCore
-    import sip
-    assert sip.getapi("QString") == 1, ("PyQt4 API version should be 1, "
-                                        "instead is %s"
-                                        % sip.getapi("QString"))
--- a/tests.py
+++ b/tests.py
@@ -30,13 +30,6 @@ def pyside():
     os.environ.pop("QT_PREFERRED_BINDING")
 
 
-def clean():
-    """Provide clean working environment"""
-    sys.modules.pop("Qt", None)
-    os.environ.pop("QT_PREFERRED_BINDING", None)
-    sys.modules.pop("sip", None)
-
-
 def test_environment():
     """Tests require PySide and PyQt4 bindings to be installed"""
 
@@ -48,7 +41,6 @@ def test_environment():
     assert_raises(ImportError, imp.find_module, "PyQt5")
 
 
-@with_setup(clean)
 def test_preferred():
     """Setting QT_PREFERRED_BINDING properly forces a particular binding"""
     import Qt
@@ -66,7 +58,6 @@ def test_preferred():
                                          "instead got %s" % Qt)
 
 
-@with_setup(clean)
 def test_preferred_none():
     """Preferring None shouldn't import anything"""
 
@@ -75,7 +66,6 @@ def test_preferred_none():
     assert Qt.__name__ == "Qt", Qt
 
 
-@with_setup(clean)
 def test_coexistence():
     """Qt.py may be use alongside the actual binding"""
 
@@ -90,7 +80,6 @@ def test_coexistence():
         assert PySide.QtGui.QStringListModel
 
 
-@with_setup(clean)
 def test_sip_api_qtpy():
     """Qt.py with preferred binding PyQt4 should have sip version 2"""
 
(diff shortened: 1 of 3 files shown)


mottosso:
Don't forget to update `.travis.yml` too.

tests.py
```diff
@@ -100,3 +89,13 @@ def test_sip_api_qtpy():
         assert sip.getapi("QString") == 2, ("PyQt4 API version should be 2, "
                                             "instead is %s"
                                             % sip.getapi("QString"))
+
+
+def test_sip_api_pyqt4():
+    """PyQt4 should have sip version 1"""
```
mottosso: What exactly are you testing here? Seem to be testing PyQt4 rather than Qt.py.

Perhaps test for something like this?

```python
import Qt
assert_raises(ImportError, __import__, "PyQt4")
```

fredrikaverpil:
> What exactly are you testing here? Seem to be testing PyQt4 rather than Qt.py.

That was [a suggestion](https://github.com/mottosso/Qt.py/pull/41#issuecomment-223098678) by @justinfx; testing the sip API version when importing PyQt4 without explicitly setting the version. I guess it would just make sure that PyQt4 doesn't start setting v2 by default all of a sudden. Do you feel it's an unnecessary test?

mottosso:
Yes I know what you're trying to test, but think about it. That test can only ever fail if PyQt4 changes. It has nothing to do with Qt.py.

We should however have a test ready for those exception we're throwing in case sip is already set to sip v1.

- [Qt.py#L60](https://github.com/mottosso/Qt.py/blob/bd5bda163f1706317015b80ac6137fa9a6b5f007/Qt.py#L60)

fredrikaverpil:
> That test can only ever fail if PyQt4 changes. It has nothing to do with Qt.py.
> 
> We should however have a test ready for those exception we're throwing in case sip is already set to sip v1.

Good points. Will fix ... and find out why Travis screams about an ImportError...

mottosso:
Try this, in addition to what you already have.

```yaml
- sudo pip3 install nosepipe
```

mottosso:
Are you removing commits? :O I'm looking for the commit with the travis test that partially failed, but it looks like commits are missing.

Try not to force push or anything, at least not until we merge.

fredrikaverpil:
> Try not to force push or anything, at least not until we merge.

Learning [Gitkraken](https://www.gitkraken.com), sorry. I did a few force pushes and will try to avoid that from now on.

mottosso:
How come the test didn't work?

fredrikaverpil:
Whoa, okay now I'm not sure what the heck I did... I intended to rebase and bring back the failing PyQt4 test as I also wish to know why it was failing.

mottosso:
Are you deleting comments now too? :)

fredrikaverpil:
Ah, yeah sorry, I shouldn't have done that but I realized what I had just done and it was actually not so bad :)

fredrikaverpil:
Okay. Switching the order of the PyQt4 check vs the Qt.py one didn't make a difference, which is good in a sense. That means the isolation works.

fredrikaverpil:
> How come the test didn't work?

PyQt4 for Python 3 sets API v2:

```python
Python 3.4.3 (default, Oct 14 2015, 20:28:29)
[GCC 4.8.4] on linux
>>> from PyQt4 import QtCore
>>> import sip
>>> sip.getapi('QString')
2
```

mottosso:
> PyQt4 for Python 3 sets API v2

Wuw! How unexpected. Gotta research this..

mottosso:
Ok, looks like it's a feature, not a bug.

- https://wiki.python.org/moin/PortingPythonToPy3k/PyQt4

fredrikaverpil:
Sorry, I'm testing a few things now back and forth with Travis.

mottosso:
That's totally fine, that's what it's for.

fredrikaverpil:
We now get this using Python 3.4:

```
Qt.py should cause ImportError when sip API v1 was already set ... 
Ran 7 tests in 1.674s
OK
```

I would have expected something like this:

```
Qt.py should cause ImportError when sip API v1 was already set ... SKIP
Ran 7 tests in 1.674s
OK (SKIP=1)
```

But aside from this, I'd say we're ready to merge.

tests.py
```diff
@@ -100,3 +107,21 @@ def test_sip_api_qtpy():
...
+        with pyqt4():
+            def import_qt():
+                import Qt
+
+            from PyQt4 import QtCore
+            import sip
+            sip.setapi("QString", 1)
+            assert_raises(ImportError, import_qt)
+    else:
+        raise SkipTest("Skipped because Python version is not 2.x")
```
mottosso: Hmmmm, this itches somewhat.

How about this?

```python
if py == 3:
  def test_sip_api_already_set():
```

Where py is defined as..

```python
py = sys.version_info[0]
```

That way, tests stay focused on a single version.

fredrikaverpil:
Looks like your suggestion did work nicely.

```python
~/code/github/Qt.py feature/nosepipe*
❯ docker run -ti --rm -v $(pwd):/Qt.py mottosso/qt.py3
Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
PyQt4 default sip API version (Python 2.x = 1, Python 3.x = 2) ... ok
Qt.py with preferred binding PyQt4 should have sip version 2 ... ok

Ran 6 tests in 1.526s

OK

~/code/github/Qt.py feature/nosepipe
❯ docker run -ti --rm -v $(pwd):/Qt.py mottosso/qt.py
Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
PyQt4 default sip API version (Python 2.x = 1, Python 3.x = 2) ... ok
Qt.py with preferred binding PyQt4 should have sip version 2 ... ok
Qt.py should cause ImportError when sip API v1 was already set ... ok

Ran 7 tests in 1.250s

OK
```

tests.py
```diff
@@ -100,3 +107,21 @@ def test_sip_api_qtpy():
...
+    if sys.version_info[0] == 2:
+        # Python 2.x
+        with pyqt4():
+            def import_qt():
+                import Qt
+
+            from PyQt4 import QtCore
+            import sip
+            sip.setapi("QString", 1)
+            assert_raises(ImportError, import_qt)
```
mottosso: I think this can be simplified to:

```python
assert_raises(ImportError, __import__, "Qt")
```

tests.py
```diff
@@ -100,3 +107,21 @@ def test_sip_api_qtpy():
         assert sip.getapi("QString") == 2, ("PyQt4 API version should be 2, "
                                             "instead is %s"
                                             % sip.getapi("QString"))
+
+
+def test_sip_api_already_set():
+    """Qt.py should cause ImportError when sip API v1 was already set
```
mottosso: First line of a docstring should be a short summary, with no line breaks.

Subsequent lines can explain in more detail. Also not necessary to mention Qt.py, as all tests are about Qt.py.

tests.py
```diff
@@ -16,7 +16,7 @@
     assert_raises,
 )
 
-PYTHON_VERSION = sys.version_info[0]  # e.g. 2 or 3
+PY_VER_MAJOR = sys.version_info[0]  # e.g. 2 or 3
```
mottosso: Aw, why so cryptic? KISS. `PYTHON` reads well I think.

```python
if PYTHON == 3:
  # then do this..
```

fredrikaverpil:
> I think this can be simplified to:
> 
> `assert_raises(ImportError, __import__, "Qt")`

Indeed, thanks!

mottosso:
I see no error. I'm happy for you to merge. :)

fredrikaverpil:
> I see no error. I'm happy for you to merge. :)

Great!

When there are this many commits, do you just merge or do you squash first?

mottosso:
Good question..

Part of me feels would like to keep the history, but perhaps that's the hoarder in me. A single commit, for such a small and singular-purpose PR, would make it a lot easier to spot when browsing through history.

How about, when the PR itself is singular, like this one, squashing is fine.

On the other hand, when there is a commit that does *more than one thing*, we'll prefer keeping the commits.

I'm mainly interested in the cherry-pickability of commits, and if there are more than one thing happening in a single commit, things get messy.

fredrikaverpil:
That sounds good. I'll squash this time, as I don't know of any other way to not make this PR's commit history into a big mess in case you want to review the changes later.

mottosso:
Now that you've merged, there's one last thing that I have been thinking about how to articulate. About the test, `test_sip_api_pyqt4` and why I think it's unnecessary.

As a rule of thumb, a test should be implemented *before* the feature or bug it is testing. That's the pinnacle of "test-driven development" and it's a good measuring stick for what goes and what doesn't.

In this case, there is no way you could have written that test before the feature, because the feature itself is outside the scope of this project, and this project can do nothing to affect the outcome of this test.

Here's what tests do for me.

- Help articulate a feature, by writing the interface and use cases for it before implementing it.
- Ensure that a bugfix actually fixes a bug, by capturing before fixing.
- Keep an eye on code that changes, such that intended behavior remains the same.

This test does nothing but confirm what we already know, like how the earth is round, and can never fail regardless of what we do. It's a plain statement of fact. This makes it a needless test.

Let me know what you think.

fredrikaverpil:
Oh, crap. I intended to leave that out  I removed it one point but reverted it because I thought of it as a good way to rule out that there was an isolation problem. Then I forgot to remove it again.

I'm with you 100%. In my defence, I would have never even written that to being with unless Justin had mentioned it.

PR #62, opened 2016-06-05
Travis testing Python versions 2.6-3.5

fredrikaverpil:
Let's see what Travis says about this...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ sudo: required
 dist: trusty
 
 python:
+  - 2.6
   - 2.7
   - 3.4
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ services:
 env:
   QT_VERBOSE=true
 
+virtualenv:
+  # apt-install ends up amongst system packages
+  system_site_packages: true
+
 install:
   # Based on http://stackoverflow.com/a/24545890/478949
   - sudo apt-get install -y qt4-qmake libqt4-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,12 @@ dist: trusty
 python:
   - 2.6
   - 2.7
+  - 3.0
+  - 3.1
+  - 3.2
+  - 3.3
   - 3.4
+  - 3.5
 
 services:
   - docker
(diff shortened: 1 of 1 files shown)


mottosso:
Feel free to add 3.0, 3.1, 3.2, 3.3 and 3.5 for completeness.

That way, we can say `Qt.py supports Python 2.6-3.5`.

mottosso:
Hm, not sure what's going on there with 2.6..

fredrikaverpil:
I think it's related to the system wide site-packages setting:
https://github.com/travis-ci/travis-ci/issues/2219#issuecomment-41804942

I'll try that workaround. What an ugly one...

mottosso:
That was needed with Ubuntu 12.04, I'd expect it to be the same here, but perhaps not.

The reason it's there is because apt-get installs to the system-wide site-packages directory, but the mandatory Travis virtualenv doesn't include them, unless you tell it to.

I dunno, looks like a temporary bug, I wouldn't sweat it and wait till tomorrow or so.

fredrikaverpil:
[Here's another reference](https://github.com/travis-ci/travis-ci/issues/4260) to this (open issue since a year back).

But yeah, let's wait and see.

mottosso:
Maybe let's just find a way to have it work with Docker.

The only problem I have with [this](https://github.com/mottosso/Qt.py/pull/59/commits/e4cbdbe00c6fb7003b29233922a1c51a25feb6b3) is..

1. Maintaining one Dockerfile per version of Python.
2. The time it takes to build a Dockerfile (~2 minutes), relative how long it takes without it (~30 seconds).
3. And not being able to do it simultaneously, like we can here with many versions of Python.

If we can solve (1) and (3), (2) won't matter as much, and we'd gain the additional advantage of a 1-1 build between Travis and local.

Alternatively, we could just stop here, test against 2.7 and 3.4 (which already works) and call it done. We aren't gaining *that* much for being this meticulous.

fredrikaverpil:
> Alternatively, we could just stop here, test against 2.7 and 3.4 (which already works) and call it done. We aren't gaining that much for being this meticulous.

Must say I'm all for this option, as we don't really have a problem today. I've tested Qt.py in all scenarios where I need it to work. And it works fine.

Let's close this for now.

PR #67, opened 2016-06-14
Fixes #66 - Also copy binding module to __name__ in sys.modules

### LINKED ISSUE #66: Support alternate install locations / import paths of Qt module
`Qt.py` wants to replace itself with the resolved Qt bindings, by updating `sys.modules`. But it uses a hard-coded "Qt" string instead of `__name__`. This causes it not to work in my case where I am currently vendoring the `Qt.py` dependency into my project.

My import path looks something like this:

```python
from project.packages import Qt
```

I had to update the locations that did this:

```python
sys.modules["Qt"] = available[preferred]()
```

With

```python
sys.modules["Qt"] = available[preferred]()
sys.modules[__name__] = sys.modules["Qt"]
```

justinfx:
This merge request fixes #66 by also setting the imported bindings module to the path of the original `__name__`, and not just "Qt"

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -182,8 +182,6 @@ def _init():
     """
     
     # Save reference to module list
-    _mods = sys.modules
-    
     preferred = os.getenv("QT_PREFERRED_BINDING")
     verbose = os.getenv("QT_VERBOSE") is not None
 
@@ -191,7 +189,7 @@ def _init():
 
         # Debug mode, used in installer
         if preferred == "None":
-            _mods[__name__].__wrapper_version__ = __version__
+            sys.modules[__name__].__wrapper_version__ = __version__
             return
 
         available = {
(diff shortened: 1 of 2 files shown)


mottosso:
Thanks! Could you take a peek at the errors on Travis? We'll need a test for this as well.

justinfx:
I will have to find time to write tests. But I pushed a fix that made travis succeed. I still don't know why yet...but for some reason the `sys` module becomes `None` after the first time you updated it with "Qt".

Qt.py
```diff
@@ -180,7 +180,8 @@ def _init():
     this has executed.
 
     """
-
+    
+    # Save reference to module list
```
mottosso: Seems this line isn't necessary anymore.

mottosso:
Awesome, could you take a final look around to clean things up, and if it works for you I'm happy to merge.

justinfx:
Notes addressed.

mottosso:
Thanks!

Issue #68, opened 2016-06-17
Remaps: QItemSelection, QItemSelectionModel


fredrikaverpil:
I'm finishing up some stuff and just quickly jotting this down so I can adress this later on...

**QItemSelection**
- PyQt4: [QtGui.QItemSelection](http://pyqt.sourceforge.net/Docs/PyQt4/qitemselection.html)
- PySide: [QtGui.QItemSelection](https://srinikom.github.io/pyside-docs/PySide/QtGui/QItemSelection.html)
- PyQt5: [QtCore.QItemSelection](http://pyqt.sourceforge.net/Docs/PyQt5/api/qitemselection.html)
- PySide2: where are the docs?
- Qt5: [QtCore.QItemSelection](http://doc.qt.io/qt-5/qitemselection.html)

**QItemSelectionModel**
- PyQt4: [QtGui.QItemSelectionModel](http://pyqt.sourceforge.net/Docs/PyQt4/qitemselectionmodel.html)
- PySide: [QtGui.QItemSelectionModel](https://srinikom.github.io/pyside-docs/PySide/QtGui/QItemSelectionModel.html)
- PyQt5: [QtCore.QItemSelectionModel](http://pyqt.sourceforge.net/Docs/PyQt5/api/qitemselectionmodel.html)
- PySide2: where are the docs?
- Qt5: [QtCore.QItemSelectionModel](http://doc.qt.io/qt-5/qitemselectionmodel.html)

So it seems we should be able to do:

```python
def _pyqt4():
    # Remap
    PyQt4.QtCore.QItemSelection = PyQt4.QtGui.QItemSelection
    PyQt4.QtCore.QItemSelectionModel = PyQt4.QtGui.QItemSelectionModel

def _pyside():
    # Remap
    PySide.QtCore.QItemSelection = PySide.QtGui.QItemSelection
    PySide.QtCore.QItemSelectionModel = PySide.QtGui.QItemSelectionModel
```

eoyilmaz:
Hey Fredrik,

I keep finding your stuff all over the internet while switching to Maya 2017 and PySide2 :)

Cheers,

fredrikaverpil:
Hi Erkan :)
There's more work than usual with Maya 2017. Within a few weeks I'll put the new Render Setup to the test and I plan to write some about my findings on that too.

PR #69, opened 2016-06-19
Add remap for QItemSelection and QItemSelectionModel

fredrikaverpil:
- Fix https://github.com/mottosso/Qt.py/issues/68
- Version bump to 0.3.2

Do we add tests for remaps?

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -21,7 +21,7 @@ Usage:
 import os
 import sys
 
-__version__ = "0.3.1"
+__version__ = "0.3.2"
 
 
 def _pyqt5():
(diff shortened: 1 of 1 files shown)


mottosso:
> Do we add tests for remaps?

That could be challenging..

If there was some way of getting a list of members from each submodule, such as `QtCore`, then we could run that against what is currently available in Qt.py to see what's there but shouldn't, along with what is missing.

The "getting of the list" could be live, coming from either Qt source, or documentation, during the test itself and in that way we'd always stay up to date.

fredrikaverpil:
> getting a list of members from each submodule, such as QtCore, then we could run that against what is currently available in Qt.py

That's an interesting idea. I think it would take quite a bit of research, at least on my part, so let's not try to achieve that for this PR. 😉

mottosso:
> so let's not try to achieve that for this PR.

Certainly not, was just spawning ideas. It will eventually be required, I'm sure. Maybe not this approach, but somehow testing that our remaps are indeed correct and consistent.

PR #70, opened 2016-06-19
Remove whitespace on blank lines

fredrikaverpil:
Sorry if this comes around as a little silly, but I like to keep things tidy 😉


mottosso:
Not silly at all, did not see this!

Issue #74, opened 2016-06-26
Caveats tests


mottosso:
### Goal

Ensure caveats are actually caveats, from first writing through future changes.

### Implementation

Automatically test documented caveats. We could have a look at running the examples we type into CAVEATS.md, and seeing whether they produce the expected output or throws the expected errors.

We'd have to be more careful in how we format it, but in general should be able to conform to the doc-test format.

**Before**

```python
# PySide
>>> QtGui.QItemSelection.empty()

# PyQt4
>>> QtGui.QItemSelection.isEmpty()
```

**After**

```python
# PySide
>>> from Qt import QtCore
>>> assert not hasattr(QtCore.QItemSelection, "isEmpty")
>>> assert hasattr(QtCore.QItemSelection, "empty")
```

```python
# PyQt4
>>> from Qt import QtCore
>>> assert hasattr(QtCore.QItemSelection, "isEmpty")
>>> assert not hasattr(QtCore.QItemSelection, "empty")
```

Where each block is run, like other tests, in an independent process, using the binding specified on the first line.

mottosso:
Instruction on how to contribute new caveats with tests, see [intro of CAVEATS.md](https://github.com/mottosso/Qt.py/blob/fdf573cdb19a94f56e51ff433614658c0d7d84fc/CAVEATS.md).

PR #75, opened 2016-06-26
Implement #74, CAVEATS tests

mottosso:
Here's a suggestion for how we can both format and automatically test the caveats in CAVEATS.md.

In a nutshell:

1. For each pull-request..
  2. CAVEATS.md is parsed and converted into a Python module
  3. The module contains a doctest per Python block, i.e. \`\`\`python  code here \`\`\`
  4. The tests are included in the nose unittesting

At the moment, I've only formatted one of the caveats to actually work. The rest will need formatting too.

I've added the rules under which each caveat must be formatted here.

- [Rules](https://github.com/abstractfactory/Qt.py/blob/23dabcb02f2b13d5220364e133a4f7f9bc84571a/CAVEATS.md)

Let me know what you think.


fredrikaverpil:
Nicely done. I really like how these rules are dictating how the CAVEATS.md can grow and will be easily appended by anyone who just follows the rules – without us having to spend time on verifying the caveats manually. It's like killing two flies with one clap (or however that translates into English).

My only comment to that is that the example code could get a little more difficult to read and we might want to require a few lines of text describing the caveat with human-readability in mind.

I say let's implement this!

mottosso:
> My only comment to that is that the example code could get a little more difficult to read and we might want to require a few lines of text describing the caveat with human-readability in mind.

Agreed. Let's try and ensure that by eye in PRs.

mottosso:
@fredrikaverpil could you have a look at formatting the caveats such that it passes the tests?

I've updated the README with how to run those tests.

- https://github.com/abstractfactory/Qt.py#developer-guide

fredrikaverpil:
Sure. Can I somehow take over the PR or should I create a new one with your modifications so far?

mottosso:
You can pull from the source repo, [here](https://github.com/abstractfactory/Qt.py), and pull-request to it. When I merge there, this one will update like usual. :)

mottosso:
Aha, now here's an interesting artifact. Perhaps not unexpected. But bindings under Python 2 can behave differently from 3.

- https://travis-ci.org/mottosso/Qt.py/jobs/140401297#L870

Possibly just a notation issue, but I'd expect more fine differences to appear as we go along. I'd suggest we include additional syntax to the doctest format.

```python
# <Binding>, <Interpreter version, optional>
>>> Test here..
```

For example..

```python
# PySide, Python3
>>> print("Running under Python 3, using PySide")
```

```python
# PyQt4
>>> print("Running under Python 2, using PyQt4")
```

Where Python 2 is the default, unless 3 (even a particular minor version, such as 3.3?) is specified.

fredrikaverpil:
> bindings under Python 2 can behave differently from 3.

Argh.. Yeah, that makes sense to use `# PySide, Python3` for example.

fredrikaverpil:
I'm going to go ahead and make a PR with what I have currently.

- I haven't looked at making tests for "Closures". @justinfx could you provide us with at working example?
- I can't get PyQt4 to throw a `TypeError` with `QtGui.QRegExpValidator`. @justinfx do you think this really was an issue?
- No tests in Python 3

mottosso:
Go right ahead, I'll merge into abstractfactory so that it updates here, and we'll continue from there.

fredrikaverpil:
Quoting @mottosso from [my fork/PR](https://github.com/fredrikaverpil/Qt.py/commit/6080bae300c1f309ef2d46ecbdd4bf40dae19353)...

> Hmm.. I'm not sure about the assert True and assert False.

It's just the very last test now that has that. Haven't had time to look into it further yet.

CAVEATS.md
```diff
@@ -35,20 +55,26 @@ NameError: free variable 'self' referenced before assignment in enclosing scope
 <br>
 
 
-#### QtCore.QAbstractModel.createIndex
```
fredrikaverpil: It's possible that this was a mistake on my part.

CAVEATS.md
```diff
@@ -94,7 +138,19 @@ In PySide, the constructor for `QtGui.QRegExpValidator()` can just take a `QRegE
...
+>>> assert QtGui.QRegExpValidator(regex, None)
+>>> assert QtGui.QRegExpValidator(regex)
+```
+
+```python
+# PyQt4
+>>> from Qt import QtCore, QtGui
+>>> regex = QtCore.QRegExp("\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}")
+>>> assert QtGui.QRegExpValidator(regex, None)
+>>> assert_raises(TypeError, QtGui.QRegExpValidator, regex)  # Seems this works fine in PyQt4?
```
fredrikaverpil: @justinfx this is the one which doesn't raise the `TypeError`.

mottosso:
Awesome! So close! I'll add the Python chooser syntax tomorrow.

fredrikaverpil:
Nice :) Good night!

CAVEATS.md
justinfx: Which version of PyQt4 are you testing? Under PyQt4 4.8.5 / Qt 4.8.5, it produces a TypeError for me, because the second argument (the parent) doesn't use a default value. I would have to try it on a newer PyQt4 when I am not at work.

fredrikaverpil: > Which version of PyQt4 are you testing?

Hm. Actually, when glancing through the Travis build, there's no trace of an exact PySide or PyQt4 version anywhere.

@mottosso we should probably write that out somewhere (the exact version of the bindings used in the testing), perhaps via the new `build_caveats_tests.py`?

mottosso:
Considering that the tests work except for just `1L`, maybe a Python specifier isn't necessary. Instead maybe we could do..

```python
# PySide
>>> from Qt import QtGui
>>> model = QtGui.QStandardItemModel()
>>> index = model.createIndex(0, 0, -1)
>>> assert int(index.internalId()) == -1
```

fredrikaverpil:
If that works around the problem for now, let's not over-engineer this. Your suggestion works for me.

fredrikaverpil:
One quick question though, should we use `assert_equals` in that example?

fredrikaverpil:
To my understanding, `assert` will only throw "AssertionError" while `assert_equals` will present a more fine detailed error.

mottosso:
`assert_equals` is a nose function, it isn't in the standard library. We could use that, but I think it might complicate the rules for the CAVEATS doctests.

You're right, it does produce better error messages. But we can add our own message to `assert`, which I think it better still. As then it's part of the documentation. (What else should be included?)

```python
>>> assert int(index.internalId()) == -1, "internalId() should have returned -1"
```

That message is then included in the message when the assertion fails, as though you did this.

```python
raise AssertionError("internalId() should have returned -1")
```

mottosso:
I can see you're using `assert_equals` currently, I'm surprised that works. It isn't imported in the generated test module. Must be nose performing some magic under the hood!

It does look clear, and if that works without intervention, then I think it's fine. The alternative is a series of try/except which is more difficult to read.

fredrikaverpil:
But, do you really think that we shouldn't just use what's already there for us with nose?
We're already relying on it for the docker tests.

> It isn't imported in the generated test module

It's imported [here](https://github.com/abstractfactory/Qt.py/blob/9c688d26b46ac708f111bef20c430c5176785dea/build_caveats_tests.py#L70).

fredrikaverpil:
Umm... I'm not using `assert_equals`, but I'm using `assert_raises` already.

fredrikaverpil:
I couldn't figure out a clean way to assert True for TypeError, so that's why I used assert_raises. In my case, I believe I would have needed a lot more (and a little confusing try/except) code. So even if it seems it may complicate things, it actually makes for shorter and simpler examples.

mottosso:
> It's imported here.

Aaah, now I see. That's a clever way of including it! Ok, then we need to add this to the rules of the doctests at the top of CAVEATS, that this function is available to use. Along with any other additional functions are included.

> I'm using assert_raises already.

Sorry, that's what I meant.

fredrikaverpil:
> Ok, then we need to add this to the rules of the doctests at the top of CAVEATS

Yes, definitively. Sorry, forgot to bring that up yesterday in my little bullet list.

Also, recapping a few other things that might need attention:
- I probably renamed [the **QtCore.QAbstractItemModel.createIndex** header](https://github.com/abstractfactory/Qt.py/blob/9c688d26b46ac708f111bef20c430c5176785dea/CAVEATS.md#qtcoreqabstractitemmodelcreateindex) in error.
- Seems most recent PyQt4 versions doesn't raise the error @justinfx is seeing on his end
  - Can we deal with such issues?
  - We should print out the Qt.__binding_version__ somewhere so it's visible in the tests
- We need working examples for closures.

mottosso:
> Can we deal with such issues?

Spontaneously, we could start testing against multiple versions of each binding. But that could easily get out of hand.

We could either find an older version and test against that, assuming later versions don't break backwards compatibility, or say screw it and only support the latest (or whichever is available to Ubuntu 14.04).

> We should print out the Qt.binding_version somewhere so it's visible in the tests

We could, but there should only be one binding available to Ubuntu 14.04, and it looks like 4.8.0-1.

- http://packages.ubuntu.com/trusty/python-qt4

That said, we could still print that at the top of the Travis test, for completeness. I'd add to to the top of the `script:` section of `.travis.yml`.

fredrikaverpil:
> Spontaneously, we could start testing against multiple versions of each binding. But that could easily get out of hand.

Yeah, that's what I was thinking too. I don't know... then we've got testing on Windows and macOS around the corner as well...

> I'd add to to the top of the `script:` section of `.travis.yml`.

Sounds good!

mottosso:
> then we've got testing on Windows and macOS around the corner as well...

Travis can do Mac too and there's AppVeyor which is the Windows equivalent. I just don't want us to end up testing a binding, rather than testing Qt.py. I don't want any part in testing differences in the bindings themselves across platforms, that's for The Qt Company and Riverbank Software to do.

mottosso:
Ok, so who's doing what?

I can add to `.travis.yml` and fix the Python 3 issue, then we'll await Justin after which you'll implement the final test?

fredrikaverpil:
> Travis can do Mac too and there's AppVeyor which is the Windows equivalent. I just don't want us to end up testing a binding, rather than testing Qt.py. I don't want any part in testing differences in the bindings themselves across platforms, that's for The Qt Company and Riverbank Software to do.

Agreed! Let's just not go down that path at this point in time. So what will we do about the second "QtWidgets.QAction.triggered" test which doesn't raise the `TypeError`?
Do we remove it or do we add this into a section of the CAVEATS.md which isn't parsed and tested?

> Ok, so who's doing what?
> 
> I can add to `.travis.yml` and fix the Python 3 issue, then we'll await Justin after which you'll implement the final test?

Sounds great. I did try to implement a closures example yesterday, but I found Justin's example a little hard to follow. So I'll just wait in case Justin can prove an example 😄

mottosso:
Ok, the ball is in your court!

fredrikaverpil:
Awesome!

I'm noticing that PyQt4 is of version 4.10.4. So that would explain the issues @justinfx are seeing with PyQt4 4.8.4.

mottosso:
> I'm noticing that PyQt4 is of version 4.10.4. So that would explain the issues @justinfx are seeing with PyQt4 4.8.4.

Aah, there we go. Could you have a look at whether it's possible to get an older version on Ubuntu 14? Preferable the oldest one available.

fredrikaverpil:
Should we create a new "Fixed caveats" headline of CAVEATS.md, where we mention this issue and make sure that a code snippet isn't being tested?

mottosso:
Fixed, as in, the caveat is only an issue in a particular version of the binding?

Yeah, if we test against the oldest version, maybe we could add something like..

### My header

| Affects      | Version
|:--------------|:---------
| PyQt4       | <= 4.8.2
| PySide      | ==1.1.0

My description

```python
# PySide
>>> from Qt import QtCore
>>> my test
```

fredrikaverpil:
> Could you have a look at whether it's possible to get an older version on Ubuntu 14? 

Looking now...

> Yeah, if we test against the oldest version, maybe we could add something like..

I like that. Let's do that.

fredrikaverpil:
Seems like 4.10.4 is the only one available:

```bash
root@929d1c4cf6d5:/# apt-cache policy python-qt4
python-qt4:
  Installed: (none)
  Candidate: 4.10.4+dfsg-1ubuntu1
  Version table:
     4.10.4+dfsg-1ubuntu1 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
```

mottosso:
> Seems like 4.10.4 is the only one available:

It doesn't have to be via apt-get.

fredrikaverpil:
> It doesn't have to be via apt-get.

I'm not very familiar with how Ubuntu deals with different releases and its apt-get eco system. I found these old .deb files: http://ftp.us.debian.org/debian/pool/main/p/python-qt4/

However, after having attempted to `dpkg -i python-qt4_4.9.3-4_amd64.deb` (detects loads of missing dependencies) followed by `sudo apt-get -f install` (downloads and install the recently detected missing dependencies along with PyQt4) ... I ended up with having PyQt4 4.10.4 installed.

fredrikaverpil:
Here are the oldest sources I can find: https://download.qt.io/archive/qt/4.8/

However, I don't think it's viable that we attempt to build from source.

mottosso:
We could compile it ourselves each time we test, but that adds ~20 minutes of testing time and adds the potential of tests failing due to taking too long (Travis has a timeout).

We could compile it ourselves, and upload a Python wheel somewhere. Then we could install using this wheel in all of our tests.

- [Reference](http://stackoverflow.com/a/24545890/478949)

fredrikaverpil:
> We could compile it ourselves, and upload a Python wheel somewhere. Then we could install using this wheel in all of our tests.

That's probably the only reasonable way out of this.

With SIP v2, we basically said Qt.py doesn't support PyQt4 < v4.6. So we could either go for 4.8.something, which seems the most widespread ... or go as low as 4.6.

fredrikaverpil:
Actually, it looks like we can't go further back than PyQt-4.9.4, since there's no older source.

fredrikaverpil:
I'm checking with the riverbank mailing list to see if we can get a hold of the 4.6 or the 4.8.4 PyQt source.

mottosso:
> Actually, it looks like we can't go further back than PyQt-4.9.4, since there's no older source.

Haha, that's hilarious.

@justinfx from which tomb did you dig up this version of yours??

justinfx:
On Tue, 28 Jun 2016, 6:04 AM Marcus Ottosson <notifications@github.com>
wrote:

> Actually, it looks like we can't go further back than PyQt-4.9.4, since
> there's no older source.
>
> Haha, that's hilarious.
>
> @justinfx <https://github.com/justinfx> from which tomb did you dig up
> this version of yours??
>
We have an internal packaging and environment management system. So many
things are still available. Might be because it is highly compatible with
older Maya installsinstalls? I don't know. But obviously it isn't the
newest. If you want to draw the line at the oldest version available in
Ubuntu 14.04, then that would make sense.

fredrikaverpil:
@justinfx the oldest version in Ubuntu 14.04 would be the only one available; v4.10.4. Also, would you be able to share a somewhat more complete examples for the [closures caveat](https://github.com/abstractfactory/Qt.py/blob/6b32369b3f6e2b78da914ab2774726e65c7d37f0/CAVEATS.md#closures)?
I'm having troubles making that into something reproducible.

justinfx:
I'm having trouble reproducing it now. Maybe just strike it from the list.
If I can get a more complete working example again, I will update.

fredrikaverpil:
> I'm having trouble reproducing it now. Maybe just strike it from the list.
If I can get a more complete working example again, I will update.

Ok, thanks. [I made a PR](https://github.com/abstractfactory/Qt.py/pull/2) yesterday where I took it out for now. The PR [passes the tests](https://travis-ci.org/fredrikaverpil/Qt.py/builds/140665851).

fredrikaverpil:
I also [attempted to build wheels](https://gist.github.com/fredrikaverpil/52dc5b2cdcb17001c7b0c13bdba4d494#file-ubuntu_14-04_compile_and_install_pyqt4_and_pyside-md), but I don't think it's possible to build wheels for PyQt4.

I got response yesterday [[1](https://www.riverbankcomputing.com/pipermail/pyqt/2016-June/037681.html)] [[2](https://www.riverbankcomputing.com/pipermail/pyqt/2016-June/037682.html)] on the Riverbank mailinglist regarding how to get access to older PyQt4 sources. However, I'm not so sure about us compiling from source. I'd rather just run the tests from time to time and keep CAVEATS.md updated accordingly.

If we were to compile from source, we should try to avoid to compile Qt4 from source as that takes a very long time. Perhaps `sudo apt-get install -y qt4-qmake libqt4-dev` is enough and PyQt4 can link against that. However, I cannot figure out which sip version should go with PyQt4 4.8.4. @justinfx Do you know which sip version you're on?
Here's what I have so far:

```bash
docker run -ti ubuntu:14.04 bash
apt-get update && apt-get install -y wget qt4-qmake libqt4-dev python-dev python-pip
wget https://sourceforge.net/projects/pyqt/files/sip/sip-4.13.3/sip-4.13.3.tar.gz/download
mv download sip-4.13.3.tar.gz
tar -xvf sip-4.13.3.tar.gz
cd sip-4.13.3
make
make install
cd ..
wget ftp://ftp.csc.fi/.m/mirrors1e/ftp.netbsd.org/pub/NetBSD/packages/distfiles/PyQt-x11-gpl-4.8.4.tar.gz
tar -xvf PyQt-x11-gpl-4.8.4.tar.gz
cd PyQt-x11-gpl-4.8.4
python configure.py --qmake=/usr/bin/qmake  # Errors out because of wrong sip version?
# make
# make install
```

mottosso:
Great, great work, Fredrik. Having a closer look.

mottosso:
> I'm not sure if or how you could make PySide wheels completely standalone. They need to link against compiled Qt4.

Maybe that's ok. We can still install Qt4 via apt-get, and use a wheel of just the bindings. It'd mean two small wheels, one for PyQt4 and one for PySide, that reuse the same binaries. As opposed to having two duplicate binaries. Maybe it's preferable even.

fredrikaverpil:
> Maybe that's ok. We can still install Qt4 via apt-get, and use a wheel of just the bindings. It'd mean two small wheels, one for PyQt4 and one for PySide, that reuse the same binaries. As opposed to having two duplicate binaries. Maybe it's preferable even.

That sounds nice. But I doubt it's going to be easy to build PyQt4 wheels, if at all possible.

mottosso:
> That sounds nice. But I doubt it's going to be easy to build PyQt4 wheels, if at all possible.

It doesn't have to be wheels either. Basically anything goes where we can speed up the install of PyQt and PySide. The most straightforward, and perhaps least attractive, option is to simply compile it, and upload it to something like Dropbox. Tests could then download it from there.

There are equivalent services, like BinTray, that we might be able to use. Wheels have the advantage of looking nice, via the `pip install` syntax. But as long as it looks clear, and is hard to break, anything goes.

mottosso:
Ok, I've pushed a few changes that I think makes the tests more readable and informative.

1. No more assertions. Assertions hide information from the written CAVEATS.md document that would be interesting, such as return values, exception type and message.
2. Every call now visualises the value it produces. For example, I noticed that `emit()` in PySide returns a boolean value, where PyQt4 does not. The doctests can tell us this.
3. I removed the Fixed caveats, I don't think we need to bother including them.

Let me know what you think.

mottosso:
I went ahead and implemented the Python chooser syntax, because even exceptions print different things under Python 2 and 3 (string versus bytes)..

Let me know what you think

fredrikaverpil:
Now this is what I would call black magic!
I'm impressed. Very, very nice!

I need to dig deeper into `build_caveats_tests.py` to see exactly how you do this, as this is plain awesome. 👍

mottosso:
Well I can't take much credit, most of it is just good old doctest. If what you're thinking of is the checking and comparing of output?

fredrikaverpil:
Ah. I've never used doctest. Nevertheless, very nicely implemented. I should look into doctest for my own projects.

I see you kept the "Fixed caveats" in the TOC at the top of the CAVEATS.md. And on that note, don't you think it's useful to keep fixed caveats documented?
I can see at least two pros:

- Primary: Users can see they should upgrade to a newer release of their binding (very useful info IMHO). Personally, I would subscribe to to changes on this page to stay up to date.
- Secondary: Avoid spending time making a PR with a caveat example/doctest which is fixed in the binding versions we're testing with.

I really don't see any cons with keeping them in CAVEATS.md although in an ideal world I wanted to see them actually being a caveat during tests, but I of course trust Justin here.

If we were to keep fixed caveats documented, it would make sense to also keep the table where you show in which version it was a problem or up to which version it was a problem. But I only think we need that table for fixed caveats. No need to have the tables for current caveats.

Perhaps we should call them "Current caveats" and "Past caveats"?

fredrikaverpil:
Regarding testing using specific versions of PySide/PyQt4, do you feel we really need that?
To me this sounds like it's going to be hard to maintain.

At what point do we "freeze" certain versions of a binding so that tests will stay using a certain version?
Perhaps this will just be done manually from time to time?

I wonder if we can somehow get notified on a new binding version release, so that we can trigger a new test?

mottosso:
> And on that note, don't you think it's useful to keep fixed caveats documented?

I suppose you're right.

I think my main gripe was with the title, as "fixed". It sounded as something we've fixed. Maybe we can find a way to integrate them better with the surrounding caveats.

We could do something like..

```python
# Untested
>>> my doctest
```

For tests that don't apply to the version we're testing with, and add a..

| Affects | Version
|:---------|:---------
| PyQt4 | =<4.8.1

To the top.

However, I don't want contributors to apply `# Untested` to each of the doctest they fail to reproduce.. Maybe that's the wrong approach to deal with it?

> Regarding testing using specific versions of PySide/PyQt4, do you feel we really need that?

No, I think we should just find an older version, and stick with that (forever). And trust newer versions retain backwards compatibility.

I'd rather not go for whatever is currently available in whatever OS we happen to use. I did not choose Ubuntu 14.04, it's simply what is available today. It might not be available next year, or the year after that, in which case we'll have to update our toolchain and break our tests.

That's why I think a dedicated wheel or apt-get package is the way to go.

mottosso:
> Could we try and avoid that and make the separator be the same on Linux/macOS/Windows?

I think we should stick to convention. `os.pathsep` isn't used just for paths, but as a generic list separator.

Aside from PATH, here's one more example http://environmentvariables.org/PathExt.

fredrikaverpil:
> Maybe we can find a way to integrate them (untested caveats) better with the surrounding caveats.

I think we should be very clear on which caveats are "current" and are being tested - and which caveats are of status "past caveats" or whatever. So I'd propose we group all "current caveats" and then we group all "past caveats" so it becomes clear. Kind of like I did initially in the table of contents. I don't think it's really useful to mix them without grouping.

Apart from this comment, I agree with your suggestions on `# Untested` and the versions table. I kind of agree with you that `# Untested` will make it easy to skip defining the doctest properly. If we put all "past caveats" under its own headline, we could just skip all tests beneath this headline. This is what I did in my previous PR. It's your call.

> No, I think we should just find an older version, and stick with that (forever). And trust newer versions retain backwards compatibility.
> ...
> That's why I think a dedicated wheel or apt-get package is the way to go.

Agreed. The only wheel out there which really works is for PyQt5 and Python 3. I suppose we'll need to look to conda or apt-get for the others.

fredrikaverpil:
> I think we should stick to convention. os.pathsep isn't used just for paths, but as a generic list separator.
> 
> Aside from PATH, here's one more example http://environmentvariables.org/PathExt.

I guess you're right, principally.

This means I'll have to do something like this:

```python
os.environ['QT_PREFERRED_BINDING'] = 'PySide' + os.pathsep + 'PyQt4'
```

...which is totally fine.

mottosso:
> I think we should be very clear on which caveats are "current" and are being tested - and which caveats are of status "past caveats" or whatever.

Maybe let's take a step back from testing and completeness and think about how CAVEATS is meant to be used.

To me, it'll be a reference for problems I have encountered while developing. I might run into an error when using `QtCore.Slot`, so I'd head to CAVEATS.md to (1) see if there are any notes for it. If there are, I'd be interested in (2) what to look out for, aside from the exact problem I was having (this is where the "affects" and versioning comes in), along with (3) a potential solution or workaround.

I think what I would not expect, is to have a section on `QtCore.Slot` with problems current to whatever version we use to test with (implementation detail) and another section for problems that fall outside of those tests.

Odds are that version won't match my version anyway, like in Justin's case.

I think we should aim to fulfil (1), (2) and (3) with CAVEATS.md.

What usecase do you have in mind for it? At what point would you head to the bottom of the page to look for fixed problems?

fredrikaverpil:
I agree with your use case but wish to append 4):

4) if you're on an older binding version than what is currently being tested by Qt.py, it is useful to be able to find information in CAVEATS.md which are true to the version you are currently on.

Just like in Justin's case, the Qt.py project may not be using the same binding versions end-users are using for a particular project. If you're on an older version of e.g. PyQt4, you want to contribute and document issues which only apply to a certain version or versions.

With time, it makes sense to me to keep whatever it is we put into CAVEATS.md and file it under some kind of "past caveats" section or whatever, if a caveat no longer passes our tests (to be a caveat).

Does that make sense?

mottosso:
I think (4) is the same as (2). To know what to look out for, in which case we can include the "affects" section.

> and file it under some kind of "past caveats" section

But it's "past" or "fixed" relative the version we use for testing. That version is an implementation detail. Users of Qt.py won't and shouldn't know nor care about what we use to test. They'll use a particular version, and care about whether a particular caveat applies to that particular version.

Maybe we're saying the same thing, here's what I'm suggesting.

---

#### QtCore.QItemSelection

Description here.

```python
# PySide
>>> doctest here
```

<br>

#### QtGui.QRegExpValidator

| Affects       | Version
|:--------------|:-----------------
| PyQt4         | <= 4.8.4

Description here

```python
# PyQt4, untested
>>> doctest here
```

---

Points:

1. The "affects" table is present only when the problem is limited to a particular version
2. The caveat from a particular version doesn't get special treatment, but is amongst other caveats.

fredrikaverpil:
> Maybe we're saying the same thing, here's what I'm suggesting.

I think we are saying the same thing. I like your example. 👍

mottosso:
Ok, let me know what you think.

mottosso:
I split parsing functions into it's own module, intend on testing that independently as the number of features and keywords increase. It's quite forgiving at the moment, but I suspect we might run into cases where we might want even more freedom in formatting the CAVEATS.md.

mottosso:
Merging this beast, any problems, let's make it a new PR or issue.

fredrikaverpil:
Great job, looks good! :)

PR #77, opened 2016-07-19
Add who's using Qt.py

mottosso:
Word gets around. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,15 @@ None yet.
 <br>
 <br>
 
+### Who's using Qt.py?
+
+Send us a pull-request with your studio here.
+
+- Framestore
+- Weta Digital
+- Disney Animation
+- Industriromantik
+
 ### Projects using Qt.py
 
 Send us a pull-request with your project here.


fredrikaverpil:
Awesome!
LGTM

mottosso:
I think we might have hit a bubble with this project, now that people are updating to Qt 5, but still need to maintain compatibility with Qt 4. It's possible this project could become *the* way to do that if we play our cards right.

fredrikaverpil:
Still on vacation with just my phone, so I haven't kept up with any developments. I saw on the vfx platform mailing list about Disney, which was really cool.

I think that we'll get more attention as soon as Maya 2017 is out. We should put up a guide for it and how to use Qt.py in it, I think. I'm sure that alone would gain us some traction. Perhaps strike a deal with e.g. CGSociety or some other place which would help market such a guide directly and Qt.py indirectly. I'd be happy to write start writing a guide on that.

mottosso:
> We should put up a guide for it and how to use Qt.py in it, I think.

I think that's a good idea!

fredrikaverpil:
I've written a very basic introduction on how to take your old Python/PySide scripts and make them run in Maya 2017 with the help of e.g. Qt.py. It's focusing more on what's happened since the last major release (Qt4 vs Qt5) rather than actual development with Qt.py. I'll publish this on [my blog](https://fredrikaverpil.github.io) as soon as Maya 2017 is officially released (I'm guessing sometime next week). At the end of the post, I explain what Qt.py is in a **very** simplistic manner. I'm not entirely sure if this post is worth posting a link to from the `README`. The aim of this blog post is to basically point out the problems Qt5 will cause studios who wants to move onto Maya 2017 and some ways of dealing with that.

I've intentionally left out Qt.py in-depth explanations or details such as installation in Maya from this blog post (it became quite long anyway). I'd like this blog post to refer to another follow-up post which would be more in-depth and which would be useful for people who have just installed Maya 2017 and realize their scripts don't work any more. It could perhaps be called "Getting started with Qt.py in Maya".

So, in this second post I think it would be helpful to show a couple of different installation methods (pip/virtualenv, dropped into Maya's installation directory) as well as show examples on how it can be used with both Python 2 and Python 3 outside of Maya with any of the bindings. Also mentioning that the (non DCC-specific) code you write with Qt.py will also run in Nuke (and other DCC applications) whenever they move to Qt5. Explanation on how it works with the remappings to increase understanding without getting too complex, how you can contribute, mention the caveats and its testing etc etc. All of this to show the openness of the project and the will to get this flying. I think this in-depth post should be written in a personal style, and from real experiences and from the Maya user perspective rather than be too generic.
I'd also like this in-depth post to include images and screenshots of UIs in and outside of Maya.

So, in a way, the strategy:

1. My first blog post to catch general interest in Qt.py
2. The second, more in-depth, post to get increased understanding of Qt.py and prepare for development with Qt.py
3. Then you use the Github repository while you're developing (issues, README, PRs)

I could just start writing and publish this second in-depth post on my blog but I'd like your feedback prior to publishing. Or we could write it together and in that case, perhaps it should be hosted in the project repository?

I've been thinking that perhaps we should put out a simple press release, referring these posts/guides. We could send the press release to Lesterbanks, CGSociety, VFXWorld...
But I'm not sure if those are the best outlets. Perhaps 3DPRO, Tech-Artists.org and the other dev mailing lists are better ways of getting the word out.

Let me know what you think about all of this!

fredrikaverpil:
I'm emailing you a link to where you can read that first blog post of mine.

PR #81, opened 2016-07-28
Add support secondary arguments support to load_ui

### LINKED ISSUE #80: Add support for additional arguments to load_ui
Per the python mailing list discussion, add support for the 2nd argument to uic.loadUi

Here is an example using PyQt4

http://nullege.com/codes/search/PyQt4.uic.loadUi

Here is an example of an implementation for PySide that accepts the same arguments

https://gist.github.com/cpbotha/1b42a20c8f3eb9bb7cb8

fredrikaverpil:
Fix #80 

We can now support the base instance argument and the custom widgets argument in `load_ui` thanks to the custom class in [this gist](https://gist.github.com/cpbotha/1b42a20c8f3eb9bb7cb8).

Let's discuss.

How do we deal with the copyright message at the top of the gist?
We should give Sebastian Wiesner (lunaryorn@gmail.com) credit for this.

@bpabel - do you know how to use the third (custom widgets) argument?
I've added support for it but I'm not sure about its use case. I'd like to see an example of it being used so that I can add tests for it.

@mottosso I was unable to remove QStatusBar and QMenuBar from the tests UI without crash.


bpabel:
In that example, it would be a dictionary of class name: class objects.  I think the most common use is to just import all the custom widget classes into your module and then just pass `globals()` in as the `customWidget` argument.  In the `.ui` file, you would use the *promote to* feature in Qt Designer to specify the class name for the custom widget.

There might be a better way to do this though, because `PyQt` doesn't require the customWidgets dictionary, and is able to pull the import path from the `.ui` file.  I believe there is a convention where you set the *header file* attribute to the import path for the class, and loadUi will try to load the custom widget class using that import path.  

I can test and make an example, but probably not until the weekend.

fredrikaverpil:
> I can test and make an example, but probably not until the weekend.

That would be great, thanks. 👍

mottosso:
What are our thoughts on this?

> Hi list,

> We've recently noticed that our current QUILoader implementation is
broken in subtle ways, and fixing the issue will in practice require a
complete rewrite of the functionality in Python.

> We're prepared to do the rewrite, but this will be post-1.0 content for
us. Hence, we recommend PySide developers to refrain from using
QUILoader for the time being. Instead, you can use pyside-uic to convert
the .ui files to Python and use them that way.

> Sorry for any inconvenience.

> Cheers,

> ma.

- https://groups.google.com/forum/?fromgroups=#!topic/pyside/_s1HPe6XTZs

fredrikaverpil:
> What are our thoughts on this?

Hm. That seems almost six years old and this is the first time I've seen this being mentioned. I wonder if we can find any info on whatever issues they're referring to ever got fixed. I'm noticing they refer to QUILoader while today this is actually called QUiLoader ... perhaps that's just a coincidence, or it was rewritten and renamed?

To be honest, I've always used pysideuic in the past although I'm using QUiLoader now (since I'm using the `load_ui` function of Qt.py). So far I haven't had any issues. The problem with pysideuic is that e.g. The Foundry decided to not bundle that module with Nuke. I contacted the Foundry and asked them to do that a long time ago, and their response was I should use QUiLoader. At the time I didn't do that and instead I made sure pysideuic was available to Nuke.

I'd like to propose the following: If we see issues with QUiLoader, we could make Qt.py use pysideuic first, if that is available, and if not Qt.py would fall back onto QUiLoader. But I'd like to give it some time to see if that's really necessary before implementing something like that. From what I've read on mailing lists and over at e.g. Stackoverflow, QUiLoader is the go-to mechanism to load UIs in PySide.

fredrikaverpil:
It seems at least some bug fixing activity was made after 2010:
* https://github.com/PySide/PySide/commits/master/PySide/QtUiTools/typesystem_uitools.xml
* https://github.com/PySide/PySide/search?utf8=✓&q=QUiLoader

In the Google Groups post, bug 552 was mentioned, which they have an old test for, created in 2012:
https://github.com/PySide/PySide/blob/bcf6058e9d0e8513f58a0f1062b428ea7eb03f0c/tests/QtUiTools/bug_552.py

Also, they say they added a warning to the documentation on QUiLoader:
https://groups.google.com/d/msg/pyside/_s1HPe6XTZs/WpmIsUeVmQ0J

This warning is nowhere to be seen today:
http://pyside.github.io/docs/pyside/PySide/QtUiTools/QUiLoader.html
... so I'm starting to lean towards that those issues were fixed.

fredrikaverpil:
> I can test and make an example, but probably not until the weekend.

@bpabel In case you have the possibility to write an example, that would be great. We've written our tests around `load_ui` like this:

```python
def test_load_ui_into_self_pyside():
    """load_ui: Load widgets into self using PySide"""

    with ui() as fname:
        with pyside():
            from Qt import QtWidgets, load_ui

            class MainWindow(QtWidgets.QMainWindow):
                    QtWidgets.QMainWindow.__init__(self, parent)
                    load_ui(fname, self)

            app = QtWidgets.QApplication(sys.argv)
            window = MainWindow()

            # Inherited from .ui file
            assert hasattr(window, "pushButton")
```

You can view all 4 tests [here](https://github.com/fredrikaverpil/Qt.py/blob/cae6aff042a102e6b187cf0ed2e25d742eeaf766/tests.py#L219-L282).

fredrikaverpil:
@mottosso For some reason, I'm getting issues with the docker tests with this PR. Do you have any idea why that is?

```bash
➜ ~/code/repos/Qt.py (feature/load_ui) docker run --rm -v $(pwd):/Qt.py mottosso/qt.py
Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
Setting QT_PREFERRED_BINDING to more than one binding excludes others ... ok
Preferred binding PyQt4 should have sip version 2 ... ok
Qt.py may be bundled along with another library/project ... ok
load_ui: Load widgets into self using PySide ... ERROR
load_ui: Load widgets into self using PyQt4 ... ERROR
load_ui: Load .ui file ... ERROR
load_ui: Load widgets into custom using PyQt4 ... ERROR
Raise ImportError if sip API v1 was already set (Python 2.x only) ... ok

======================================================================
ERROR: load_ui: Load widgets into self using PySide
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/local/lib/python2.7/dist-packages/nosepipe.py", line 179, in __call__
    (header + data).decode("latin1"))
Exception: short message body (want 1852797797, got 35)
Something went wrong
Message: nosetests: cannot connect to X server 


======================================================================
ERROR: load_ui: Load widgets into self using PyQt4
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/local/lib/python2.7/dist-packages/nosepipe.py", line 179, in __call__
    (header + data).decode("latin1"))
Exception: short message body (want 1852797797, got 35)
Something went wrong
Message: nosetests: cannot connect to X server 


======================================================================
ERROR: load_ui: Load .ui file
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/local/lib/python2.7/dist-packages/nosepipe.py", line 179, in __call__
    (header + data).decode("latin1"))
Exception: short message body (want 1852797797, got 35)
Something went wrong
Message: nosetests: cannot connect to X server 


======================================================================
ERROR: load_ui: Load widgets into custom using PyQt4
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/local/lib/python2.7/dist-packages/nosepipe.py", line 179, in __call__
    (header + data).decode("latin1"))
Exception: short message body (want 1852797797, got 35)
Something went wrong
Message: nosetests: cannot connect to X server 


Ran 22 tests in 3.105s

FAILED (errors=4)
```

mottosso:
> Do you have any idea why that is – and, do you see this on your end as well?

I remember you asking this elsewhere, and I was about to answer it there but couldn't find it. :S

Yes, I've got an idea, it's because for a QApplication to be instantiated, there needs to be an X-server running and we haven't started one in our Docker image.

On Travis, one is being setup [here](https://github.com/mottosso/Qt.py/blob/20247745907366495fc5b8146c7575c000de45cd/.travis.yml#L40). As Travis doesn't have access to a graphics card or display, this one is a virtual display, called Xvfb. We should be able to use this in Docker too.

fredrikaverpil:
Ah, I've been searching and I did find references to xvfb.

Do you think that's the best way forward?
I know it's possible (not how though) to run GUIs in Docker containers, somehow mapping the display. Perhaps that's too cumbersome and difficult and not cross platform?

fredrikaverpil:
Need to somehow suppress all that stuff Xvfb spews out.

This is from my OS X desktop, running Docker for Mac 1.12.0-rc-beta20:

```
$ docker run --rm -v $(pwd):/Qt.py mottosso/qt.py

/usr/bin/Xvfb
Initializing built-in extension Generic Event Extension
Initializing built-in extension SHAPE
Initializing built-in extension MIT-SHM
Initializing built-in extension XInputExtension
Initializing built-in extension XTEST
Initializing built-in extension BIG-REQUESTS
Initializing built-in extension SYNC
Initializing built-in extension XKEYBOARD
Initializing built-in extension XC-MISC
Initializing built-in extension SECURITY
Initializing built-in extension XINERAMA
Initializing built-in extension XFIXES
Initializing built-in extension RENDER
Initializing built-in extension RANDR
Initializing built-in extension COMPOSITE
Initializing built-in extension DAMAGE
Initializing built-in extension MIT-SCREEN-SAVER
Initializing built-in extension DOUBLE-BUFFER
Initializing built-in extension RECORD
Initializing built-in extension DPMS
Initializing built-in extension Present
Initializing built-in extension DRI3
Initializing built-in extension X-Resource
Initializing built-in extension XVideo
Initializing built-in extension XVideo-MotionCompensation
Initializing built-in extension SELinux
Initializing built-in extension GLX
[dix] Could not init font path element /usr/share/fonts/X11/cyrillic, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/Type1, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi, removing from list!
Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
Setting QT_PREFERRED_BINDING to more than one binding excludes others ... ok
Preferred binding PyQt4 should have sip version 2 ... ok
Qt.py may be bundled along with another library/project ... ok
load_ui: Load widgets into self using PySide ... ok
6 XSELINUXs still allocated at reset
SCREEN: 0 objects of 264 bytes = 0 total bytes 0 private allocs
DEVICE: 0 objects of 96 bytes = 0 total bytes 0 private allocs
CLIENT: 0 objects of 144 bytes = 0 total bytes 0 private allocs
WINDOW: 0 objects of 48 bytes = 0 total bytes 0 private allocs
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
1 PIXMAPs still allocated at reset
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
4 GCs still allocated at reset
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 5 objects, 72 bytes, 0 allocs
1 CURSORs still allocated at reset
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 1 objects, 8 bytes, 0 allocs
1 CURSOR_BITSs still allocated at reset
TOTAL: 0 objects, 0 bytes, 0 allocs
[dix] Could not init font path element /usr/share/fonts/X11/cyrillic, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/Type1, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi, removing from list!
load_ui: Load widgets into self using PyQt4 ... ok
6 XSELINUXs still allocated at reset
SCREEN: 0 objects of 264 bytes = 0 total bytes 0 private allocs
DEVICE: 0 objects of 96 bytes = 0 total bytes 0 private allocs
CLIENT: 0 objects of 144 bytes = 0 total bytes 0 private allocs
WINDOW: 0 objects of 48 bytes = 0 total bytes 0 private allocs
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
1 PIXMAPs still allocated at reset
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
4 GCs still allocated at reset
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 5 objects, 72 bytes, 0 allocs
1 CURSORs still allocated at reset
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 1 objects, 8 bytes, 0 allocs
1 CURSOR_BITSs still allocated at reset
TOTAL: 0 objects, 0 bytes, 0 allocs
[dix] Could not init font path element /usr/share/fonts/X11/cyrillic, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/Type1, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi, removing from list!
load_ui: Load .ui file ... ok
6 XSELINUXs still allocated at reset
SCREEN: 0 objects of 264 bytes = 0 total bytes 0 private allocs
DEVICE: 0 objects of 96 bytes = 0 total bytes 0 private allocs
CLIENT: 0 objects of 144 bytes = 0 total bytes 0 private allocs
WINDOW: 0 objects of 48 bytes = 0 total bytes 0 private allocs
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
1 PIXMAPs still allocated at reset
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
4 GCs still allocated at reset
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 5 objects, 72 bytes, 0 allocs
1 CURSORs still allocated at reset
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 1 objects, 8 bytes, 0 allocs
1 CURSOR_BITSs still allocated at reset
TOTAL: 0 objects, 0 bytes, 0 allocs
[dix] Could not init font path element /usr/share/fonts/X11/cyrillic, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/Type1, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi, removing from list!
load_ui: Load widgets into custom using PyQt4 ... ok
6 XSELINUXs still allocated at reset
SCREEN: 0 objects of 264 bytes = 0 total bytes 0 private allocs
DEVICE: 0 objects of 96 bytes = 0 total bytes 0 private allocs
CLIENT: 0 objects of 144 bytes = 0 total bytes 0 private allocs
WINDOW: 0 objects of 48 bytes = 0 total bytes 0 private allocs
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
1 PIXMAPs still allocated at reset
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
4 GCs still allocated at reset
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 5 objects, 72 bytes, 0 allocs
1 CURSORs still allocated at reset
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 1 objects, 8 bytes, 0 allocs
1 CURSOR_BITSs still allocated at reset
TOTAL: 0 objects, 0 bytes, 0 allocs
[dix] Could not init font path element /usr/share/fonts/X11/cyrillic, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi/:unscaled, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/Type1, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/100dpi, removing from list!
[dix] Could not init font path element /usr/share/fonts/X11/75dpi, removing from list!
Raise ImportError if sip API v1 was already set (Python 2.x only) ... ok

Ran 22 tests in 3.555s

OK
```

There's also a strange error appearing on my OS X laptop (running Docker for Mac 1.12 stable):

```
```

mottosso:
> An error has occurred while reading the UI file at line 1, column 0: Premature end of document.

I got this as well, I replaced the .ui contents with this instead, that I got from an empty Qt Designer with a single pushbutton, and then never saw the error again. Not entirely sure it was the cause, possibly the lacking newline at the end, possibly the encoding or xml version. Haven't had the opportunity to investigate much further.

```python
    source = """\
<ui version="4.0">
 <class>Form</class>
 <widget class="QWidget" name="Form">
  <property name="geometry">
   <rect>
    <x>0</x>
    <y>0</y>
    <width>400</width>
    <height>300</height>
   </rect>
  </property>
  <property name="windowTitle">
   <string>Form</string>
  </property>
   <property name="geometry">
    <rect>
     <x>140</x>
     <y>120</y>
     <width>90</width>
     <height>27</height>
    </rect>
   </property>
   <property name="text">
    <string>PushButton</string>
   </property>
  </widget>
 </widget>
 <resources/>
 <connections/>
</ui>
"""
```

mottosso:
> Do you think that's the best way forward?

Yes, without a graphics card or physical display we'll need to go virtual.

> I know it's possible (not how though) to run GUIs in Docker containers

You can run GUIs with Xvfb too, they will simply be rendered to an in-memory location (that you could potentially render to a file or redirect to a display).

To redirect GUIs directly, we'll need an X-server to redirect to, or something like an X-server emulator and VNC service like [noVNC](https://kanaka.github.io/noVNC/). In a desktop environment, that's possible. But it's less automatic and headless, which is what we need for tests.

fredrikaverpil:
Ok, let's go with Xvfb  then. 👍

Dockerfile
```diff
@@ -7,21 +7,16 @@ RUN apt-get update && apt-get install -y \
     xvfb
 
 # Nose is the Python test-runner
-RUN pip install nose nosepipe
+RUN pip install nose nosepipe xvfbwrapper
```
mottosso: About this, have a look at [xvfb-run](http://manpages.ubuntu.com/manpages/xenial/man1/xvfb-run.1.html). It does the same thing, but without infecting our tests. This might limit tests from running on anything but Linux, and adds that additional indentation to a lot of lines.

```bash
$ xvfb-run docker run ...
```

fredrikaverpil:
Tests with docker seems to run smoothly now:

```
docker run --rm -v $(pwd):/Qt.py mottosso/qt.py
Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
Setting QT_PREFERRED_BINDING to more than one binding excludes others ... ok
Preferred binding PyQt4 should have sip version 2 ... ok
Qt.py may be bundled along with another library/project ... ok
load_ui: Load widgets into self using PySide ... ok
load_ui: Load widgets into self using PyQt4 ... ok
load_ui: Load .ui file into custom using PySide ... ok
load_ui: Load widgets into custom using PyQt4 ... ok
Raise ImportError if sip API v1 was already set (Python 2.x only) ... ok

Ran 22 tests in 4.415s

OK
```

However, I'm still having some odd issues on my laptop... I wonder if it could be Docker 1.12 stable.

fredrikaverpil:
> xvfb-run docker run ...

Hm, I'm not too keen on that since it'll make it impossible to run that on OS X which I like to use.

mottosso:
Understood, then we might need to find a more cross-platform way as X is only (mostly) relevant to Linux (and OSX?), and now the tests are coupled with it.

fredrikaverpil:
> we might need to find a more cross-platform way

That would be nice. Hm, but what would that be?

I think that should be its own issue/PR.

mottosso:
> That would be nice. Hm, but what would that be?

At the moment, our tests require a window manager to render windows into. Fulfillment of this requirement varies on every platform.

As it happens, we're running tests on Travis which so happens to be running Ubuntu. On Ubuntu, there is a window manager emulator called Xvfb. That solves the problem for Travis, but because our test are now hardcoded to only work with Xvfb, it eliminates their potential for running on other platforms or flavours of Linux that doesn't use X (e.g. Wayland).

I don't think we should accept a PR that eliminates cross-platform testability. I think we should isolate use of Xvfb to Travis and our Docker image, without affecting other platforms.

fredrikaverpil:
> I don't think we should accept a PR that eliminates cross-platform testability. I think we should isolate use of Xvfb to Travis and our Docker image, without affecting other platforms.

I agree. What's your suggestion on how to eliminate xvfbwrapper from tests.py?
Should I do a `platform.system()` check and only if not Windows is detected, I use vfbwrapper?

mottosso:
> What's your suggestion on how to eliminate xvfbwrapper from tests.py?

Let's not write any platform-specific code in the module.

With xfvb-run, users on Linux can run these tests. We'll need to find similar solutions for other platforms and put it in the README.

fredrikaverpil:
I found a solution on OS X which doesn't involve a virtual framebuffer... and I wrote down some notes on it [here](https://fredrikaverpil.github.io/2016/07/31/docker-for-mac-and-gui-applications/).

It's not scriptable, unfortunately. But when you've got XQuartz set up this works:

```
$ docker run --rm -v $(pwd):/Qt.py -e DISPLAY=$ip:0 -v /tmp/.X11-unix:/tmp/.X11-unix mottosso/qt.py

Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
Setting QT_PREFERRED_BINDING to more than one binding excludes others ... ok
Preferred binding PyQt4 should have sip version 2 ... ok
Qt.py may be bundled along with another library/project ... ok
load_ui: Load widgets into self using PySide ... ok
load_ui: Load widgets into self using PyQt4 ... ok
load_ui: Load widgets into custom using PySide ... ok
load_ui: Load widgets into custom using PyQt4 ... ok
Raise ImportError if sip API v1 was already set (Python 2.x only) ... ok

Ran 22 tests in 6.406s

OK
```

fredrikaverpil:
Perhaps [this](https://github.com/docker/docker/issues/8710#issuecomment-135109677) could be something for Windows?

mottosso:
Ok, I think we've got a complication.

Are you by any chance looking to solve running tests within our Docker image (which is Ubuntu) where Docker is running natively on MacOS? This may be why you aren't finding much help online. The native Docker on MacOS and Windows are brand new, I wouldn't expect this kind of functionality to work well just yet.

What I'm talking about is running it on Travis, or locally via Docker, where Docker is running within something like VirtualBox. Which is what you'd have to do before these native clients came out.

If we leave the brand new native clients of Docker aside, tests already run locally on Windows without Docker, and with Docker via Xvfb because Docker itself is running under Linux. On MacOS, I'd expect the same to be true.

I'll have a look at setting up headless testing on Windows just to get our bases covered.

fredrikaverpil:
> Are you by any chance looking to solve running tests within our Docker image (which is Ubuntu) where Docker is running natively on MacOS?

Yes, but only for personal development. I figured I might add a comment on that in the README if I got it to work (which I did).

> This may be why you aren't finding much help online. The native Docker on MacOS and Windows are brand new, I wouldn't expect this kind of functionality to work well just yet.

Actually, and perhaps limited to macs, it's seems to me that it's easier than with the old Docker client which runs in Virtualbox on Macs. You can basically skip the `socat` step to redirect sockets to XQuartz.

But you are right in the sense that some features such as audio isn't as easily solved, as the past solution was dependent on Virtualbox.

> If we leave the brand new native clients of Docker aside, tests already run locally on Windows without Docker, and with Docker via Xvfb because Docker itself is running under Linux. On MacOS, I'd expect the same to be true.

I'm sorry... I think I lost you there. What did you expect to be true on OS X with the old Docker clients?  
You could install PySide, PyQt4, nose, nosepipe etc and run `tests.py` natively (without docker or any kind of xvfb thing) on OS X, if that is what you meant.

___

Bottom line is I wanted to see if I could get the tests working on OS X using the `mottosso/Qt.py` docker container, when running on the new "Docker for Mac", which I'm using. And it worked without issues and without modifying the Dockerfile other than what was added to it in this PR for Xvfb.

I think we're good with just adding a comment to the README about what these recent tests additions mean in terms of requiring some kind of X Window System. Since if you just blindly follow the instructions right now, you'll get the tests running but will get errors on the ones added in this PR.

What do you think?

mottosso:
> we could add a comment to the README about what these recent tests additions mean in terms of requiring some kind of X Window System.

The tests doesn't require X. The tests require a window manager and X is one of many window managers on Linux. XQuartz that you are using looks like an OSX port of X, but OSX already has a window manager. Isn't it Quartz? Windows also already has a Window manager.

In any case, I think we've gone a bit far.

At the moment, the Docker container contains the OS, the software and the tests. It's completely isolated and platform-indepentent. Except for one thing; graphics drawing.

Docker is currently trying to reach outside of its image and ask the OS for help with drawing something, and we are handing it an OS-specific handle.

What we can do instead, is to handle all graphics requests right inside of the Docker image. Looking at your commit history, it looks like [you've already done this](https://github.com/mottosso/Qt.py/pull/81/commits/6dad71fcd528af54cae87793d15f0d297b541ec9), but then later added the xvfbwrapper.

Have a look at this image, it's similar to the one you committed, with the addition of squelching the output coming from xvfb.

```dockerfile
FROM ubuntu:14.04

RUN apt-get update && apt-get install -y \
    python-qt4 \
    python-pyside \
    python-pip \
    xvfb

# Nose is the Python test-runner
RUN pip install nose nosepipe

# Enable additional output from Qt.py
ENV QT_VERBOSE true
ENV DISPLAY :99.0

WORKDIR /workspace/Qt.py
ENTRYPOINT \
	Xvfb :99 -screen 0 1024x768x16 2>/dev/null & \
	sleep 3 && \
	cp -r /Qt.py /workspace && \
	python build_caveats_tests.py && \
	nosetests \
		--verbose \
		--with-process-isolation \
		--with-doctest \
		--exe 
```

This image makes no request to the OS for graphics, and should be entirely cross-platform.

mottosso:
Actually, it looks like the current state of things includes this already? 

- [69bb7](https://github.com/fredrikaverpil/Qt.py/blob/69bb70f6b50f673e07b5af8d3c6aa5fd44ecee13/Dockerfile)

You shouldn't need to pass a handle to your physical display in that case.

```bash
-v /tmp/.X11-unix:/tmp/.X11-unix
```

fredrikaverpil:
> The tests doesn't require X. The tests require a window manager and X is one of many window managers on Linux.

Ah, yes. True!

> XQuartz that you are using looks like an OSX port of X, but OSX already has a window manager. Isn't it Quartz?

X11 was previously bundled with OS X but not any more: https://support.apple.com/en-us/HT201341

> This image makes no request to the OS for graphics, and should be entirely cross-platform.

Ok, great, let's try to make it work like that!

> You shouldn't need to pass a handle to your physical display in that case.

Ah, yes - you are so right!
See below:

```bash
$ open -a XQuartz

$ ip=$(ifconfig en0 | grep inet | awk '$1=="inet" {print $2}')

$ xhost + $ip
192.168.0.123 being added to access control list

$ docker run --rm -e DISPLAY=$ip:0 -v $(pwd):/Qt.py mottosso/qt.py

Tests require PySide and PyQt4 bindings to be installed ... ok
Setting QT_PREFERRED_BINDING properly forces a particular binding ... ok
Setting QT_PREFERRED_BINDING to more than one binding excludes others ... ok
Preferred binding PyQt4 should have sip version 2 ... ok
Qt.py may be bundled along with another library/project ... ok
load_ui: Load widgets into self using PySide ... ok
load_ui: Load widgets into self using PyQt4 ... ok
load_ui: Load widgets into custom using PySide ... ok
load_ui: Load widgets into custom using PyQt4 ... ok
Raise ImportError if sip API v1 was already set (Python 2.x only) ... ok

Ran 22 tests in 5.098s

OK
```

___

I'm just about to step out the door and won't be back until tomorrow. I see we get the Designer error in AppVeyor, even after changing the UI template into the one you [posted](https://github.com/mottosso/Qt.py/pull/81#issuecomment-236425103). Not sure what that's about. We should probably change the UI back into being a QMainWindow either way.

mottosso:
The errors you're getting on Windows now may be related to this.

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). If delete is true (the default), the file is deleted as soon as it is closed.

- https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile

To resolve it, we can stop relying on `NamedTemporaryFile` and instead write to a file ourselves.

```python
...
PYTHON = sys.version_info[0]  # e.g. 2 or 3

self = sys.modules[__name__]


def setup():
    self.tempdir = tempfile.mkdtemp()
    self.ui = os.path.join(self.tempdir, "temp.ui")

    ui = """\

UI HERE

"""

    with open(self.ui, "w") as f:
        f.write(ui)


def teardown():
    shutil.rmtree(self.tempdir)
...
```

Then use `self.ui` instead of `with ui()`.

mottosso:
> I'm just about to step out the door and won't be back until tomorrow. 

No problem, I'll take a look at this.

Qt.py
```diff
@@ -118,60 +118,184 @@ def _pyside():
     return PySide
 
 
-def pyside_load_ui(fname):
+def pyside_load_ui(fname, base_instance=None, custom_widgets=None):
     """Read Qt Designer .ui `fname`
 
     Args:
         fname (str): Absolute path to .ui file
+        base_instance (widget): Optional instance of the Qt base class.
```
mottosso: This should be.

```python
base_instance (widget, optional): Instance of the Qt base class.
```

mottosso:
Can we remove the `custom_widgets` argument?

Qt.py
```diff
@@ -118,60 +118,184 @@ def _pyside():
...
+        custom_widgets (widget): ?
 
     Usage:
         >> from Qt import load_ui
         >> class MyWindow(QtWidgets.QWidget):
         ..   fname = 'my_ui.ui'
```
mottosso: This example doesn't looks right, shouldn't there be an `__init__()`?

mottosso: If you use `>>>` the example will get run on Travis.

Qt.py
```diff
@@ -118,60 +118,184 @@ def _pyside():
...
+    Note:
+        This function is based on the gist:
+        https://gist.github.com/cpbotha/1b42a20c8f3eb9bb7cb8
 
+    """
 
-def pyside2_load_ui(fname):
+    from PySide import QtUiTools, QtCore
+
+    class UiLoader(QtUiTools.QUiLoader):
```
mottosso: Not a big fan of declaring the class twice. :(

bpabel:
How do you normally run the tests?  Do you run them locally, or do you run them all through travis?  They don't seem to be passing on windows.  I can setup a VM I suppose, but I don't see any reason why the tests should work differently on Windows.

mottosso:
Make sure you have PyQt4 and PySide available, along with nose and nosepipe.

Then try this.

```bash
$ cd Qt.py
$ nosetests --verbose --with-process-isolation --with-doctest
```

mottosso:
Alternatively, if you have Docker, see this.

- https://github.com/mottosso/Qt.py#developer-guide

bpabel:
Hmm.. this is the output I get

```
PS C:\Users\brendan\workspace\Qt.py> nosetests --verbose --with-process-isolation

Ran 0 tests in 1.200s

OK
```

Might be related to the nosepipes plugin, since it will run the tests if I don't include the `--with-process-isolation` flag; though, a few of the tests fail, I think because it's not resetting the `sys.modules`.

mottosso:
> I think because it's not resetting the sys.module

That's part of what nosepipe is managing. It's running each test in a separate process, so that Qt.py can be imported over and over. You'll need it to pass the tests.

What happens if you add tests.py to the end?

```bash
$ nosetests --verbose --with-process-isolation tests.py
```

Have a look at nosepipe and whether it installed correctly. Here's a full transcript of what works for me.

```cmd
$ virtualenv nosepipetest --no-site-packages
$ nosepipetest\Scripts\activate
(nosepipe) $ pip install nose nosepipe
:: These must be provided by you
(nosepipe) $ set PYTHONPATH=C:\modules\python-qt4;C:\modules\python-pyside
(nosepipe) $ set PATH=C:\modules\python-qt4\PyQt4;%PATH%
(nosepipe) $ nosetests --verbose --with-process-isolation --with-doctest
Test qtgui_qabstractitemmodel_createindex ... ok
Test qtgui_qabstractitemmodel_createindex ... ok
...

Ran 22 tests in 10.641s

OK
```

mottosso:
> PS C:\Users\brendan\workspace\Qt.py> nosetests --verbose --with-process-isolation

Wait, are you running in PowerShell? I'm not sure whether that works, could you try in `cmd.exe`?

fredrikaverpil:
> This example doesn't looks right, shouldn't there be an __init__()?
> If you use >>> the example will get run on Travis.

You are correct. I've fixed this now, but there's no `my_ui.ui` file.

> Can we remove the custom_widgets argument?

Well. We could. If the binding used is PyQt, this third argument will be accepted since it will map straight onto `uic.loadUi`. For this reason, I think why not keep it in there for PySide as well?
All I need to understand is how we should test it ... and how to describe its argument. I haven't had the time to look into it.

> Not a big fan of declaring the class twice. :(

Neither am I. Will look into it.

fredrikaverpil:
> Because of its docstring; a questionmark. Because you don't seem to understand it. Because you don't know how to test it. And because it is a lot easier to add than it is to remove.

It's true I don't understand it. I'm waiting for @bpabel to provide an example. 

This is from the PyQt4 docs:
* `uifile` – the file name or file-like object containing the .ui file.
* `baseinstance` – the optional instance of the Qt base class. If specified then the user interface is created in it. Otherwise a new instance of the base class is automatically created.
* `package` – the optional package that is the base package for any relative imports of custom widgets.
- [Source](http://pyqt.sourceforge.net/Docs/PyQt4/designer.html#PyQt4.uic.loadUi).

Unfortunately, I don't understand what that does by just reading it and I still haven't found any examples. If we don't see any examples of this I'm all for removing it.

mottosso:
Yeah, sorry, re-reading that made me realize that it came off much too strong. Was in the middle of something and was writing from an iPad. But it's important we implement only what we fully understand, and that we test it thoroughly. I see no issue in excluding this last argument this time around, it can be added later if need be.

fredrikaverpil:
Sure, I wouldn't have any problem with that.

@bpabel, were you looking to use this `custom_widget` attribute?

fredrikaverpil:
Hm. Travis failed because this was added (because of me fixing those docstrings?) to tests:

    Doctest: PySide.load_ui ... ERROR

```
======================================================================
ERROR: Doctest: PySide.load_ui
  File "/usr/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/usr/local/lib/python2.7/dist-packages/nose/failure.py", line 42, in runTest
    raise self.exc_class(self.exc_val)
ValueError: No such test pyside_load_ui
```

I'll look into it .. but, @mottosso do you know why that happened and why it is now looking for a test called `pyside_load_ui`?

mottosso:
About the breaking doctest, it's breaking because as you said, it's missing the file it's referencing.

Part of the suggestion was for you to edit the doctest into something complete such that it can be tested. It's tough testing something that relies on opening a file, without having this file. But that's where we are.

There are ways around it, one example is NamedTemporaryFile like we were using before. Another is monkey-patching `open` such that the function uses this instead. Neither is great of course.

The great thing about these doctest is that they force you to look at how complex a simple example is. Usually being forced to monkey-patch your way around and produce boilerplate is a sign of complexity.

Ideally I'd rather have a function that is testable. For example, something that takes a string with the contents of a file, rather than a path. That's break backwards compatibility and familiarity, but might yield better stability long-term.

Maybe let's exclude the doctest for now, and have a chat about that in another PR?

fredrikaverpil:
> Maybe let's exclude the doctest for now, and have a chat about that in another PR?

I think so, yeah.

I've now made sure to define the UiLoader class once. However, this made the code a bit complex. What do you think, do you have a better suggestion on how to solve that?

fredrikaverpil:
All that remains right now are:

* Keep or ditch `custom_widgets` attribute. Example needed to keep it.
* Current implementation of `pyside_load_ui_wrap` and `pyqt_load_ui_wrap` - are they fine?

mottosso:
> However, this made the code a bit complex.

A bit complex yes. I think all we need is a factory function. Something that produces a class, or instance of a class.

I'll PR a suggestion.

fredrikaverpil:
> I'll PR a suggestion.

Thanks, I appreciate it! 🙇

fredrikaverpil:
Hm. How do you retry the AppVeyor build?

mottosso:
> Hm. How do you retry the AppVeyor build?

Maybe only the owner can trigger a rebuild. I did that now, and we're all good.

In rewriting the UiLoader class, I found a few things left to test.

- [What if parent is not None?](https://github.com/fredrikaverpil/Qt.py/blob/7703fda5083011b86beff8d8a9092fdefcd128b7/Qt.py#L142)
- [What is a widget is not available?](https://github.com/fredrikaverpil/Qt.py/blob/7703fda5083011b86beff8d8a9092fdefcd128b7/Qt.py#L145)
- [Can this fail?](https://github.com/fredrikaverpil/Qt.py/blob/7703fda5083011b86beff8d8a9092fdefcd128b7/Qt.py#L148)
- [What happens in both of these cases?](https://github.com/fredrikaverpil/Qt.py/blob/7703fda5083011b86beff8d8a9092fdefcd128b7/Qt.py#L151)

fredrikaverpil:
Awesome, I'll look into it as soon as I can. Will also look over the README on the arguments and add a note on Xvfb.

mottosso:
Take your time! I thought you were on holiday? :)

fredrikaverpil:
> Take your time! I thought you were on holiday? :)

I am! :)

fredrikaverpil:
For those tests, do you mean we should re-create the `UiLoader` class in the tests rather than use the one in Qt.py?

mottosso:
> For those tests, do you mean we should re-create the UiLoader class in the tests rather than use the one in Qt.py?

The tests should be testing the one in Qt.py.

Here's what I have in mind.

```python
def test_load_ui_thisandthat():
  with pyside():
    import Qt
    widget = Qt.load_ui("my.ui")
    assert this
    assert that
```

The goal is to produce `.ui` files and use them in such a way that they exercise everything there is to know about `UiLoader` and `load_ui`.

bpabel:
> Wait, are you running in PowerShell? I'm not sure whether that works, could you try in cmd.exe?

Got the same results with the standard cmd.exe.

fredrikaverpil:
I'm not sure if I can really test the `createWidget` method based on the UI file and the resulting widget - or if I've simply misunderstood how you meant to perform these tests...

Anyway; I've added some assertions to the existing tests to make sure that e.g. the `QMainWindow` has a parent of `None` and that `QPushButton` has a parent of `QWidget`. When I did some print debugging, I noticed it was only `MainWindow` which passed the `if parent is None and self.base_instance` if condition.

fredrikaverpil:
Moving on to the if condition where a class name is checked against `availableWidgets`...

For some reason, I can't perform an `assert_raises` without crashing here. If I don't do the assertion and just let the test run it exits with the expected exception along with an additional print:

```python
ERROR: load_ui: Invalid class name in .ui
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/local/lib/python2.7/dist-packages/nosepipe.py", line 179, in __call__
    (header + data).decode("latin1"))
Exception: short message body (want 1416782179, got 267)
Something went wrong
Message: Traceback (most recent call last):
  File "/workspace/Qt.py/Qt.py", line 166, in createWidget
    raise Exception("\"%s\" not available." % class_name)
Exception: "QInvalidWidget" not available.
"Empty widget item in QGridLayout 'gridLayout'." 

addSuccesstopTest

Ran 23 tests in 5.253s

FAILED (errors=1)
```

mottosso:
Can't tell why that is just from looking and I haven't got a computer for another week (this time it's me who's on holiday :).

But I wanted to share that qtpy has a similar solution that we should learn from.

- https://github.com/spyder-ide/qtpy/blob/master/qtpy/uic.py

tests.py
```diff
@@ -299,6 +341,37 @@ def test_load_ui_into_custom_pyqt4():
...
+     'Parent:', <PySide.QtGui.QMainWindow object at 0x1036d3ea8>)
+    ('Name:', u'statusbar', 'Class:', u'QStatusBar',
+     'Parent:', <PySide.QtGui.QMainWindow object at 0x1036d3ea8>)
+
+    """
+
+    with pyside():
+        from Qt import QtWidgets, load_ui
+
+        class MainWindow(QtWidgets.QMainWindow):
```
mottosso: Do we need this class to test this feature? Looks like a plain load should excersize the same functionality.

fredrikaverpil: True, I'll fix that.

mottosso:
About the assert_raises, if you haven't yet solved it, a regular try/except should work fine in the meantime.

fredrikaverpil:
That's what's bothering me, it doesn't help doing a try/except:

```python
def test_load_ui_invalid_class_name():
    """load_ui: Invalid class name in .ui

    """

    with pyside():
        from Qt import QtWidgets, load_ui

        app = QtWidgets.QApplication(sys.argv)
        try:
            assert_raises(Exception,
                          load_ui,
                          sys.modules[__name__].ui_invalid_class)
        except:
            print('okay')
```

```
======================================================================
ERROR: load_ui: Invalid class name in .ui
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/local/lib/python2.7/dist-packages/nosepipe.py", line 179, in __call__
    (header + data).decode("latin1"))
Exception: short message body (want 1416782179, got 267)
Something went wrong
Message: Traceback (most recent call last):
  File "/workspace/Qt.py/Qt.py", line 164, in createWidget
    raise Exception("\"%s\" not available." % class_name)
Exception: "QInvalidWidget" not available.
"Empty widget item in QGridLayout 'gridLayout'." 

addSuccesstopTest

Ran 23 tests in 5.860s

FAILED (errors=1)
```

fredrikaverpil:
Okay, so about the custom widgets... I put that back in, read up on it and made tests for it. It's kind of hacky but it works. Not sure what you guys think about it.

fredrikaverpil:
What annoys me a bit is I can't create a test where I'm trying to use an invalid widget. This could e.g. be a custom widget which I haven't defined properly. So I've left this test commented out.

appveyor.yml
```diff
@@ -22,7 +22,7 @@ install:
   # the registry key.
 
   - REG ADD HKCU\Software\Python\PythonCore\2.7\InstallPath /f /ve /t REG_SZ /d %PYTHON%
-  - ps: (new-object net.webclient).DownloadFile($env:PYQT, "C:\install-PyQt4.exe")
+  - ps: do{sleep 5;(New-Object Net.WebClient).DownloadFile($env:PYQT, "C:\install-PyQt4.exe")}while(!$?)
```
fredrikaverpil: Hopefully no more b0rked tests because this wasn't downloaded 😎

mottosso: Yes, hopelessly unreliable the sourceforge downloads.. :(

mottosso:
I still think we should leave `custom_widgets` out.

Now it looks like we added a feature we don't fully understand because we were unable to test the features we already had in place without it. This is no good.

mottosso:
> Anyway, I've left this test commented out. Perhaps we should instead remove it?

Yes, if we must.

Features must still be tested however. Have a look at coverage.py for what is currently being tested and what is not. I will add it to Travis so that we can keep travk of it, but for now you'll need to run it by hand.

fredrikaverpil:
> Now it looks like we added a feature we don't fully understand because we were unable to test the features we already had in place without it. This is no good.

Actually, I think you have gotten that wrong. I understand how the custom widgets works. And I'm testing successfully for that. There are four tests, all passing.

The test which I can't get to work is still a problem regardless of having the custom widgets in there or not. As you can see from this commit, I were having issues with that test after you removed custom widgets and before I put it back: https://github.com/mottosso/Qt.py/pull/81/commits/7dd4d6c56cf2d1dbc708f73b0b67445a2f958988

That particular test should raise an `Exception` if something is defined in the .ui which doesn't exist, like a `QFunkyStuffsWedgeSandwich`.

So that issue remains regardless of if we take custom widgets out or not.

> I still think we should leave custom_widgets out.

Do you think so even if the problematic test doesn't really have anything to do with custom widgets?

fredrikaverpil:
I've removed support for custom widgets and I can add that back in as a separate PR later, and we can discuss that then. Or I could revert this commit if you think it should be in this PR.

[This](https://github.com/mottosso/Qt.py/pull/81/commits/a2f8c475b7b66fed5768434a9d6c00b9cba79a78#diff-cd1d7c0a982bce680146d5b248151256R164) is untested at the moment, as it seems nose crashes when this exception is hit.

crimsoncor:
I will say that I tried using code very similar to this in QtPy and ran into massive issues in Maya when loading uifiles that pointed to QMainWindows or QDialog, which were parented to the main Maya window. This is apparently a known issue that a lot of people have run into in Maya
http://tech-artists.org/forum/showthread.php?5157-PySide-in-Maya-quot-Internal-C-object-already-deleted-quot-for-returned-PySide-objects
The PySide release notes claim they have fixed the bugs with things that extend QUiLoader, but based on this I don't believe them (or there is something in the Maya release of PySide which is not correctly patched).  

I have fixed these issues and have everything working great in pyside by using the pysideuic module and some extra code from this gist:
https://gist.github.com/mstuttgart/bc246b25b8e0f7edd743
After using that code to implement loadUiType (which also has the advantage of providing more compatibility with PyQt's uic module) I then reimplemented loadUi as follows:

```
    def loadUi(uifile, baseinstance=None, workingDirectory=None):
        """
        An implementation that uses the custom loadUiType above.
        This seems to work correctly in Maya as well as outside of it as
        opposed to other implementation what overrride QUiLoader.
        """
        form_class, base_class = loadUiType(uifile)
        if not baseinstance:
            baseinstance = base_class()
            typeName = form_class.__name__
        else:
            if not isinstance(baseinstance, base_class):
                raise RuntimeError(
                    "The baseinstance passed to loadUi does not inherit from"
                        " needed base type (%s)" % type(base_class))
            typeName = type(baseinstance).__name__
        baseinstance.__class__ = type(typeName,
                                      (form_class, type(baseinstance)),
                                      {})
        baseinstance.setupUi(baseinstance)
```
It is currently ignoring working directory, since i don't have any code that uses that option, but I think it would be possible to extend. I can confirm, that at least for my current examples, that this loads all ui files in Maya without a problem.

fredrikaverpil:
@crimsoncor so your code snippet uses the `pysideuic` module?
Could you please provide a full working example of your suggestion?

I opted to not use `pysideuic` as this is not included with e.g. Nuke. I've seen the `Internal C++ object already deleted` error appear when using `pysideuic` as well, which has been the goto solution in our studio for at least two years now.

mottosso:
Thanks a lot for this, @crimsoncor. That is just the right insight needed for a compatibility project like this.

mottosso:
With #100 in mind, and for completeness, could we open up the discussion about how to achieve the desired results with already existing functionality?

That is, how can we use or hack `Qt.load_ui("my.ui")` to do the things these additional arguments can do? For example, can these ui files be computed beforehand, eliminating the use of `load_ui` altogether? What other options are there.

crimsoncor:
It is possible to get an `Internal C++ object already deleted` error for legit reasons in `PySide`. There are places, like `setCentralWidget` in `QMainWindow`, where `Qt` takes ownership of the widget and will delete it when it goes out of scope (in the C++ layer) even if you still have a python handle to the object. There are a bunch of other places in the `Qt` API with similar behavior. However (and take this with multiple caveats because 90% of our stack is PyQt and not PySide) I have not seen properly formed `Qt` code fail with that error in `PySide`, except when using a ui loader that extends `QUILoader` in Maya and when parenting a window to the main Maya UI window. Even if a valid python pointer to that `QMainWindow` is held, it gets garbage collected in the C++ layer. Exact same code work fine in `PyQt`. And using `pysideuic` resolved that. I don't know what your error cases are, so maybe there are some fun bugs with `pysideuic` in my future.

We're not a Nuke shop so I wasn't aware that some developers were shipping embedded PySide without pysideuic. One solution might be to include both code paths as an option with the pysideuic one preferred if the import exists, otherwise with it falling back to the QUILoader code path. something like:
```
try:
    import pysideuic
    # Do pysideuic import path
except ImportError:
    # Take UILoader path.
```
It might be worth the community talking to the Foundry about including `pysideuic` with Nuke since that seems to be the more robust way of loading UI files in PySide.

My entire code block for loading UI files in pyside follows. You'll have to the change the qtpy import line to be relevant to your compatibility later. This code also appears to work fine with PySide2 with the only change being replacing `pysideuic` with `pyside2uic`. We're still in the testing process with PySide2, so I can't swear that there aren't still some remaining issues.
```
    def loadUiType(uiFile):
        """
        Pyside equiv for the loadUiType function in qt.
        """
        import qtpy.QtWidgets as QtWidgets
        import pysideuic
        import xml.etree.ElementTree as ElementTree
        from cStringIO import StringIO

        parsed = ElementTree.parse(uiFile)
        widget_class = parsed.find('widget').get('class')
        form_class = parsed.find('class').text

        with open(uiFile, 'r') as f:
            o = StringIO()
            frame = {}

            pysideuic.compileUi(f, o, indent=0)
            pyc = compile(o.getvalue(), '<string>', 'exec')
            exec pyc in frame

            #Fetch the base_class and form class based on their type in
            #the xml from designer
            form_class = frame['Ui_%s' % form_class]
            base_class = eval('QtWidgets.%s' % widget_class)
        return form_class, base_class


    def loadUi(uifile, baseinstance=None, workingDirectory=None):
        """
        An implementation that uses the custom loadUiType above.
        This seems to work correctly in Maya as well as outside of it as
        opposed to other implementation what overrride QUiLoader.
        """
        form_class, base_class = loadUiType(uifile)
        if not baseinstance:
            typeName = form_class.__name__
            finalType = type(typeName,
                             (form_class, base_class),
                             {})
            baseinstance = finalType()
        else:
            if not isinstance(baseinstance, base_class):
                raise RuntimeError(
                    "The baseinstance passed to loadUi does not inherit from"
                        " needed base type (%s)" % type(base_class))
            typeName = type(baseinstance).__name__
            baseinstance.__class__ = type(typeName,
                                          (form_class, type(baseinstance)),
                                          {})
        baseinstance.setupUi(baseinstance)
        return baseinstance
```

fredrikaverpil:
@crimsoncor That's excellent, thank you for the complete code example! :smiley: 

> With #101 in mind, and for completeness, could we open up the discussion about how to achieve the desired results with already existing functionality?

I think in that case (#101) maybe we shouldn't implement a custom class, or extended class and instead just support what we currently have which really are mappings. Supporting a base instance in PySide would mean adding at least 100 lines of code to Qt.py no matter how you look at it, as far as I know.

I'm not sure how #101 changes our goal with e.g. supporting this base instance argument in PySide. Do you feel we should not extend PySide's functionality to match that of PyQt in general - and instead focus on simpler remappings between PyQt4-PyQt5 and PySide-PySide2?

> For example, can these ui files be computed beforehand, eliminating the use of load_ui altogether?

Yes, I believe you can generate .py files from these .ui files. I've just never done that. There are commandline tools for that; `pyside-uic`, `pyuic4`.

mottosso:
> I'm not sure how/if you can load these .py files into a base instance.

Ok, maybe let's break this down and have a look at what we are actually getting from this function. What does it mean to load a ui file into a base instance? Is there any way of doing this without using this function?

What doesn't this technique do that load_ui does?

```python
from Qt import QtWidgets, load_ui

class MyWidget(QtWidgets.QWidget):
  pass

ui = load_ui("my.ui")
my_widget = MyWidget()
for member in dir(ui):
  setattr(my_widget, member, getattr(ui, member))

my_widget.show()
```

> Do you feel we should not extend PySide's functionality to match that of PyQt in general

I think we don't have to limit ourselves to whatever our predecessors have done, and can instead be free to think creatively about how to best support cross-binding development.

Their implementations did not have our requirements but instead tailored a soution fit for theirs. I would imagine that developing with ui files and Qt.py could simply be different from that and simply better suited for software that works across multiple bindings.

fredrikaverpil:
Interesting idea. 

When the .ui is based on QMainWindow, like in the .ui in our tests, this crashes when `member` is `__class__` or `__dict__`:

    # TypeError: __class__ assignment: only for heap types

    # AttributeError: attribute '__dict__' of 'Shiboken.Object' objects is not writable

...but other from that, it seems to work with the .ui from our tests. I have no idea if that would work if we used a QDialog or QWidget instead of QMainWindow though...

I wonder how we could more safely except certain attributes.

fredrikaverpil:
That's funny. I've successfully loaded three different .ui files now. They're based on QMainWindow, QWidget and QDialog respectively.

I actually think this might work.

mottosso:
> I actually think this might work.

I'm sure there will be edge cases that won't, and I think it's reasonable to expect different practices and gotchas alongside the cross-binding benefit you get.

For example signals is one thing I'd expect will need special treatment, they probably get automatically connected via some form of logic you may be able to reproduce if you wanted. The same goes for `.Property` and `.Slot`.

Or **don't**. Instead make it part of the "best practices" section. Whichever produces the best and most robust interface, and conversely the most robust and easily understood software built with it that works across bindings. That's all that matters.

For kicks, here's another one that might work.

```python
from Qt import QtWidgets, load_ui

Ui = type(load_ui("my.ui"))  # Get the un-instantiated class

class MyWidget(Ui):          # Inherit from it.
    super(MyWidget, self).__init__(parent)

    self.some_signal.connect(self.my_method)

  def my_method(self):
    print("Triggered!")

my_widget = MyWidget()
my_widget.show()
```

> For now, I'm just checking member against a tuple ('__class__', '__dict__') to avoid crashes.

I think you can safely ignore members that `startswith("_")`, or perhaps more safely `startswith("__")`.

fredrikaverpil:
> For kicks, here's another one that might work.

Nice! Added a variation of that as a test.

> I think you can safely ignore members that startswith("_"), or perhaps more safely startswith("__").

I performed a few local tests and yes, I'm now skipping anything with double leading underscores.

I also cleaned up the descriptions in tests.py as well as adjusted the README accordingly.

mottosso:
I think we're making great strides here.

Based on the new tests, and the fact that this simplification of manually augmenting another class with a ui file works, I've got a suggestion on how to boil it all down into a minimal addition to the project.

How about we leave the additional arguments out, and make these the official methods for working with ui-files and Qt.py?

**Method 1**

```python
import Qt

widget = Qt.load_ui("my.ui")
widget.show()
```

**Method 2**

```python
Class = type(setup_ui())

class MyWidget(Class):
  pass

widget = MyWidget()
widget.show()
```

If it so happens that this covers all use cases, then we'd be adding one hell of a feature, at the cost of an ultra-minimal amount of code (i.e. 0 lines) and a tiny amount of documentation.

fredrikaverpil:
I love your suggestion, but I don't understand how that is achievable.

Your first example is what we have today. But I don't understand your second example. The test I added which uses `setup_ui` assumes our `load_ui` function accepts the second `base_instance` argument and if so does its `setattr`/`getattr` thing.

What I wish to achieve with the base instance argument in `load_ui` is I want to somehow specify into which already existing widget the UI should be loaded. Here's a working example with the added second argument to `load_ui` (as seen in the latest test added):

```python
import sys
from Qt import QtWidgets, load_ui

def setup_ui(base_instance=None):
    return load_ui('ui.ui', base_instance)

app = QtWidgets.QApplication(sys.argv)
Class = type(setup_ui())

class MyWidget(Class):
        super(MyWidget, self).__init__(parent)
        setup_ui(self)

widget = MyWidget()
widget.show()
app.exec_()
```

If we were to add 0 lines of code and skip this PR you'd need to do something like this ... working example:

```python
import sys
from Qt import QtWidgets, load_ui

def setup_ui(base_instance=None):
    ui = load_ui('ui.ui')
    if not base_instance:
        return ui
    else:
        for member in dir(ui):
            if not member.startswith('__') and member is not 'staticMetaObject':
                setattr(base_instance, member, getattr(ui, member))
        return ui


app = QtWidgets.QApplication(sys.argv)
Class = type(setup_ui())

class MyWidget(Class):
        super(MyWidget, self).__init__(parent)
        setup_ui(self)


widget = MyWidget()
widget.show()
app.exec_()
```

___

Perhaps it is time to decide whether or not we should at all provide a this second argument to `load_ui`, since it seems we must then be able to test for it and maintain it in case it turns out not being stable.

I'm all for presenting the above example code instead of accepting this PR, don't get me wrong. But I believe we've reached a point where if we can't collectively decide on which of all the different routes which have been presented so far the end-user is better off deciding themselves on which route to take. If this is the case, I'd like to propose we create an "examples" folder or something, where we can document usage and where anyone can contribute their hacks which utilises Qt.py. For instance, we could add all of the proposed solutions on how to load the .ui in PySide to mimic the behavior in PyQt4. We can add comments and valuable lessons learned etc.

mottosso:
Ok, looks like we've caught a few different understanding of things, that's a good thing.

This is the part I'm having trouble with.

```python
Class = type(setup_ui())

class MyWidget(Class):
        super(MyWidget, self).__init__(parent)
        setup_ui(self)
```

If you inherit from the GUI built from the `ui` file, then there shouldn't be any need to run `setup_ui`. It should already have inherited everything.

What does `setup_ui` do, that isn't already done via inheritance?

fredrikaverpil:
```python
Class = type(setup_ui())
class MyWidget(Class):
pass
```

is the same as 

```python
Class = type(QtWidget.QMainWindow)
class MyWidget(Class):
    pass
```

...and therefore it doesn't set up the ui from the file.

In this PR's tests.py, I added a test where I also created a `setup_ui` function which takes a base instance argument. If added, it will perform the setattr/getattr to this base instance.

This means that you could do this to also set up the ui (pseudo code):

```python
Class = type(QtWidget.QMainWindow)
class MyWidget(Class):
    setup_ui(self)
```

fredrikaverpil:
Sorry for all the editing of my previous post. I'm done now. :)

mottosso:
> is the same as

Aha! This is what I didn't understand. I would have thought that the instance returned by `load_ui` would be an instance of the class generated from the ui file.

E.g.

```python
def load_ui():
  class MyCustomUi(QMainWindow):
    pass

  return MyCustomUi()
```

And in that way, getting the class from the instance would retrieve the original, that still contained the ui; i.e. `MyCustomUi`.

The fact that it instead returns an instance of QMainWindow is utterly strange. :S

fredrikaverpil:
> The fact that it instead returns an instance of QMainWindow is utterly strange. :S

Yeah, it's a bit odd. I agree 😕

fredrikaverpil:
But anyway, even if that would have worked... that approach wouldn't had worked if you wanted to load your ui into something else which already existed.

mottosso:
> that approach wouldn't had worked if you wanted to load your ui into something else which already existed.

That sounds backwards to me. Are you building off of a ui file, or is the ui file building off of your Python class? Treating the ui file as a class from which to inherit sounds at least reasonable, but again I haven't used this feature so I don't know.

fredrikaverpil:
Today the functionality overwrites any existing widget:

```python
>>> widget = QtWidgets.QWidget()
>>> widget = load_ui(fname)
>>> print(type(widget))
<class 'PyQt5.QtWidgets.QMainWindow'>
```

The second argument is intended to *add* the ui elements to the already existing widget. So when you provide `self`, the ui elements are added to it without overwriting self completely.

However, I'm not 100% sure of where things are being overwritten and where they are not.

> Are you building off of a ui file, or is the ui file building off of your Python class?

Consider this... you may be writing an UI which also uses .ui files as "modules" within that UI. So, okay, this is not the best example, but bare (bear) with me...

You're making some kind of image viewer in Qt Designer to show the overall interface. So you base your main app/UI against the .ui file. Then you would perhaps want to code a widget to represent each instance of the actual image thumbnail. Ok. Let's call each such instance a "module".

Now, what if you didn't make a simple image viewer, but some other kind of viewer which would show more complex "modules". I've resorted to using Qt Designer to design such modules. This means I use an .ui for the main interface but then I also load .ui files which are not supposed to serve the main interface so to speak.

Perhaps this was overly explanatory and perhaps not answering what you were asking...?

fredrikaverpil:
The thing to focus on, is really that when you load an .ui you often want to control into which location the UI's widgets end up.

This is the main reason why I started this PR.

mottosso:
Ok, I've thought about it, and I don't think I'm not in a great position to make this call. I think what will need to happen is that we finish implementing the tests such that known corner cases are at least documented and tested, and then you'll decide whether we add it or not, @fredrikaverpil.

This can go 1 of 3 ways.

1. We add the additional argument to Qt.py
2. We move `load_ui` to it's own, standalone and optional module, `uic.py`
3. We don't add the argument

Here's what I think so far.

`load_ui` truly does add unique value not otherwise achievable without great effort. It also adds something unique to the project on a technical level, in that it adds actual implementation, rather than simply rewiring things. We'll need to decide whether that's a direction we want to take, because there might be more, and it might overtake the otherwise simple nature of Qt.py. On the upside, it makes it more useful as well. Separating it into its own module would make it more challenging to install, but at the same time insulate it from bloat for those who aren't using it.

What I don't like about it is simply how this is used. The fusing of an automatically generated GUI with something written by hand. But, having [read up on it](https://doc.qt.io/qt-5/designer-using-a-ui-file.html), this is how the Qt developers intended, so it's only fair we do the same.

I would however like to offer one final challenge to the additional argument before fully embracing it. 

You mention using one ui file to augment your subclass, and then using minor ones as standalone additions to it.

What is the below example missing.

```python
class Gui(object):
  def __init__(self):
    ui = load_ui("main.ui")

    layout = ui.body.layout()
    for page in ("page1.ui", "page2.ui"):
      widget = load_ui(page)
      layout.addWidget(page)

    # Store reference
    self.ui = ui

  def show(self):
    self.ui.show()

gui = Gui()
gui.show()
```

Let me know if you need a full runnable example, or whether this communicates the point fully.

And have a think about whether a dedicated `uic.py` module is feasible. It's possible that could solve the issue of Qt.py remaining simple and solid, whilst still enabling additional development happen on the side and remain optional to those that do not need it.

fredrikaverpil:
> The fusing of an automatically generated GUI with something written by hand.

Yes, I believe that was the design intent.

> What is the below example missing.

In your example, you first load the main UI into `self.ui`, which makes all of its widgets reside in `self.ui`. What if you wanted the rest of the widgets from `page1.ui` and `page2.ui` to be accessible from `self.ui`?

I think *that's* what's missing, with our discussions as context.

However, I don't personally think there's anything wrong with your code example and I'm not on any kind of crusade here. I think your comments on this is sound and makes sense. I'm rewriting a lot of code at the moment using `Qt.load_ui` without this second base argument and I'm doing just fine.  It's just that it seems people are using that second argument and would like to keep using it.

> This can go 1 of 3 ways.

Let me propose a fourth way :) (which I actually did propose in some other issue or PR, can't remember)

How about we store testable examples of how to use the current functionality of `load_ui` (meaning no second argument added). This would almost mean the same thing as maintaining `uic.py` but without relying on having to maintain it. This could also open up for other "examples", where we feel a pull request is more suited as an example rather than an actual implementation in Qt.py.

I think back on #101 and what you said there...

> I would much rather see software developed with cross-compatibility in mind, than software relying on a shim for missing functionality.

I feel `load_ui` today is just offering cross-compatibility and not much more. Which is great, because it provides a lot of aid when loading a .ui file. It does what Qt.py set out to do. So, adding this layer of reliability on `load_ui` I must say troubles me somewhat as well.

So, I'd say we have four options:

* We add the additional argument to Qt.py
* We move load_ui to it's own, standalone and optional module, uic.py
* We don't add the argument
* We don't add the argument, but provide a facility for advanced usage of it

mottosso:
That approach feels comfortable to me.

It means we won't have to make too many contraptions or add to much opinion to the codebase, but instead leave that for suggestions via examples and best practices. It will leave more code for the end-user. At first. But potentially save code long-term, considering obscurities and subtle differences unaccounted for or worked around in *our* codebase that would result in additional in code that way.

I'd say it'd be a defining direction for the project to take. It will not sit well with everyone. Some may choose to build something in parallel to better suit their needs and that project may very well overshadow ours. Or not. Either way, I think it might ultimately help guide the way, if only to find which route *not* to take.

So, reducing the four technical options into two, we're left with these high-level options.

1. Wrap bindings to provide a single, unified interface across all bindings.
2. Remap similarities and offer advice and best practice on how to develop software which works across bindings.

If possible, I'd like to hear from @bpabel and @hdd before dropping the hammer.

fredrikaverpil:
> If possible, I'd like to hear from @bpabel and @hdd before dropping the hammer.

I'd like that too. Ping...

fredrikaverpil:
> 2. Remap similarities and offer advice and best practice on how to develop software which works across bindings.

Having given this some thought, I'm personally leaning towards this.

hdd:
Hi , sorry for the delay on joining the chat , just come back from a quick break.

To me , and how I did started approaching the patches, the best way to go, is to start with the new api in mind (so, relying on QtWidgets), and provide patched methods to be able to run the legacy code (PySide/PyQt) (eg: #98 ).

This allow the users to start using , and getting used to the new api, but having the ability to have their code working with not much to change. 

We should, though, warn the users, that they are using these patched methods (whether are forward or backward compatibility), so they are aware.

My2c.
L.

mottosso:
> We should, though, warn the users, that they are using these patched methods (whether are forward or backward compatibility), so they are aware.

Could you open an issue about this? I see a number of conversation topics on how to achieve this technically.

hdd:
done, see #109

mottosso:
Ok, let's commit to the goals discussed in this issue!

Henceforth, Qt.py is about linking together commonalities between PySide2 and other bindings, and documenting differences, so as to foster a mindful developer of cross-compatible software.

This is in contrast to wrapping all functionality that differs PySide2 with our own ad-hoc implementations, which we believe will cause more problems than it solves long-term.

For alternative ideals, see the following forks and projects.

- https://github.com/mottosso/Qt.py/pull/107
- https://bitbucket.org/efestolab/qtext
- https://github.com/spyder-ide/qtpy 

Any more? Let us know!

fredrikaverpil:
For anyone reading this enormous thread, check out this pull request: https://github.com/mottosso/Qt.py/pull/119

PR #82, opened 2016-07-30
Feature/bindings info

### LINKED ISSUE #73: Add links to binding repos
I was just informed that the PySide2 development by the Qt Company is not being done over at the [Github PySide2 repository](https://github.com/PySide/pyside2). Instead it is being carried out here: http://wiki.qt.io/PySide2

I think it would be useful to include such info somewhere, probably in the README.

Basically, for each binding make sure it's easy to find the github repo ... or:

- source / binary distributions
- bug tracker
- mailing list (or equivalent)

fredrikaverpil:
Fix #73 

Not sure if these links will just be tedious to keep updated. I think they're good to have gathered in one place when developing with Qt.py.

You decide whether we should add these to the repository or not 😄


mottosso:
Yes, good question.. I can see the value, but also the mess and potential for broken links.

Let's keep it open for a while and think it over.

mottosso:
Added this to the [wiki](https://github.com/mottosso/Qt.py/wiki/Links-and-resources).

Let me know what you think, we can re-open this one if you want.

fredrikaverpil:
That's fine! 👍

PR #83, opened 2016-07-31
Add external guides to README

mottosso:
These should be in the README I think. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,11 @@ button.show()
 app.exec_()
 ```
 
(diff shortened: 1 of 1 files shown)


fredrikaverpil:
Sounds good!

It seems you got the `load_ui` changes into this PR by mistake.

mottosso:
> It seems you got the load_ui changes of the README into this PR by mistake.

So I did, well spotted, fixing now.

PR #95, opened 2016-08-09
wrap QtWebkit in try/except as is an optional library and might not b…

hdd:
Wrap QtWebKit in QtWebKitWidgets to match Qt5

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,4 +34,5 @@ setup(
     zip_safe=False,
     py_modules=["Qt"],
     classifiers=classifiers,
+    setup_requires=['nose>=1.0']
 )
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,5 @@ setup(
     license="MIT",
     zip_safe=False,
     py_modules=["Qt"],
-    classifiers=classifiers,
-    setup_requires=['nose>=1.0']
+    classifiers=classifiers
 )
--- a/tests.py
+++ b/tests.py
@@ -188,15 +188,3 @@ if PYTHON == 2:
             import sip
             sip.setapi("QString", 1)
             assert_raises(ImportError, __import__, "Qt")
-
-
-def test_QtWebKitWidgets_available_in_pyside():
-    with pyside():
-        from Qt import QtWebKitWidgets
-        assert QtWebKitWidgets
-
-
-def test_QtWebKitWidgets_available_in_pyqt4():
-    with pyqt4():
-        from Qt import QtWebKitWidgets
-        assert QtWebKitWidgets
--- a/tests.py
+++ b/tests.py
@@ -190,7 +190,13 @@ if PYTHON == 2:
             assert_raises(ImportError, __import__, "Qt")
 
 
-def test_qtwebkit():
+def test_qtwebkit_pyside():
     with pyside():
+        from Qt import QtWebKitWidgets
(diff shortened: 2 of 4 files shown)


fredrikaverpil:
This is awesome, thank you @hdd ! 😃 

@mottosso this looks good to me.

fredrikaverpil:
@hdd do you think you could also provide tests in [tests.py](https://github.com/mottosso/Qt.py/blob/master/tests.py)?

hdd:
Sure thing I can, This was just a "early" pull request to show you where I was going.
I'll try to put in place some tests for this.

hdd:
Hope you don't mind, as part of this ticket I improved a bit the setup.py so now you can run:

```
python setup.py nosetests
```

setup.py
```diff
@@ -34,4 +34,5 @@
     zip_safe=False,
     py_modules=["Qt"],
     classifiers=classifiers,
+    setup_requires=['nose>=1.0']
```
mottosso: This line warrants its own PR, documentation and tests I think. Could we remove this for now, and implement it fully later on?

tests.py
```diff
@@ -188,3 +188,15 @@ def test_sip_api_already_set():
             import sip
             sip.setapi("QString", 1)
             assert_raises(ImportError, __import__, "Qt")
+
+
+def test_qtwebkit_pyside():
```
mottosso: These tests aren't testing anything of value but rather adds bulk to the project. If we want to test that members of Qt.py are mapped (or not) we should probably do so all together in a single test function.

mottosso:
> Hope you don't mind, as part of this ticket I improved a bit the setup.py so now you can run:

That's allright, but it can be tricky to keep conversations straight when more than one feature is being added.

Great with an example of the end result, but what does `python setup.py nosetests` do? Haven't used it before.

hdd:
I'll be removing it .
Added in the rush of simplifying the testing but better keep it for a separate ticket.
Overall though having nosetests in setup py allows you to wrap flags and features so you don't have to remember them.

setup.py
hdd: it'll be removed, in case we'll be opening a ticket to improve the setup.py

tests.py
hdd: Usually is better tests separated from each other (one test , one assert) , but of course there are exceptions.

Ops, noticed now, half of the test went away before the commit.

mottosso:
> Overall though having nosetests in setup py allows you to wrap flags and features so you don't have to remember them.

I like that, at the moment, these flags are documented in the Dockerfile, but having them in setup.py is more discoverable.

Let's open up a separate PR for it, and make sure to remember nosepipe, --verbose, --with-process-isolation and --with-doctest as well.

tests.py
```diff
@@ -188,3 +188,15 @@ def test_sip_api_already_set():
             import sip
             sip.setapi("QString", 1)
             assert_raises(ImportError, __import__, "Qt")
+
+
+def test_QtWebKitWidgets_available_in_pyside():
```
mottosso: Let's remove these tests, they aren't doing any good.

We could start testing with runnable examples, but I'm uncertain about what that would achieve also.

mottosso:
This looks good, apart from that one inline comment!

Issue #98, opened 2016-08-10
QCoreApplication::UnicodeUTF8 is deprecated


hdd:
Found this :https://wiki.qt.io/Transition_from_Qt_4.x_to_Qt5
about this issue.

This gets raised, when trying to use compiled widget (pyside-uic....), which has the retranslateUi() .

``` PySide
QApplication::translate("Href_Gui", "Text:", 0, QApplication::UnicodeUTF8)
```

```PySide2
QApplication::translate("Href_Gui", "Text:", 0)
```

Not really sure how to tackle this down.

hdd:
guess we can implement a mocked QApplication.UnicodeUTF8 property
and provide an override for PySide2/Qt5 on translate, so they can accept an argument more.

Thoughts ?

hdd:
I've been opening a pull request : https://github.com/mottosso/Qt.py/pull/99

mottosso:
Sorry to ask, but is this a problem you need solved? Or are you trying to cover all incompatibilities?

hdd:
I'm porting a quite complex stack of code over Qt.py and it was using compiled ui ( designer-qt4 ) through pyside-uic code, which by default sneak in the translate function.

Without this fix, any interface build with PySide/PyQt won't simply be loading, and error, and possibly make the app segfault.

mottosso:
Ah, excellent.

An example in the form of a test would be lovely to really drive this feature home.

hdd:
thought the same, but I need #93 first .
So we can test both.

mottosso:
#93 implemented.

hdd:
tests are now available

mottosso:
For completeness, the missing piece of information is an example that illustrates the problem.

mottosso:
Implemented in [0.5.0](https://github.com/mottosso/Qt.py/releases/tag/0.4.0).

@hdd, let me know how this works for you!

hdd:
Hi Marcus, 
Does this solve also the UnicodeUTF8 Remapping issue on QApplication ?
(as it can't simply be ignored on old PyQt4/PySide code)

mottosso:
Hey @hdd,

It solves it by compiling your .ui files with PySide2. That compiler won't produce references to this member.

mottosso:
[...]

Issue #90, opened 2016-08-06
backward compatible to allow pyside1 code to work with pyside2


artbycrunk:
this module could benefit from adding support 
for a code base that is in PySide 1 but installed version is pyside2, 
as is the case of Maya 2017 for using legacy tools created for pyside1.

so basically doing a remap from qtwidgets back to qtgui.

mottosso:
Thanks for bringing this up, it is one of the two routes to take in a compatibility layer such as this. 

We can either:

1. Bring the past to the future, or
2. Bring the future to the past

Qt.py is about enabling future software to coexist with an old platform as we transition away from an old ways of doing things.

It's important we have an open discussion about it I think. Spontaneously, I think I would rather see the kind of project you have in mind as a separate project, as it would likely come with its own conflicting requirements.

The benefit with what you are looking for is, as you say, a direct mapping for old software to run on a new platform. But there is a limit to how far we can take this. At some point, there *will* be new technology that will be *unable* to faithfully replicate old behaviour.

What that means for an organisation is that you can continue to run your legacy software on new technology, such as Maya 2017. But at some point you will be back to where you are today and will again have to refactor. Only this time you are in even deeper trouble due to hoop jumping through compatibility hoops.

As such, my stand on the topic is that this feature would not solve as many problems as it would create. That it would postpone the inevitable. As a compatibility layer it is this project's responsibility to ensure a long-lasting, maintainable and future-proof product.

That said, I am open to letting the conversation the follows determine which route to go.

artbycrunk:
Well, the perspective I was talking in is that currently there might be legacy software that doesn't need an update to pyside2 module structure but still needs to be able to run in a pyside2 environment. A case of this could be time investment in the software is not really feasible, and can just run it as in till its end of life with the features it provides. Like you said there is obviously a limit on how far we can take this, as expecting things to remain the same all the time, not the right mentality to have in any context. 

I agree with you when you say that this feature will postpone the inevitable, but it also will give the devs the ability to make a choice for backwards compatibility as much as possible, until the codebase is no longer able to sustain it.

So to sum it up, one can actively choose the direction required in the software, 

1. Keep the software as is, without needing to change to qtWidgets, for the reason that the tool can gracefully reach its end of life.
2. Change to using qtWidgets as and when updates to the software are in progress, while still allowing the dev the freedom to keep part of the codebase in the old qtGUI style.
[...]

PR #99, opened 2016-08-10
Unicode utf8 is deprecated

hdd:
Please consider this (as much as the others) as an early pull request.
Put in place so we can chat on top of it.

It does lack of tests at the moment , but for the ones I've been running, seems to be able to leverage the differences.


hdd:
tests has now been added to check backward compatibilty

tests.py
```diff
@@ -208,3 +208,81 @@ def test_sip_api_already_set():
...
+
+def test_translate_and_UnicodeUTF8_in_pyside():
+    with pyside():
+        from Qt import QtWidgets
+
+        # this does exist in PySide
+        assert QtWidgets.QApplication.UnicodeUTF8 is not None
+
+        # use patched method with old arguments
+        QtWidgets.QApplication.translate(
```
mottosso: Would it be possible to include a return value in the test, to make sure it produces the expected value?

hdd: Sure thing, just wanted to check the overall testing first (which is failing).
I'll add a check of the return value once I have something basic working ;)

mottosso: Ah ok. No problem.

As tests now require all bindings, I'd imagine it can get tricky to replicate this environment locally. The developer guide has instructions on how to run these via Docker, if you have access to this. It'll let you run tests locally without installing any bindings, exactly the way Travis is running it.

hdd: cool I'll have a look, if you don't mind though , for now, I'll hammer a bit your remote testing ;)

mottosso:
For unicode strings, try putting a `u` infront of the strings. E.g. `u"MyString"`.

Python 3 is picky with that.

tests.py
hdd: checking now on local docker installation so I don't mess to much up with travis

hdd:
Cheers for that, keep forgetting about py3k and unicode stuff.... (not using py3k at all really ;))
installed all the docker locally and tests are now passing.
pushing the fixes.

hdd:
added tests for return values.

Qt.py
```diff
@@ -204,18 +225,18 @@ def _init():
...
     """
-
+    
     preferred = os.getenv("QT_PREFERRED_BINDING")
     verbose = os.getenv("QT_VERBOSE") is not None
     bindings = (_pyside2, _pyqt5, _pyside, _pyqt4)
 
     if preferred:
-
+
```
mottosso: Have a look and see if you can't remove these whitespaces inbetween lines. A linter could help here.

hdd: always.
it has now been removed, and code cleaned up from trailing spaces.

mottosso:
So just to confirm, is it the case that PySide, PyQt4, and PyQt5 all behave like PySide2 now? I wouldn't expect to see any changes to PySide2 on this, but I haven't had a close enough look yet as to what that change is or whether it's needed.

hdd:
worth closing this one as well.
Cheers.

mottosso:
Let's not get hasty. :) It looks like we do need this in order to develop for Qt 5.

- https://wiki.qt.io/Transition_from_Qt_4.x_to_Qt5#QCoreApplication::UnicodeUTF8_is_deprecated

Only it looks like you've patched PySide2 and PyQt5 to conform to PySide and PyQt4, when it should have been the other way around.

Remember, PySide2 is the reference implementation, right after the official docs of Qt 5.

- https://github.com/mottosso/Qt.py#rules

Having said this, I'd still be keen to see an actual example of it in use, and where it causes a problem. You [mention](https://github.com/mottosso/Qt.py/issues/98#issuecomment-239163592) you are using it with pyside-uic? Could you provide an example?

fredrikaverpil:
I have second what @mottosso said; we need example code to fully grasp what this is about.

hdd:
Hi, unittests are provided to test and show forward and backward compatibility approaches.

If you compile a ui, (pyside-uic <.ui file> -o <.py file>) is quite easy to spot (thought I did mention at the beginning of the ticket...) In any case, Mr Google can help on this : http://stackoverflow.com/questions/4442286/python-code-genration-with-pyside-uic

>> It looks like we do need this in order to develop for Qt 5.

Yeap that's why I did work on this.

>> Only it looks like you've patched PySide2 and PyQt5 to conform to PySide and PyQt4, when it should have been the other way around.

You simply can't, as PySide2 and PyQt5 remove an argument from the function and a property from the object, therefore is not possible doing this on PySide and PyQt4, what this does, is to provide PySide2/Qt5 with this extra argument, so doesn't break, same for the property, which is now missing .

Out of curiosity, how would do the other way round ?

Marcus, I'm not being hasty just objective. #96  has been closed for good, and is fine for me, but either I get to use a  module which allows me to do the work I need (and therefore I'm happy to contribute back) , or I better put my efforts on something separated (which is exactly what I did started doing yesterday eve):
https://bitbucket.org/efestolab/qtext, which simply use Qt.py as dependency ad does patch for what I need.

mottosso:
> In any case, Mr Google can help on this

I would very much rather see how *you* encounter this issue, and how a solution might help you. What I need is to be able to reproduce the problem and see that we're really solving something that needs solving. I expect this from every single issue and pull-request and I don't think it's too much to ask.

Tell me this, are you expecting a shim providing a PySide2 binding to work with a GUI generated for PySide(1)?

If so, may I direct you to #90 followed by #107? The problem being that the code produced by pyside-uic is for Qt 4, whereas this shim enables code written for Qt 5 to also run on Qt 4.

Alternatively, can you compile your UI files with the PySide2 or PyQt5 equivalent? This should work well on Qt 5, and whatever problems you have running this code on Qt 4 that's what we should be resolving here.

Until I see a reproducible example, this is a non-issue.

mottosso:
Hi @hdd, I have got some time this weekend to dive into this.

If you could share a snippet of code that breaks for you without this fix, then I can have a look at making it work!

salvaom:
Hello!

I believe that the problem encountered by @hdd is the same I've been having. I've uploaded a gist to illustrate: [https://gist.github.com/salvaom/afe312982140c926c21005d822c0f116](https://gist.github.com/salvaom/afe312982140c926c21005d822c0f116). _widget_designer1.ui_ is created with Designer from PySide(1) and _widget_designer2.ui_ with Qt Creator Community (Qt 5.6.1). I've compiled only the first one (since they are identical) with the pyside-uic(1) from PyPI and with Maya's (2017) pyside2uic.

I also have a thousand .ui files compiled for various applications and it's not a big deal to change all _from PySdie ..._ to _from Qt ..._, but the UTF8 issue remains since the syntax varies from one PySide to another.

I think this subject was also mentioned in [#129](https://github.com/mottosso/Qt.py/issues/129)

mottosso:
Hi @salvaom, thanks for this!

With #129 in mind, in addition to search-and-replacing `PySide2` with `Qt`, do you think it would also make sense to replace calls made to `QApplication.translate` with our own version, at for example `Qt.translate`?

That way we might be able to sidestep the issue of monkey patching the class.

salvaom:
Hey @mottosso, I thought about that as well, but the problem still remains since PySide2 does not have _QtWidgets.QApplication.UnicodeUTF8_, so even if I were to change _QtGui.QApplication.translate_ to _Qt.translate_, the code will still error on the last argument in files compiled with PySide(1).

The only solutions I can think of are either what @hdd has proposed or if there is no intention of monkey patching the module itself, having a variable called (for example) _UnicodeUTF8_ which in PySide is a reference to _QtGui.QApplication.UnicodeUTF8_ and on PySide2 just -1, this way it would be a matter of changing _QtGui.QApplication.UnicodeUTF8_ to _Qt.UnicodeUTF8_.

mottosso:
Ah, I see.

How about this; in addition to search-and-replacing `translate()` with our own, we also remove this last argument?

```python
>>> line = """self.label.setText(QtGui.QApplication.translate("uic", "Bienvenidos a Espa?a", None, QtGui.QApplication.UnicodeUTF8))"""
>>> new_line = line.replace("QtGui.QApplication.translate", "Qt.translate")
>>> new_line = new_line.replace(", QtGui.QApplication.UnicodeUTF8", "")
>>> print(new_line)
'self.label.setText(Qt.translate("uic", "Bienvenidos a Espa?a", None))'
```

mottosso:
>  if there is no intention of monkey patching the module itself, having a variable called (for example) UnicodeUTF8 which in PySide is a reference to QtGui.QApplication.UnicodeUTF8 and on PySide2 just -1, this way it would be a matter of changing QtGui.QApplication.UnicodeUTF8 to Qt.UnicodeUTF8.

Sorry I just re-read this, yes! This sounds better!

mottosso:
With #131, this should be solved.

@salvaom, @hdd could you let me know whether this works for you?

salvaom:
Hello!

Not quite, for what I'm seeing convert only supports .ui files compiled with PySide2 but the problem I (and I belive @hdd as well) was having was the thousand (approx) of .ui files compiled with PySide(1) I've got that I would like to have converted to Qt.py.

But even if we were to add [another](https://github.com/mottosso/Qt.py/blob/master/Qt.py#L65) two conversion lines in the _convert()_ function, the problem still remains since the _QtGui.QApplication.translate_ does exists in both PySides with the same argument count but the variable _QtGui.QApplication.UnicodeUTF8_ doesn't in PySide2, which inevitably makes any .ui file compiled with PySide(1) fail, no matter the _translate_ function that is being used.

I can make a pull request that includes the changes I think would be necessary to make Qt.py compatible with any PySide(*) compiled .ui file, let me know.

Just to better illustrate:

**PySide**

```
>>> import PySide
>>> PySide.__version__
'1.2.2'
>>> from PySide import QtGui
>>> QtGui.QApplication.translate('MyText', 'MyText', None, QtGui.QApplication.UnicodeUTF8)
u'MyText'
>>> import Qt
>>> Qt.__version__
'1.2.2'
>>> from Qt import QtWidgets
>>> QtWidgets.QApplication.translate('MyText', 'MyText', None, QtGui.QApplication.UnicodeUTF8)
u'MyText'
>>>
```

**PySide2**

```
>>> import PySide2
>>> PySide2.__version__
'2.0.0~alpha0'
>>> import Qt
>>> Qt.__version__
'2.0.0~alpha0'
>>> from PySide2 import QtWidgets
>>> QtWidgets.QApplication.translate('MyText', 'MyText', None, QtWidgets.QApplication.UnicodeUTF8)
AttributeError: type object 'PySide2.QtWidgets.QApplication' has no attribute 'UnicodeUTF8'
>>> Qt.QtCompat.translate('MyText', 'MyText', None, QtWidgets.QApplication.UnicodeUTF8)
AttributeError: type object 'PySide2.QtWidgets.QApplication' has no attribute 'UnicodeUTF8'
>>> QtWidgets.QApplication.translate('MyText', 'MyText', None, -1)
u'MyText'
>>>
```

So I think this is still the best option:

> if there is no intention of monkey patching the module itself, having a variable called (for example) UnicodeUTF8 which in PySide is a reference to QtGui.QApplication.UnicodeUTF8 and on PySide2 just -1, this way it would be a matter of changing QtGui.QApplication.UnicodeUTF8 to Qt.UnicodeUTF8.

Or in any case to _Qt.QtCompat.UnicodeUTF8_

mottosso:
Hey @salvaom,

> thousand (approx) of .ui files

What about compiling them with PySide2?

All of your existing Qt Designer files should work well with Qt.py, given you compile them with PySide2, the reference binding.

I'd have a chat with @artbycrunk in #90 and #107 about joining the effort in making PySide(1) code run on other bindings, I can certainly see the demand for it. But personally, and with this project, the philosophy is to enable future code to run on older platforms, not the other way around.

It opens your code up to advances to Qt, whilst enabling you to transition smoothly across. Consider it?

salvaom:
Hello!

I understand and share the philosophy, but even if I was to compile with PySide2 all files again, the problem remains since _translate_ in PySide(1) accepts as a 4th argument only an instance of _QCoreApplication::Encoding_, which has [only three possible values](http://doc.qt.io/qt-4.8/qcoreapplication.html#Encoding-enum) and _-1_ is not one of them, so as you can probably guess already, running a PySide2 compiled file with a PySide(1) environment yields an error:

```
>>> from Qt import QtWidgets
>>> QtWidgets.QApplication.translate('Foo', 'Foo', None, -1)

  File "<pyshell#8>", line 1, in <module>
    QtWidgets.QApplication.translate('Foo', 'Foo', None, -1)
TypeError: 'PySide.QtCore.QCoreApplication.translate' called with wrong argument types:
  PySide.QtCore.QCoreApplication.translate(str, str, NoneType, int)
Supported signatures:
  PySide.QtCore.QCoreApplication.translate(str, str, str = None, PySide.QtCore.QCoreApplication.Encoding = CodecForTr)
  PySide.QtCore.QCoreApplication.translate(str, str, str, PySide.QtCore.QCoreApplication.Encoding, int)
```

I understand your concern, and if there is another way to skip this I'm open to try it out and embrace it, but so far this is the only way I can think of making this work for both PySide(1) and PySide2.

Please let me know what you think.

mottosso:
> the problem remains since translate in PySide(1) accepts as a 4th argument 

Ah! Sorry, I should have made myself more clear. This has actually already been addressed in the fix above!

Qt.py ships with it's own `translate` that the conversion redirects the original calls to.

See here:

- [`QtCompat.translate`](https://github.com/mottosso/Qt.py/blob/67e34d5a9c3d5a0dafac8f8e03d04baafa476ae6/Qt.py#L231)
- [`--convert`](https://github.com/mottosso/Qt.py/blob/67e34d5a9c3d5a0dafac8f8e03d04baafa476ae6/Qt.py#L66)

The result is a `.py` file with no calls to the original `.translate`.

I'd love for you to try compiling just one, and see whether that works for you. The goal is full compatibility with any and all Qt Designer files. If something isn't working right, then that's a bug and we'll need it squashed.

mottosso:
If this remains an issue, feel free to re-open!

PR #107, opened 2016-08-12
add base prototype for backward compatible Qt module.

artbycrunk:
i liked the QtLegacy idea, as it can isolate the above cases.
Here is something, to begin with...
prototype #90

### DIFF
--- a/QtLegacy.py
+++ b/QtLegacy.py
@@ -1,0 +1,31 @@
+"""Map all bindings to PySide
+
+This module is a backward compatible for of Qt.py
+
+Resolution order:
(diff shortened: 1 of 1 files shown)


mottosso:
This could work. As a separate module altogether, it can live side-by-side and remain opt-in without interfering with the Qt.py module.

Let the tests begin.

Try to capture common usage scenarios in the form of tests. See existing tests for examples.

mottosso:
I see no reason to close this, but would rather keep it open for ideas, suggestions and contributors in this direction for the project.

Issue #93, opened 2016-08-08
Qt 5 and continuous integration


mottosso:
### Goal

Enable testing of Qt.py with Pyside2 and PyQt5.

### Motivation

At the moment, testing only happens with PySide and PyQt4, but the codebase and tests remain open for expansion towards bindings for Qt 5. The reason they haven't yet seen the day of light has been due to difficulties in installing each binding in an efficient manner.

PyQt4 and PySide both have pre-compiled counterparts that take a matter of seconds to install, whereas Qt 5 bindings are more complex, especially PySide2, requiring manual compilation. (On Python < 3.5).

Docker has always been an option, but the problem has been losing out on being able to concurrently run tests, one build per version of Python. But this technique should enable greater concurrency, and greater distribution of tasks with one build per version of Python and binding.

### Implementation

The idea is simple. Utilise the build matrix functionality to provide many identical environments with varying environment variables. The environment variables can then be used to programatically select an appropriate Docker image.

mottosso:
For the one interested in implementing this, here are a few hints.

1. Modify the current Dockerfile to provide each binding for a single version of Python, Ubuntu 14 defaults to Python 2.7, whereas 16.04 I think defaults to 3.4, the next to 3.5 and so on.
2. Rename this to e.g. Dockerfile.py27 and create another one for 2.6, 3.4 and 3.5. These should be enough
3. In the .travis.yml, add each Python version. The version will then be available as an environment variable.
4. In the test section, use this variable to select an appropriate Dockerfile, build it, and run it.
5. Once it works, we can upload each image to the docker hub, such that each becomes a download, rather than being built each time. This should shave off lots of time per build.

hdd:
Thanks for this, does this also implement the contextmanager for PySide2 and Qt5 in tests ?

mottosso:
It does, yes. See the top of tests.py.

hdd:
yeap, sorry , forgot to merge upstream back.

PR #108, opened 2016-08-13
Qt 5 support for tests

mottosso:
Implements #93.

Here we go. Tests running via 2.7 and 3.5, the two flavors most easily installed via Ubuntu 16.04.

This does limit us to these two exact versions, as well as the exact versions that happens to be in their package management system. At some point, we will either need to upgrade to the next Ubuntu and thus indirectly also upgrade to newer packages, or compile ourselves and upload images to Docker hub. Whichever comes first. I planted build scripts for PyQt5 and PySide2 in the git history for that time.

The main thing is that the pipeline works and that it remains quick to test.

Some tests will break, as it assumed no Qt 5 bindings. Will work on those next.


mottosso:
There we go. Parallel builds on Travis, with Docker.

I've excluded AppVeyor (i.e. testing on Windows) for now, as getting all builds ready for both Python 2 and 3 on a single platform (Ubuntu) was quite a lengthy process, and I'd rather get this out the door first.

fredrikaverpil:
I saw this just now. This is awesome. :+1: :+1:

Issue #109, opened 2016-08-22
Warn users about Qt.py-only functionality


hdd:
As a user and developer I would like to be notified when using backward or forward compatibility  methods when using them.

hdd:
Worth using python warning system , https://docs.python.org/2/library/warnings.html
so we can filter them through the logging system as well.

mottosso:
Thanks @hdd.

I agree this would be good and I've got in mind one way of achieving this, but it comes at a cost. The cost is likely too high, so I'm interested in alternative ways of achieving this result.

#### The idea

The inherent problem of triggering anything when accessing a member variable, such as `QtWidgets.QWidget` versus `QtGui.QWidget` is that there isn't any code inbetween asking for this member and getting it.

The idea is to convert `QtWidgets` from being a module into being an object and override `__getattr__` such that asking for a member would allow us to trigger these warnings.

```python
class QtWidgets(object):
  def __getattr__(self, attr):
    warnings.warn("Not backwards compatible!")
    return getattr(PyQt5.QtWidgets, attr)

Qt.QtWidgets = QtWidgets()
```

I'm a bit weary however of doing this, as we would move away from wrapping a module - e.g. `PySide.QtGui` - with another module to wrapping it with a class, possibly and most likely breaking all sorts of things.

#### Alternative

So, what's another way of achieving this?

hdd:
agree that would be the ideal way of doing it , but I found issues on wrapping the classes and replace them in place as you are describing. Especially when you retrieve the class from another instance and not from the top module (Qt).  Using unbounded methods (function patching) seems to be working fine so far. 

I would suggest to keep going with unbound methods for the time being, while looking into a proper way to have the full widget/object class patched with bound methods, so users can use and test the module and implementation in the meanwhile.

mottosso:
> Using unbounded methods (function patching) seems to be working fine so far. 

Could you show an example of this?

hdd:
You can see this in use on #98 where I patch the translate function as well as in #96 for the headers.
This allow you to retrieve the header from the instanced object and access the patched method.
These patched methods results to be unbounded.

mottosso:
Sure, that shouldn't be a problem.

Did you not also want warnings when users try and use, e.g. `QtGui.QWidget` when they should have used `QtWidgets.QWidget`? Or was it strictly related to direct function calls?

hdd:
[...]

Issue #101, opened 2016-08-10
code structure reorganisation proposal


hdd:
After having spent some time trying to get around the various differences in the framework, has become pretty clear to me, that all these in place replacements (still many to come I fear), might bring soon the code to a point where is going to be hard to maintain, especially if many other contributors are coming in.

Here an idea on how the code could be reorganised :

```
├── docs
│   └── source
├── setup.py
├── source
│   └── Qt
│       ├── __init__.py
│       └── bindings
│           ├── __init__.py
│           ├── _pyqt4_.py
│           ├── _pyqt5_.py
│           ├── _pyside2_.py
│           └── _pyside_.py
└── tests
    ├── test_pyside2.py
    ├── test_pyside.py
    ├── test_qt4.py
    └── test_qt5.py
```

Any thought on this ?
Cheers.

fredrikaverpil:
Hm. Do we really need to extract the entire bindings like that?

I'm thinking there's going to be a lot of simple remappings and then we're going to add classes and functions to support remappings. I think it could be valuable to have all remappings in its own file for overview purposes. Then make custom classes etc into submodules.

```
├── docs
│   └── source
├── setup.py
├── source
│   └── Qt
│       ├── __init__.py
│       └── lib
│           ├── __init__.py
│           ├── loadUi_.py
│           ├── custom1_.py
│           └── custom2_.py
└── tests
    ├── test_pyside2.py
    ├── test_pyside.py
    ├── test_qt4.py
    └── test_qt5.py
```

But as a whole, I think it's probably inevitable that we need to split Qt.py up into pieces at some point.

hdd:
Yeap, same here, soon or later it'll have to happen so I think is better look into it asap.

The idea of splitting the binding is my main point for the restructure, so we can have dedicated modules for each one, and we can keep any patch/fix scoped properly. Could this be leading to some code duplication ? Sure, there's the risk, but the feeling is that it'll bring more benefits than problems.

mottosso:
Thanks for bringing this up, it was bound to be discussed at some point and its a healthy discussion to have. Both now and for the future.

But before this train of though travels any further, I'll say; no. This isn't a direction I'm interested in taking this project.

This will instead be what separates Qt.py from other project of similar goals, the driving motivation being dead simple adoption, use and maintenance.

[...]

PR #110, opened 2016-08-24
Implement #109

mottosso:
Here's one way to solve #109.

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -243,7 +243,7 @@ AttributeError: 'PySide2.QtWidgets.QHeaderView' object has no attribute 'setResi
 >>> header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
 Traceback (most recent call last):
 ...
-AttributeError: 'PySide.QtWidgets.QHeaderView' object has no attribute 'setSectionResizeMode'
+AttributeError: 'PySide.QtGui.QHeaderView' object has no attribute 'setSectionResizeMode'
 ```
 
 ##### Workaround
@@ -257,8 +257,8 @@ Use a conditional.
 >>> view = QtWidgets.QTreeWidget()
 >>> header = view.header()
 >>> if __binding__ in ("PyQt4", "PySide"):
-...   header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
-... else:
 ...   header.setResizeMode(QtWidgets.QHeaderView.Fixed)
+... else:
+...   header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
 ```
 
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -216,3 +216,49 @@ Traceback (most recent call last):
 ...
 TypeError: ...
 ```
+
+
+#### QtWidgets.QHeaderView.setResizeMode
+
+`setResizeMode` was [renamed](http://doc.qt.io/qt-5/qheaderview.html#setSectionResizeMode) `setSectionResizeMode` in Qt 5.
+
+```python
+# PySide2
+>>> from Qt import QtWidgets
+>>> app = QtWidgets.QApplication(sys.argv)
+>>> view = QtWidgets.QTreeWidget()
+>>> header = view.header()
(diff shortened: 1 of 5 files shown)


hdd:
Hi , not sure I'm getting this commit.
How does this solve #109 ?
Can't see any warning for the users.

mottosso:
Hey, It's an attempt at solving the underlying issue. About users not knowing what is and isn't being added by Qt.py.

Solving it via the documentation route.

It's just a thought. The problem I see with warnings, is that it might get inconsistent; where only the things we add that are pure functions can give warnings, whereas other things, such as plain attributes ala `__binding__` won't actually throw a warning. This way, we could potentially cover all bases.

mottosso:
Ah, darn. This should have gone into it's own PR.

mottosso:
Ok, done. Let me know what you think @fredrikaverpil! Especially about the [README](https://github.com/abstractfactory/Qt.py/blob/7d3bc1c5626852a4fe1800d945a726a4feda9ec0/README.md). My motivation was to better highlight important bits, i.e. docs, goal and usage.

fredrikaverpil:
Ok let's see. So what was being discussed is whether we should print warnings to explain whether e.g. `QtWidgets.QWidget` or `QtGui.QWidget` is being actually called, right?

Like mentioned in https://github.com/mottosso/Qt.py/issues/109#issuecomment-241363172 we're just making simple remappings and there's no code inbetween to catch such events. Also, I'm not sure I understand why such warnings being printed is useful - or if we could at all implement such a feature reliably. The sole purpose of this shim is to just take care of this as good as possible "under the hood" and provide the means to debug and catch these kinds of things yourself when needed. 
@hdd Perhaps I'm missing something?

I agree that taking the documentation route makes more sense to me as well.
I've read the README and I think it looks good, but I think it doesn't fully communicate what was being discussed in #101 which perhaps would be a good time to address;

* We'd like to keep Qt.py simple, simple to grasp/read through (and therefore simple to maintain).
* We don't want software to rely on functionality provided by Qt.py other than the actual compatibility features (remappings).
* Instead of building custom (and sometimes advanced) convenience functionality into Qt.py, the primary goal is to create minimal compatibility between bindings (in most cases remappings).

I think perhaps these incentives should go into some kind of "development vision" under the development guidelines.

I've been busy lately but I've been meaning to set up a PR of a bunch of different example usage of the `load_ui` function (https://github.com/mottosso/Qt.py/pull/81), since I believe we shouldn't add the more advanced secondary arguments to that function because of the bullet points above. Perhaps in an `EXAMPLE_USAGE.md` or in its own folder `examples` containing gist-like examples.

Taking all of that into consideration, an "example" could perhaps show how to print warnings for a specific use case but I think you indirectly already covered this well in the README. However, it wouldn't hurt to be more allowing towards new examples in comparison to PRs towards actual functionality of Qt.py which needs to be thoroughly considered and discussed.

So this all looks good to me.
@hdd I'm all ears to hear what you have to say about all of this.

mottosso:
> I've read the README and I think it looks good, but I think it doesn't fully communicate what was being discussed in #101 which perhaps would be a good time to address;

Agreed. I've added "Development goals", with your two last points condensed into what I think is what we're trying to say. Let's add/remove to this list as our picture becomes clearer.

> Perhaps in an EXAMPLE_USAGE.md or in its own folder examples containing gist-like examples.

How about actual gists? That way, the library would be in our README.

E.g.

- [Coping with differences](https://gist.github.com/mottosso/41bf9c3ec622602f984ecf4e1d05dce1)
- [Testing cross-compatible code](https://gist.github.com/mottosso/41bf9c3ec622602f984ecf4e1d05dce1)
- [Catch error](https://gist.github.com/mottosso/41bf9c3ec622602f984ecf4e1d05dce1)

It would on the other hand enable gists to change under our noses, and prevent anyone from contributing.

An `examples/` directory isn't a bad idea.

fredrikaverpil:
Looks great!

I think gists (although changing/disappearing) is a good idea but I think the `/examples` approach is even better. It'll be direct and to the point to ask contributors put their PRs into an "example" rather into an actual implementation instead of asking them to go and make a gist instead.

PR #112, opened 2016-08-27
Remap setSectionResizeMode to Qt 4

mottosso:
### Goal

Enable Qt 4 bindings the use of `setSectionResizeMode`.

```python
# PySide
import sys
from Qt import QtWidgets
app = QtWidgets.QApplication(sys.argv)
widget = QtWidgets.QTreeView()
qheaderview = widget.header()
qheaderview.setSectionResizeMode(qheaderview.Fixed)
widget.show()
app.exec_()
```

- [Reference](https://doc.qt.io/qt-5/qheaderview-obsolete.html#resizeMode)

### Rationale

Originally offered [here](https://github.com/mottosso/Qt.py/pull/96) but through misunderstandings it never saw the light of day.

This enables a PyQt5/PySide2 implementation to run on PyQt4/PySide, without altering the behavior of those bindings.

### Implementation

A simple remapping. See the test and implementation for details!

@fredrikaverpil let me know what you think.


mottosso:
This works now, but there is a problem.

Importing PySide before Qt.py under Python 3 has side-effects which prevents the remapping from taking place. This may be due to some of the magic happening upon importing PySide and I'm expecting this to be outside of our control.

The exposing lines are [here](https://github.com/mottosso/Qt.py/pull/112/commits/12d55ae65362f01e957d1ec4aee22c3654f4e4fd), remapping is [here](https://github.com/mottosso/Qt.py/pull/112/files#diff-cd1d7c0a982bce680146d5b248151256R72).

This makes monkey-patching classes a non-starter.

mottosso:
Here's an alternative, to subclass rather than monkey-patch.

The problem in this case is that the function `QTreeView.header()` isn't returning *our* subclass, but rather the original class; as the call is to C++ and C++ knows nothing of what we've done externally.

A fact easily proven via:

```python
id(qheaderview) != id(QtWidgets.QHeaderView)
```

- [Stack trace](https://travis-ci.org/mottosso/Qt.py/jobs/155582820#L2715)

This makes subclassing for filling in missing functionality in older bindings a non-starter as well.

Due to these findings, and unless anyone has an alternative implementation to share, this cannot be merged. :(

mottosso:
Here's a third alternative from qtpy; also monkey-patching the class.

mottosso:
Important lessons learned here are:

1. We cannot replace members of `QtWidgets` et. al. with our own subclasses due to classes returned via internal functionality, such as `QTreeView.header()`, does not know nor care about this.
2. We cannot monkey-patch a class because (1) it modifies the original binding and/or (2) it *sometimes* doesn't work; as seen here with Python 3.5-only ([more](https://github.com/spyder-ide/qtpy/issues/70)).

Closing this, and will refer back here whenever any of these two solutions crop up in pull-requests.

mottosso:
Due to the PR coming from a master branch, rather than a feature branch, I've force-pushed it back to it's original state and stored these changes for future reference in an alternative branch.

- https://github.com/abstractfactory/Qt.py/tree/setSectionResizeMode

Issue #115, opened 2016-08-29
QSortFilterProxyModel missing


hdd:
As much as for PySide QSortFilterProxyModel should be remapped to QtCore for PyQt4

Issue #105, opened 2016-08-12
Test availability paridy with PySide2


mottosso:
### Goal

Check that all bindings provide identical members as PySide2; the reference binding.

### Motivation

It has happened that one binding was missing access to certain classes that should have been available due to a typo or forgetfulness. A "one test to rule them all" for availability should iron these out.

### Implementation

I'd imagine running a simple `dir(PySide2.QtWidgets)` and checking that against what's available elsewhere. It would be nice to also include argument signatures for these. Possibly that we can achieve this via the Python `inspect` module.

The number one goal is for each binding to provide *the same* members and argument signatures (differences break cross-binding compatibility). If we can also reach a state where they all provide *everything* that PySide2 has in an easy manner, then that would be great too, but it isn't as important and is unlikely to ever happen. Simply because some of the features provided are ad-hoc (due to trying to convert C++ methodologies to Python).

ethanhs:
There are several issues with PySide2's bindings of QtQuick and QtOpengl, but I believe the QtWidgets should be complete or nearly there.

mottosso:
Thanks @ethanhs, good to see you here. Have you by any chance changed your username recently? IronManMark20? :)

ethanhs:
@mottosso I think its been a few months, but yes. :) At some point I want to add a UI to a binding generator I've been playing around with, and came across this library. I'm surprised I didn't notice it earlier, this solves a lot of headaches I've been solving myself.

fredrikaverpil:
> There are several issues with PySide2's bindings of QtQuick and QtOpengl, but I believe the QtWidgets should be complete or nearly there.

That is correct. For completeness, I've added the changelog from https://wiki.qt.io/PySide2 below:

> 4 August 2016
> - Additional Pyside examples under review
> - Fixed warnings coming from Shiboken
> 
> 21 July 2016
> - main Pyside 2 example port done
> - additional examples to fill gaps are being ported (as per prio list)
> - CI patches running, but still gaps (should be done by next week)
> - about 86% of auto test working (failing count of test 80+ auto test)
> - Project Test Status
> 
> 14 July 2016
> - Pyside 2 examples ported
> - **OpenGL & SVG not working**
> - **QtQuick 2 is in strange situation (QtQml depends on QtQuick)**
> - Python 3 realed Unicode handling not working with Qt
> - **QMessageBox hangs**
> - no documentation for any example
> - lots of warnings when building wrapper
[...]

PR #116, opened 2016-08-29
add missing remap

hdd:
this should be fixing #115 
please have a look and let me know.


fredrikaverpil:
@hdd that's excellent.

I see this was already remapped:

    PySide.QtCore.QSortFilterProxyModel = PySide.QtGui.QSortFilterProxyModel

Do we need a test for this?

mottosso:
Thanks @hdd.

> Do we need a test for these? (I'm guessing no)

This is how I envision testing for these members.

- https://github.com/mottosso/Qt.py/issues/105

You're welcome to have a look at how to implement something like this!

hdd:
Cheers, I'll have a look as soon as I have time.
(always better have tests ;) )

2016-08-29 11:27 GMT+02:00 Marcus Ottosson <notifications@github.com>:

PR #117, opened 2016-08-29
Add map for QStringListModel

fredrikaverpil:
I found that this was missing from `_pyqt4`:

    PyQt4.QtCore.QStringListModel = PyQt4.QtGui.QStringListModel

This is already in `_pyside`:

    PySide.QtCore.QStringListModel = PySide.QtGui.QStringListModel


mottosso:
Thanks @fredrikaverpil, referencing #105 for completeness on a solution to all of these at once.

fredrikaverpil:
> referencing #105 for completeness on a solution to all of these at once.

Yes, agreed. We should do that!

PR #118, opened 2016-08-29
Add PyPi badge

fredrikaverpil:
I thought this was nice:

[PyPI version](https://pypi.python.org/pypi/Qt.py)

:)


mottosso:
Very nice indeed. :)

PR #119, opened 2016-08-29
Add examples

fredrikaverpil:
Docker for Mac is acting up here and I can't perform tests locally... so I'm going to have to rely on Travis. Bare (bear?) with me for a while...


fredrikaverpil:
Hm, not sure why you can't see the tests in Travis. This is what I see locally:

```
docker run --rm -v $(pwd):/Qt.py mottosso/qt.py27
Doctest: test_caveats.test_11_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_caveats.test_12_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_caveats.test_13_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_examples.test_1_base_instance_as_argument_using_getattr_setattr ... ok
Tests require all bindings to be installed ... ok
Setting QT_PREFERRED_BINDING to PyQt4 properly forces the binding ... ok
Setting QT_PREFERRED_BINDING to PyQt5 properly forces the binding ... ok
Setting QT_PREFERRED_BINDING to PySide properly forces the binding ... ok
Setting QT_PREFERRED_BINDING to PySide2 properly forces the binding ... ok
Setting QT_PREFERRED_BINDING to more than one binding excludes others ... ok
Preferred binding PyQt4 should have sip version 2 ... ok
Qt.py may be bundled along with another library/project ... ok
Raise ImportError if sip API v1 was already set (Python 2.x only) ... ok

Ran 25 tests in 6.910s

OK
```

So, what I did was:

* `build_caveats_tests.py` -> `build_md_tests.py` (md = Markdown)
*  `caveats.py` -> `mdparse.py`

Then I added an `/examples` folder with a `load_ui.md` example in it which gets parsed, just like CAVEATS.md. I extended `build_md_tests.py` to cover any example in `/examples`.

What do you think about this approach?

I also noticed the `sleep 3` command in the Dockerfiles wasn't enough for my Mac laptop. I increased that to 5 seconds.

mottosso:
Hey,

First off, great idea to make examples runnable. I'm not entirely keen on the idea of examples having the `>>>` prefixes, as it makes it difficult to copy/paste to run them..

Would it be possible to make them simple Python files, without the need for parsing?

fredrikaverpil:
Yes, I agree. I think that would make it easier for anyone to contribute as well. I'll try to come up with something which can be included in the tests.

fredrikaverpil:
Hm...

```
docker run --rm -v $(pwd):/Qt.py mottosso/qt.py27
Doctest: test_caveats.test_11_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_caveats.test_12_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_caveats.test_13_qtwidgets_qheaderview_setresizemode ... ok
Segmentation fault
```

mottosso:
This will take you into the image so you can test interactively.

```bash
$ docker run -ti --rm --entrypoint bash -v $(pwd):/Qt.py mottosso/qt.py27
$ cd /Qt.py
$ python build_examples_tests.py
$ python some_test.py..
```

fredrikaverpil:
What do you think so far?

fredrikaverpil:
Hm. I miss having an .md file with explanations.

examples/load_ui_base_instance.py
```diff
@@ -0,0 +1,40 @@
+import sys
+import os
+
+os.environ["QT_PREFERRED_BINDING"] = "PySide"
```
mottosso: How come you prefer PySide here? Shouldn't this work on any binding?

examples/load_ui_base_instance.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_load_ui_setup_ui_wrapper():
+    """Example: load_ui with setup_ui wrapper
+    """
+    app = QtWidgets.QApplication(sys.argv)
+    window = MainWindow()
+
+    # Tests
+    assert isinstance(window.__class__, type(QtWidgets.QWidget))
```
mottosso: `type()` returns the superclass, in this case it returns `<class 'PyQt5.QtCore.pyqtWrapperType'>` under PyQt5. It's cheating a bit, as everything is an instance of this super-superclass.

mottosso: In most cases `type(obj)` is identical to `obj.__class__`.

mottosso:
> Hm. I miss having an .md file with explanations.

Yes, an overview would be helpful.

I see two options.

1. Make your module read-friendly.
 ```python
"""Load a .ui file

This example shows this and that.
    _____________________
   | An example program  |
   |  _________________  |
   | |                 | |
   | |    A Button     | |
   | |_________________| |
   |_____________________|

Usage:
    Ensure any binding is available on your PYTHONPATH.
    >>> python this_example.py
"""

# Code starts here..
```
2. Produce a directory, with a `README.md`. The README would become the landing page for the directory, similar to how it looks on the main project page.
 ```bash

examples/
  example1/
    example1.py
    README.md
```

examples/load_ui_base_instance.py
fredrikaverpil: I get a weird thing in Python 2.7 if I don't specify this. Don't really understand why. I can remove it and you'll see in my next commit.

mottosso: Ok. If unspecified, it will try and use PySide2. Maybe there's some issue there.

examples/load_ui_base_instance.py
fredrikaverpil: True, I can change that into

    assert isinstance(window.__class__, QtWidgets.QWidget.__class__)

fredrikaverpil:
I was thinking along the lines of option 2 myself; the README.md.

I'll create it and we'll see how it feels.

examples/load_ui_base_instance.py
mottosso: No, no, no. It needs to be `isinstance(window, QtWidgets,QWidget)`.

Classes aren't instances of anything, that can never fail. To compare classes, you would have to use `issubclass`, but that isn't relevant here when you are dealing with instances.

fredrikaverpil:
Hm... I'm going to need to install pysideuic and uic in the Docker images. Not sure if pyside2uic even exists though.

mottosso:
How come?

fredrikaverpil:
I wanted to include the example from [this comment](https://github.com/mottosso/Qt.py/pull/81#issuecomment-239156239), which uses PyQt4.uic or pysideuic.

If we want to test it, we need to have pysideuic.

mottosso:
Better pre-generate those and include the resulting files in the example instead.

fredrikaverpil:
See how the test seems to have abruptly been aborted [here](https://travis-ci.org/mottosso/Qt.py/jobs/155996169#L2705)?

When I run the test locally:

```bash
$ docker run -ti --rm --entrypoint bash -v $(pwd):/Qt.py mottosso/qt.py27
$ cd /Qt.py
$ python build_examples_tests.py 
$ python test_load_ui_base_instance.py 
Trying pysideroot@21c5bc986a14:/Qt.py# 
```

fredrikaverpil:
> Better pre-generate those and include the resulting files in the example instead.

PyQt4.uic is included in PyQt4... and I can install pysideuic and pyside2uic for Python 2.7 (`pyside-tools`, `pyside2-tools`) but I can't find anything for Python 3.5.

What do you mean when you say "pre-generate"?

mottosso:
> What do you mean when you say "pre-generate"?

I mean convert them to `.py` files, and include them in the example folder. Rather than compile in-memory.

fredrikaverpil:
> I mean convert them to .py files, and include them in the example folder. Rather than compile in-memory.

In this case, the point of that example is to show how you can avoid to pre-generate the UI files.

I'd say we either skip testing when such dependencies are needed, or we use a preferred binding which in this case could make the test pass using PyQt4 since that already comes with uic.

fredrikaverpil:
Let's recap: I've now set preferred bindings for these examples.

The main reason is the tests phase bug out in the Python 2.7 container if I don't do this.

The second reason is that for the `load_ui_base_instance_pysideuic.py` example, `uic` or `pysideuic` is required. If setting PyQt4 as the preferred binding, it'll just pass the  tests and work.

I could implement something like a detection for an example to not be included in the tests, although I think this is fine.

I've now included `apt-get install pyside-tools pyside2-tools` to `Dockerfile-py2.7`. I'll now remove that since we won't really need it.

examples/load_ui_base_instance.py
fredrikaverpil: Oh holy crap I see what I did now. I must've been really distracted. I'll fix.

mottosso:
Aw man, you gotta start rebasing instead of merging. This is sparta.. I mean madness. :)

fredrikaverpil:
Wtf... I did a rebase on origin/master but is this really supposed to happen (all those commits once again)?

fredrikaverpil:
Oh well, anyways ... I see you're now explicitly defining which files to nosetest using the `--exe` option. This makes my examples not get included in the tests.

fredrikaverpil:
If we specify the `--exe` option but don't specify individual files, I think anything starting with "test" (`test*.py`) should be included automatically.

fredrikaverpil:
> If we specify the --exe option but don't specify individual files, I think anything starting with "test" (test*.py) should be included automatically.

Scratch that, I was wrong. It'll include any *.py file it seems.

mottosso:
It's not supposed to happen.

Ideally, it would be (1) the current state of this repo + (2) your new commits.

At the moment, it looks like (1) the current state of the repo, (2) some of your commits, (3) merged and (4) more of your commits on-top of this merge. = Madness

> If we specify the --exe option but don't specify individual files, I think anything starting with "test" (test*.py) should be included automatically.

Yes, it was like that, but I was having trouble. It should be that way, but the thing is, it will open every Python file, and look for functions with "test" in its name. Running the function `build_tests()` from the membership module, implicitly re-creating (and overwriting) the comparison JSON file which breaks the test.

We should change it back, but we'll also need to change those function names.

mottosso:
Do you think this is possible? `--exe test*.py`?

fredrikaverpil:
Hehe, we thought the same thing. Yes, that worked!

mottosso:
About the `examples/` directory, would it be worth putting each example in its own folder, with its own README? Or were you planning on having all examples lie flat in this directory?

The problem I'm seeing is when there are two or more examples, knowing which files belong to which example can be tricky. But I also don't want to complicate things.

Alternatively, would it be worth having a 1 file per example restriction? In this case, encoding the `.ui` file within the `.py` file.

fredrikaverpil:
> Or were you planning on having all examples lie flat in this directory?

Initially, yes. I thought it would be easy to change this if it would turn out that we'd add a lot of examples.

> Alternatively, would it be worth having a 1 file per example restriction? In this case, encoding the .ui file within the .py file.

This I think would make examples complex and both hard to follow as well as hard to contribute to. So, I'd like to avoid that.

> About the examples/ directory, would it be worth putting each example in its own folder, with its own README?

Sure, we could do that.

fredrikaverpil:
@mottosso what do you think?

examples/load_ui/load_ui_1.py
```diff
@@ -0,0 +1,42 @@
+import sys
+import os
+
+# Set preferred binding
+os.environ["QT_PREFERRED_BINDING"] = "PySide:PyQt4"
```
mottosso: Could you make this `os.pathsep.join(["PySide", "PyQt4"])` so that it also runs on Windows?

(I would also be up for discussing an alternative syntax for this, having now seen it in the real-world, e.g. `"PySide,PyQt4"`)

examples/load_ui/load_ui_1.py
```diff
@@ -0,0 +1,42 @@
...
+
+from Qt import QtWidgets, load_ui
+
+
+def setup_ui(uifile, base_instance=None):
```
mottosso: Perhaps a small docstring, explaining the purpose of each function?

examples/load_ui/load_ui_2.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+
+# Set preferred binding
+os.environ["QT_PREFERRED_BINDING"] = "PyQt4"
+
+from Qt import QtWidgets
+from Qt import __binding__
+
+
+def loadUiType(uiFile):
```
mottosso: Camel-case function names hurts my PEP08 eyes. :(

examples/load_ui/load_ui_2.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Pyside equiv for the loadUiType function in qt.
+    """
+
+    import pysideuic
+    import xml.etree.ElementTree as ElementTree
+    from cStringIO import StringIO
+
+    parsed = ElementTree.parse(uiFile)
+    widget_class = parsed.find('widget').get('class')
```
mottosso: Mixing `'` with `"` hurts my eyes. :(

examples/load_ui/load_ui_2.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # Fetch the base_class and form class based on their type in
+        # the xml from designer
+        form_class = frame['Ui_%s' % form_class]
+        base_class = eval('QtWidgets.%s' % widget_class)
+    return form_class, base_class
+
+
+def loadUi(uifile, baseinstance=None, workingDirectory=None):
+    """
```
mottosso: The Google docstring syntax is:

```python
def my_function():
  """Short, one-line description

   Long description

   """

   # Code here, notice the empty line inbetween docstring and code.
```

Could we go with that, for consistency?

- http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
```

**Comment by mottosso** [comment]
I like it!

**Comment by fredrikaverpil** [comment]
Great points, I agree on all of them. My eyes hurt as well now ;)

I'll address them all, probably tomorrow.

**Comment by fredrikaverpil** [comment]
Fixed:

* Using `os.pathsep.join`
* Docstrings (google style)
* snake_case (no more camelCase)
* Only using `"` in docstrings and `'` for everything else

I'm also copying non-python files into the working directory, so that the ui file can be defined without the relative filepath into `/examples/etc.../file.ui`. But I have another suggestion which I'll write about in a minute...

> (I would also be up for discussing an alternative syntax for this, having now seen it in the real-world, e.g. "PySide,PyQt4")

Yes, I think a comma is more suitable. I always thought `:` or `;` were indicating a filesystem path (which in this case we don't want to imply). The comma is also more suitable to indicate that there's a specific order of things. The first item in the comma separated list is first used etc...

**Comment by fredrikaverpil** [comment]
So, I'm copying files from `/examples/*/*.*` into the working directory. This is not so nice, as we might end up with filename clashes.

If I put an `__init__.py` inside of the `examples` folder and also inside of each example folder, I can add use `--exe examples` and nose will parse all example files. All I have to do is to prepend python scripts to be tested with `test`.

I'll push an example of this and we can discuss.

**Comment by fredrikaverpil** [comment]
Edited my previous post so much if you already read it I'm afraid I'd like you to re-read it. Sorry - made more sense to consolidate, I thought. 😢

**Comment by fredrikaverpil** [comment]
In the previous commit:

* No need to build examples, `build_examples.py` removed.
* Examples are being executed where they already reside, if they include a function prefixed with `test` since the `--exe` option now takes `examples/*/*.py` as value.
* I added `__init__.py` to `/examples` and `examples/load_ui` which makes these examples importable.

```python
from examples.load_ui import baseinstance1

widget = baseinstance1.setup_ui('qwidget.ui', self)
```

@mottosso what do you think about all this?

mottosso:
> Yes, I think a comma is more suitable.

I know you did, and having worked with it for a while now, I think your gut was right.

> If I put an __init__.py inside of the examples folder and also inside of each example folder,

This sounds good to me.

For the record, the `--exe` option isn't related to the next argument, it's a standalone flag meaning "run executable files, as well as non-executable files".

> I'm not saying we are to bundle that with Qt.py via PyPi but we could if we decide to do so in the future. 

We still could not, I think, as `Qt.py` and the `examples/` directory would end up directly in `site-packages/`, making `examples` a globally available Python package; e.g. `import examples`.

It doesn't hurt having a package in this way in the project however, as we explicitly specify that the file `Qt.py` is the only file to be included in the PyPI package.

> So instead of making Qt.py grow, /example can grow instead.

Conceptually, I like the sound of that, but am having trouble seeing how growth of any kind is helpful for such simple and de-coupled things as examples. I'll keep an open mind though.

> One might argue that with the importable examples, we should define our tests in test.py instead.

Do you mean in the global `github.com/Qt.py/tests.py`, or in a local `test.py` per example?

fredrikaverpil:
> Do you mean in the global github.com/Qt.py/tests.py, or in a local test.py per example?

I meant that one might argue to collect all tests into github.com/Qt.py/tests.py but in this case I think that, for completion purposes, it's nicer to define test in each example. Also, defining the tests in the global tests.py would decouple the test and the example a bit too much, I think. So I like this current approach that is in this PR.

Dockerfile-py2.7
```diff
@@ -25,11 +25,14 @@ ENV DISPLAY :99
 
 WORKDIR /workspace/Qt.py
 ENTRYPOINT cp -r /Qt.py /workspace && \
-    python build_caveats_tests.py && \
+    python build_caveats.py && \
+    python build_membership.py && \
 	Xvfb :99 -screen 0 1024x768x16 2>/dev/null & \
-	sleep 3 && \
+	sleep 10 && \
```
mottosso: Woah, seems like this one has no end.

Let's find a better way of handling this. There must be some way of waiting for a signal/sign of sorts, and continuing once that happens?

fredrikaverpil: Yes, I don't know if it's a memleak or what is going on... but I need to rebuild my containers at least once a day. And I have to restart XQuartz a few times too. So I'm not sure what's causing that.

But so 3 seconds isn't always enough... and that's why I increased it to something I don't have to go back and tweak with +1 second...

mottosso: >  XQuartz

You shouldn't need anything like that on your machine; xvfb is completely isolated within the container.

fredrikaverpil: I'm giving this a go now...

    xdpyinfo -display :99 >/dev/null 2>&1 && echo "In use" || echo "Free"

fredrikaverpil: Ah, this was a better/simpler approach:

    ps auxx | grep Xvfb

mottosso: I'm not following, but glad to hear you may have found a way.

fredrikaverpil: I'm waiting until I find the process `Xvfb` to be running. Works here on Docker for Mac.

fredrikaverpil: > > XQuartz
> You shouldn't need anything like that on your machine; xvfb is completely isolated within the container.

Ah!

fredrikaverpil:
So, now there's a loop waiting for the `Xvfb` process. Let me know what you think about this and the rest of this PR.

I think this should be ready for a merge.

Dockerfile-py2.7
```diff
@@ -25,11 +25,14 @@ ENV DISPLAY :99
 
 WORKDIR /workspace/Qt.py
 ENTRYPOINT cp -r /Qt.py /workspace && \
-    python build_caveats_tests.py && \
+    python build_caveats.py && \
+    python build_membership.py && \
 	Xvfb :99 -screen 0 1024x768x16 2>/dev/null & \
-	sleep 3 && \
+    while ! ps aux | grep -q '[0]:00 Xvfb :99 -screen 0 1024x768x16'; do echo "Waiting for Xvfb..."; sleep 1; done && \
```
mottosso: Neat!

Without the `2>/dev/null`, are we getting output from xvfb at this point?

fredrikaverpil: The new command just checks whether `grep` finds this string:

    0:00 Xvfb :99 -screen 0 1024x768x16

The original command which launches Xvfb is untouched. So the functionality should be unchanged.

mottosso: Aah, right, missed that, sorry.

That's great!

fredrikaverpil: I could make the search string include that last pipe bit. I didn't do it just to shorten down the line length.

mottosso: Is it possible to break this line?

```bash
    ...
	Xvfb :99 -screen 0 1024x768x16 2>/dev/null & \
    while ! ps aux | grep -q '[0]:00 Xvfb :99 -screen 0 1024x768x16'; \
        do echo "Waiting for Xvfb..."; sleep 1; done && \
    nosetests \
    ...
```

fredrikaverpil: > I could make the search string include that last pipe bit. I didn't do it just to shorten down the line length.

On second thought, I can't. `ps` doesn't show piping:

    root         6  0.7  1.5 207336 31256 ?        Sl   10:05   0:00 Xvfb :99 -screen 0 1024x768x16

So this is fine.

fredrikaverpil: > Is it possible to break this line?

Yes, done!

mottosso:
LGTM!

fredrikaverpil:
Awesome, I'll merge this then!

PR #121, opened 2016-08-30
Update documentation for load_ui

mottosso:
Some clarification, but also the idea to put argument signatures in the documentation, similar to how it would appear on ReadTheDocs and Sphinx-generated documentation in general.

Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,13 +79,13 @@ app.exec_()
 
 All members of `Qt` stem directly from those available via PySide2, along with these additional members.
 
-| Attribute               | Type   | Value
-|:------------------------|:-------|:------------
-| `__binding__`           | `str`  | A string reference to binding currently in use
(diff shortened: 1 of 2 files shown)


fredrikaverpil:
I didn't check the return type myself, but it looks right. LGTM!

mottosso:
About this last commit, I've been thinking about how we can run a test with every binding.

We're able to pinpoint which binding to test for, which is necessary with for example the `sip` test for PyQt4. But we'll also need to test one thing across every binding, like in this commit.

I can think of two approaches.

1. Make `tests.py` into a template, similar to `build_caveats.py`, that can be used to generate the additional tests. Then we can use something along the lines of `test_{binding}_something()` in the original module.
2. Run nose multiple times, each time setting QT_PREFERRED_BINDING. This would be the most obvious, but would also nullify the context managers to run a specific binding.

Issue #122, opened 2016-08-30
Enable running one test on each binding


mottosso:
### Goal

Simplify maintenance by running one test multiple times, exposing each binding at a time.

### Motivation

At the moment, each test is run once. A test can specify with which binding it is to get run. But as Qt.py is meant to enable the *same* code to run, regardless of binding, it would also be helpful to enable running a test with each binding.

mottosso:
This is done!

**To run a single test on each binding**

This is default, and a tests are written as a plain-old-function.

```python
def test_me():
  print("Running 4 times, now on: %s" % os.getenv("QT_PREFERRED_BINDING"))
```

**Run a single test on a single binding**

Use a conditional to "mask off" a test from other bindings. The test will be invisible for when the time comes to run tests under a binding you didn't specify.

```python
if binding("PyQt4"):
  def test_me():
    print("Running exclusively on %s" % os.getenv("QT_PREFERRED_BINDING"))
```

- [Example](https://github.com/mottosso/Qt.py/blob/86051d7141a0045bcbb6c98c0263a3b402ceec64/tests.py#L162)

PR #123, opened 2016-08-30
Implement #122

mottosso:
#122

Without access to Docker at the moment, will trust Travis to run tests for me.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -32,7 +32,7 @@ if __name__ == "__main__":
         "--exe",
     ]
 
-    argv.extend(sys.argv[1:])
+    # argv.extend(sys.argv[1:])
 
     # Running each test independently via subprocess
     # enables tests to filter out from tests.py before
@@ -48,5 +48,4 @@ if __name__ == "__main__":
         subprocess.call(argv)
 
     with binding("PySide2"):
-        argv.append("--with-doctest")  # Run doctests once
         subprocess.call(argv)
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,0 +1,52 @@
+import os
+import sys
+import contextlib
+import subprocess
+
+
+@contextlib.contextmanager
+def binding(binding):
+    """Prepare an environment for a specific binding"""
+
+    sys.stderr.write("""\
+#
+# Running tests with %s..
+#
+""" % binding)
+
+    os.environ["QT_PREFERRED_BINDING"] = binding
+
+    try:
+        yield
+    except:
+        pass
+
+    os.environ.pop("QT_PREFERRED_BINDING")
+
+
+if __name__ == "__main__":
+    argv = [
+        "nosetests",
+        "--verbose",
+        "--with-process-isolation",
+        "--exe",
+    ]
+
+    argv.extend(sys.argv[1:])
+
+    # Running each test independently via subprocess
+    # enables tests to filter out from tests.py before
+    # being split into individual processes via the
+    # --with-process-isolation feature of nose.
+    with binding("PyQt4"):
+        subprocess.call(argv)
+
+    with binding("PySide"):
+        subprocess.call(argv)
+
+    with binding("PyQt5"):
+        subprocess.call(argv)
+
+    with binding("PySide2"):
+        argv.append("--with-doctest")  # Run doctests once
+        subprocess.call(argv)
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,14 +13,6 @@ services:
 install:
   - docker build -f Dockerfile-py${TRAVIS_PYTHON_VERSION} -t mottosso/qt.py .
 
-# Enable virtual framebuffer, this way we can instantiate
-# a QApplication and create widgets.
-# https://docs.travis-ci.com/user/gui-and-headless-browsers/
-before_script:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-  - sleep 3 # give xvfb some time to start
-
 script:
   - docker run --rm -v $(pwd):/Qt.py mottosso/qt.py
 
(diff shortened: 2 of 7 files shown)


mottosso:
There we go.

Let me know how this feels to you.

mottosso:
There are 2 smaller problems in this current version.

1. Tests from CAVEATS.md are triggered, but not really run, once per binding, even though it's irrelevant as they pick their own binding at run-time.
2. Doctests are run once, but in a [not-so-intuitive way](https://github.com/mottosso/Qt.py/pull/123/files#diff-0e150adfe1f55c100b4edc3a4b107418R51) as PySide2 tests.

They won't prevent tests from running correctly, but does make the result presentation less clear.

run_tests.py
```diff
@@ -0,0 +1,52 @@
...
+    ]
+
+    argv.extend(sys.argv[1:])
+
+    # Running each test independently via subprocess
+    # enables tests to filter out from tests.py before
+    # being split into individual processes via the
+    # --with-process-isolation feature of nose.
+    with binding("PyQt4"):
+        subprocess.call(argv)
```
mottosso: This could have been:

```python
subprocess.call(argv, env={"QT_PREFERRED_BINDING": "PyQt4"})
subprocess.call(argv, env={"QT_PREFERRED_BINDING": "PySide"})
subprocess.call(argv, env={"QT_PREFERRED_BINDING": "PySide2"})
subprocess.call(argv, env={"QT_PREFERRED_BINDING": "PyQt5"})
```

But I personally think the current look is more readable.

fredrikaverpil: I think it's important we get this right. Are there any cons with the current approach?

You could always use the subprocess way and write a comment to make it more readable.

fredrikaverpil:
This is excellent. LGTM!

fredrikaverpil:
> Doctests are run once, but in a not-so-intuitive way as PySide2 tests.

Hm. They seem to be run for each binding? (see below)
How come the doctests appear for each binding but only for PySide2 it appears properly?

```
#
# Running tests with PyQt5..
#
Test qtgui_qabstractitemmodel_createindex ... ok
Test qtgui_qabstractitemmodel_createindex ... ok
Test qtcore_qitemselection ... ok
Test qtcore_qitemselection ... ok
Test qtcore_qitemselection ... ok
Test qtcore_qitemselection ... ok
Test qtcore_slot ... ok
Test qtcore_slot ... ok
Test qtwidgets_qaction_triggered ... ok
Test qtwidgets_qaction_triggered ... ok
Test qtwidgets_qheaderview_setresizemode ... ok
Test qtwidgets_qheaderview_setresizemode ... ok
Test qtwidgets_qheaderview_setresizemode ... ok
...
...
...

OK
#
# Running tests with PySide2..
#
Doctest: test_caveats.test_11_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_caveats.test_12_qtwidgets_qheaderview_setresizemode ... ok
Doctest: test_caveats.test_13_qtwidgets_qheaderview_setresizemode ... ok
```

mottosso:
> How do you mean "but not really run"?

Yeah I should have clarified.

If you look [in travis](https://travis-ci.org/mottosso/Qt.py/jobs/156365025#L3015), you'll see things like this:

```bash
Test qtcore_qitemselection ... ok
```

Whereas in the last test with docstrings, the [same test](Test qtcore_qitemselection ... ok) appears but looks like this.

```bash
```

The first time around, it just runs the empty function created to hold the docstring. It cannot fail, and is a pretty useless test as-is.

> You mean they only appear under PySide2 tests?

I mean visually, in [`run_tests.py`](https://github.com/mottosso/Qt.py/pull/123/files#diff-0e150adfe1f55c100b4edc3a4b107418R51), it looks like doctests and PySide2 are related, when they are not. Doctests pick their own binding at run-time.

fredrikaverpil:
Hm, I see. How about running the doctests separately?

Right now you run

    python -u run_tests.py test*.py

We could remove the caveats/doctest from it and execute your wrapper and then the doctest, perhaps?

    python -u run_tests.py test*.py && \
    nosetests \
        --verbose \
        --with-process-isolation \
        --with-doctest \
        --exe \
            caveats/test_caveats.py

mottosso:
> ...although that feels like starting a bowl of spaghetti...

I can also spot the potential for that, but there's something to this approach that I think could work. Will continue to investigate.

mottosso:
Ok, the Dockerfile(s) got more complex; but the end result is clean and nothing runs twice or unnecessarily.

- https://travis-ci.org/mottosso/Qt.py/jobs/156615207

Part of me thinks it's, like you said, spaghetti. Another part of me feels like the Dockerfile is a static, unchanging portion of Qt.py that doesn't necessarily need changing.

Let me know what you think.

run_tests.py
mottosso: Sorry missed this comment.

Cons, not that I can tell. There are some organisational cons; as we are now doing in Python what we could have done in the docker ENTRYPOINT, and in the ENTRYPOINT we are doing things that could have been done here.

I sense that at some point we'll either transition into doing it all there, or all in Python. But personally I feel it's too early to tell and would be happy to leave it as-is until we can point at concrete problems and manage them then. Unless you see any issues with it as-is?

fredrikaverpil:
> the end result is clean and nothing runs twice or unnecessarily.

Nice. I like it. Can you get the examples testing in there as well?


> Cons, not that I can tell. There are some organisational cons; as we are now doing in Python what we could have done in the docker ENTRYPOINT, and in the ENTRYPOINT we are doing things that could have been done here.
> 
> I sense that at some point we'll either transition into doing it all there, or all in Python. But personally I feel it's too early to tell and would be happy to leave it as-is until we can point at concrete problems and manage them then. Unless you see any issues with it as-is?

Agreed. I have no issues with this. Looks good.

mottosso:
> Can you get the examples testing in there as well?

Oh! I must have assumed they got included. Sorry about that, I'll have a look.

fredrikaverpil:
Whenever this is done, merge away :) I think this is great.

Issue #111, opened 2016-08-26
Ensure Qt.py doesn't break original binding


mottosso:
### Goal

Find a way to ensure, either via tests or rethinking the implementation, that `import Qt` doesn't break e.g. `import PySide` or any of it's members.

### Motivation

This is about future proofing more than anything as no problem has yet appeared.

At the moment, because Qt.py replaces itself with its original binding, it's possible for it to secretly break this binding. Primarily by replacing one of its members with one that is broken.

### Implementation

We can solve it by having tests confirm that nothing has broken, or potentially by *not* replacing oneself with the original binding and instead recreate the member hierarchy such that members of `import Qt` are unique. Doing so would however require more than just `Qt.QtWidgets = PySide2.QtWidgets`, due to the fact that we also replace members *within* these submodules.

So a thorough version would also include sub-submodule, which can get lengthy and potentially *error prone*. Error prone because these members are not your regular Python modules, but rather coming from C++ via black magic.

Alternatives or implementation ideas are welcome!

mottosso:
This has been implemented in [0.4.0](https://github.com/mottosso/Qt.py/releases/tag/0.4.0)

PR #130, opened 2016-09-05
Implement #111

mottosso:
Here's an idea for #111

Rather then post-checking that a binding has not been broken, we can pre-check by making sure nothing happening in Qt.py overwrites or alters any existing member.

With this, Qt.py can have no adverse effects on an original binding, other than affect other third-parties from adding members to where we also add them. Should we bother with that? Can we?

### Theory

Notice the `add` and `remap` functions.

These replace the current layout of remapping and adding.

```python
# Before
PyQt4.QtWidgets = PyQt4.QtGui

# After
remap(PyQt4, "QtWidgets", PyQt4.QtGui)
```

By using a function as opposed to directly assigning, we lose out on readability, but gain security.

The `safe=True` argument is meant to facilitate times where we *do* overwrite, but ensure to include a full array of tests to ensure the original member has not been broken.

Maybe we should also check whether we monkey-patch a class or a module, as classes has [proven problematic](https://github.com/mottosso/Qt.py/pull/112)?

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -2,7 +2,17 @@
 
 This module replaces itself with the most desirable binding.
 
-Resolution order:
+Project goals:
+    Qt.py was born in the film and visual effects industry to address
+    the growing need for the development of software capable of running
+    with more than one flavour of the Qt bindings for Python - PySide,
+    PySide2, PyQt4 and PyQt5.
+
+    1. Build for one, run with all
+    2. Explicit is better than implicit
+    3. Support co-existence
+
+Default resolution order:
     - PySide2
     - PyQt5
     - PySide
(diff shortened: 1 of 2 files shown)


mottosso:
> Maybe we should also check whether we monkey-patch a class or a module, as classes has proven problematic?

Implemented.

BigRoy:
I'm getting the following error when import Qt.py:

```python
  File "<nofile>", line 1, in <module>
  File "P:\pipeline\dev\git\Qt.py\Qt.py", line 265, in <module>
    init()
  File "P:\pipeline\dev\git\Qt.py\Qt.py", line 245, in init
    binding = binding()
  File "P:\pipeline\dev\git\Qt.py\Qt.py", line 92, in pyqt5
    add(PyQt5, "__qt_version__", QtCore.QT_VERSION_STR)
  File "P:\pipeline\dev\git\Qt.py\Qt.py", line 78, in add
    remap(object, name, value)
  File "P:\pipeline\dev\git\Qt.py\Qt.py", line 64, in remap
    "%s.%s" % (object.__name__, name))
AttributeError: Cannot override existing name: PyQt5.__qt_version__
```

It looks related to this PR. Any idea what might be going wrong?

mottosso:
Hm, could there be a `reload()` of Qt.py somewhere?

mottosso:
Could you post a complete reproducible? I'm not able to reproduce this error. :(

mottosso:
Ok, I see what's going on.

You're using this version of PyQt5.

- https://github.com/pyqt/python-qt5/blob/master/PyQt5/__init__.py

And that version just so happens to define `__qt__version__` in a similar fashion to how Qt.py does it. (Surprise surprise, that one's also from me).

Could you make an issue for this? Let's make it official.

BigRoy:
> Could you make an issue for this? Let's make it official.

Thanks for figuring that out. Was about to go on a hunt not knowing exactly what to look for. :)
Will set up an issue right away.

Issue #131, opened 2016-09-07
Qt --convert my.ui


mottosso:
### Goal

In response to #99 and #129, provide a safe conversion from `.ui` files compiled with PySide2 to an equivalent Qt.py-compatible module.

### Implementation

From a plain-text Python module, produce a module with full cross-binding compatibility with Qt.py using plain old search-and-replace.

```bash
$ pyside2-uic.exe my_ui.ui -o my_ui.py
$ python -m Qt --convert my_ui.py  # save original as `original_my_ui.py`
$ cat my_ui.py
```

mottosso:
Implemented in [0.5.0](https://github.com/mottosso/Qt.py/releases/tag/0.5.0).

fredrikaverpil:
I guess that with PR https://github.com/mottosso/Qt.py/pull/142 merged it should now also be possible to implement `--compile`.

Issue #129, opened 2016-09-03
QtWidgets.QApplication.translate syntax error using PySide


elveatles:
The translate method is different between Qt4 and Qt5. 
You can try this Python code:

```
import sys

from Qt import QtWidgets

class MainWindow(QtWidgets.QWidget):
    def __init__(self, parent=None):
        super(MainWindow, self).__init__(parent)
        title = QtWidgets.QApplication.translate("MainWindow", "MainWindow", None, -1)
        
def main():
    app = QtWidgets.QApplication(sys.argv)
    mainWindow = MainWindow()
    mainWindow.show()
    app.exec_()

if __name__ == '__main__':
    main()
```

If I have PySide2, this will work. If I have PySide, I will get the error:

> TypeError: 'PySide.QtCore.QCoreApplication.translate' called with wrong argument types:
>   PySide.QtCore.QCoreApplication.translate(str, str, NoneType, int)
> Supported signatures:
>   PySide.QtCore.QCoreApplication.translate(str, str, str = None, PySide.QtCore.QCoreApplication.Encoding = CodecForTr)
>   PySide.QtCore.QCoreApplication.translate(str, str, str, PySide.QtCore.QCoreApplication.Encoding, int)

I was trying to compile .ui files like how it is being done in #88 compiling resources . And this always generates a bunch of translate lines.

mottosso:
Thanks for reporting this @elveatles.

This is an area where I'm open to suggestions. It seems to me that the problem is compiling for a particular binding, and then attempting to run it on another. Does that sound about right?

One way to solve this might be to monkey patch the `QCoreApplication.translate` function. But the problem with this is that it breaks rule 3; co-existence with original bindings. That is, it may break non-Qt.py programs in subtle and unexpected ways.

I wonder whether a more cross-compatible way to handle this is first to compile the `ui` file into a `py` file, and perform some kind of post operation on it via Qt.py.

```bash
$ compile_ui my_widget.ui
$ Qt.py --convert my_widget.py
```

The post operation could *replace* the `translate` function with whatever it is expected to produce using some function that works across every binding.

Alternatively, if it's possible to somehow get to the generated code from the `loadUi` function and friends, and modify it *before* it executes.

Let me know what you think.

elveatles:
For my own code, I ended up monkey patching the translate function. Even though it hasn't caused me problems yet, I am worried about unforeseen consequences. 

[...]

PR #132, opened 2016-09-07
Implement #131

mottosso:
Allright folks, let's get this started!

This command converts a PySide2-compiled .ui file into a Qt.py-compatible equivalent.

```bash
$ pyside2-uic my_ui.ui -o my_ui.py
$ python -m Qt --convert my_ui.py
$ cat my_ui_.py
```

Tested on [this file](https://gist.github.com/salvaom/afe312982140c926c21005d822c0f116#file-widget_designer1_uic2-py).

Also includes a non command-line version.

```python
>>> import Qt
>>> Qt.convert(lines="""\
from PySide2 import QtCore, QtGui, QtWidgets

class Ui_uic(object):
    def setupUi(self, uic):
        self.retranslateUi(uic)

    def retranslateUi(self, uic):
        self.pushButton_2.setText(
            QtWidgets.QApplication.translate("uic", "NOT Ok", None, -1))
""".split("\n")
```

### What we need

1. Run this on your GUIs, see that it works 
2. Where it doesn't work, help me narrow down the exact line that breaks so we can include it.

Let me know what you think!

- #131

### DIFF
--- a/Dockerfile-py2.7
+++ b/Dockerfile-py2.7
@@ -19,6 +19,8 @@ RUN pip install nose nosepipe
 
 # Enable additional output from Qt.py
 ENV QT_VERBOSE true
+ENV QT_TESTING true
+
 # Xvfb
 ENV DISPLAY :99
 
--- a/Dockerfile-py3.5
+++ b/Dockerfile-py3.5
@@ -19,6 +19,7 @@ RUN pip3 install nose nosepipe
 
 # Enable additional output from Qt.py
 ENV QT_VERBOSE true
+ENV QT_TESTING true
 
 # Xvfb
 ENV DISPLAY :99
--- a/Qt.py
+++ b/Qt.py
@@ -31,7 +31,7 @@ Usage:
 import os
 import sys
 
-__version__ = "0.4.2"
+__version__ = "0.4.3"
 
 # All unique members of Qt.py
 __added__ = list()
@@ -57,7 +57,7 @@ def remap(object, name, value, safe=True):
 
     """
 
-    if safe:
+    if os.getenv("QT_TESTING") is not None and safe:
         # Cannot alter original binding.
         if hasattr(object, name):
             raise AttributeError("Cannot override existing name: "
--- a/Qt.py
+++ b/Qt.py
@@ -19,19 +19,20 @@ Default resolution order:
     - PyQt4
 
 Usage:
-    >>> import sys
-    >>> from Qt import QtWidgets
-    >>> app = QtWidgets.QApplication(sys.argv)
-    >>> button = QtWidgets.QPushButton("Hello World")
-    >>> button.show()
-    >>> app.exec_()
+    >> import sys
+    >> from Qt import QtWidgets
+    >> app = QtWidgets.QApplication(sys.argv)
+    >> button = QtWidgets.QPushButton("Hello World")
+    >> button.show()
+    >> app.exec_()
 
 """
 
 import os
 import sys
+import shutil
 
-__version__ = "0.4.3"
+__version__ = "0.5.0"
 
 # All unique members of Qt.py
 __added__ = list()
@@ -84,6 +85,31 @@ def add(object, name, value, safe=True):
     remap(object, name, value, safe)
 
 
+def convert(lines):
+    """Convert compiled .ui file from PySide2 to Qt.py
+
+    Arguments:
+        lines (list): Each line of of .ui file
+
+    Usage:
+        >> with open("myui.py") as f:
+        ..   lines = convert(f.readlines())
+
+    """
+
+    def parse(line):
+        line = line.replace("from PySide2 import", "from Qt import")
+        line = line.replace("QtWidgets.QApplication.translate", "Qt.translate")
+        return line
+
+    parsed = list()
+    for line in lines:
+        line = parse(line)
+        parsed.append(line)
+
+    return parsed
+
+
 def pyqt5():
     import PyQt5.Qt
     from PyQt5 import QtCore, uic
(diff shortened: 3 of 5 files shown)


mottosso:
I'm considering getting my hands on a large library of `.ui` files with which to test this out on. Sort of as an optional "catch-all" in addition to the more refined and specific tests like the one included here.

For that, how about finding say, 10.000 of them from GitHub via [a specific search query](https://github.com/search?utf8=%E2%9C%93&q=%3Cui+version%3D%224.0%22%3E+extension%3Aui&type=Code&ref=searchresults), storing them all in a Gist for reproducibility, and download them during testing to be run through the same conversion mechanism?

mottosso:
More discussion on [Python Inside Maya](https://groups.google.com/forum/#!topic/python_inside_maya/EFee8Qme5WA).

mottosso:
> how about finding say, 10.000 of them from GitHub

This didn't work, here's why.

1. My first approach was to utilise the GitHub Search API, but as it turns out, they've put an artificial limit on searching via the API. The limitation is that each query must include either a user, repository or organisation. This thwarts the attempt, as the idea was based on finding *all* .ui files, regardless of project or any other metric.
2. Since this limitation was not put on the web frontend of their search engine, I figured I'd just scrape it. But, after some investigation, it turns out they have employed clever mechanisms (or at least, cleverer than me) to prevent just such attempts.
3. Finally, I resorted to grabbing just a few and start with that. Turns out, searching for this particular format is (1) not precise enough to guarantee that all files found is actually Qt Designer files and (2) how can I know that these files work from the outset?

So I'll now resort to collecting any .ui files you have lying around. Let me know!

mottosso:
I'm considering releasing this as-is as an alpha.

Currently, when you run it, this happens.

```bash
$ python -m Qt --help
usage: Qt.py [-h] [--convert CONVERT] [--compile COMPILE] [--stdout] [--stdin]

optional arguments:
  -h, --help         show this help message and exit
  --convert CONVERT  Path to compiled Python module, e.g. my_ui.py
  --compile COMPILE  Accept raw .ui file and compile with native PySide2
                     compiler.
  --stdout           Write to stdout instead of file
  --stdin            Read from stdin instead of file
$ python -m Qt --convert my_ui.py
#
# WARNING: --convert is an ALPHA feature.
#
# See https://github.com/mottosso/Qt.py/pull/132
# for details.
#
Creating "my_ui_backup.py"..
Successfully converted "my_ui.py"
$ 
```

Which should be enough to keep people from relying on it, but inviting enough for them to try and find problems for us to expel.

@fredrikaverpil What do you think of all this?

fredrikaverpil:
Ok, I'm back from my honeymoon 😄

I use .ui files (from Qt Designer) and load those as they are without converting them (this means using Qt.py's `load_ui`). So I'm not entirely sure how this should work. I just attempted to do a conversion from .ui to .py:

```bash
$ python -m Qt --convert main_window.ui

#
# WARNING: --convert is an ALPHA feature.
#
# See https://github.com/mottosso/Qt.py/pull/132
# for details.
#
Creating "main_window_backup.ui"..
Successfully converted "main_window.ui"


$ ls -alh main_win*
-rwxr-xr-x  1 fredrik  staff   2.4K Sep 14 07:16 main_window.ui
-rwxr-xr-x  1 fredrik  staff   2.4K Sep 14 07:16 main_window_backup.ui
```

I'm not sure what I'm doing wrong. Perhaps you never go from .ui to .py?

mottosso:
`--convert` takes an already compiled Python module from `pyside2-uic` and makes a Qt.py-compatible version.

fredrikaverpil:
I think it looks fine to merge this, especially since you're mentioning it's an alpha. I can't say I know much about this feature.

fredrikaverpil:
Do you want to include something on this in the README?

mottosso:
Welcome back, by the way! Sorry, was deep in thought as I wrote the reply.

The feature is in relation to #99 and #129, and #131 is the source issue.

In short, some users, like yourself, prefer `load_ui` whereas others prefer compiling to Python. But compiling to Python means embedding traces from the original binding into the resulting code, like `from PySide2 import QtGui`.

This is an attempt at solving this problem by way of the command-line, the same environment in which files are compiled in the first place.

I was hoping to also have the `--compile` flag in there to take a `.ui` file and run `pyside2-uic` for you, but I haven't been able to get my hands on a [working version](https://github.com/fredrikaverpil/pyside2-wheels/issues/44) of this yet (hint hint).

Other than that, it needs testing. If you have any `.ui` files lying around, could you put them up in a gist or the like? Then we can fetch those during Travis testing.

fredrikaverpil:
So when you use the `--compile` option you're supposed to be able to get a binding-agnostic/Qt.py compatible .py based on the given .ui file. Ok, that makes sense.

About pyside2-uic, [I think](https://github.com/fredrikaverpil/pyside2-wheels/issues/44#issuecomment-246916419) it's part of pyside-tools, which doesn't seem to be compiled when building [pyside-setup](https://codereview.qt-project.org/gitweb?p=pyside/pyside-setup.git;a=tree) and thus not included in the PySide2 wheel.

But what's the use case for the `--convert` option then?

mottosso:
> So when you use the --compile option you're supposed

`--compile` would be shorthand for:

```bash
# This
$ pyside2-uic my_ui.ui -o my_ui.py

# Is this
$ python -m Qt --compile my_ui.ui
```

It's to highlight the fact that the only compiler you can use is the one from PySide2.

> But what's the use case for the --convert option then?

When used in conjunction with `--convert`, you get a one-stop-shop.

```bash
$ python -m Qt --compile --convert my_ui.ui
$ cat my_ui.py
```

fredrikaverpil:
Okay, now I get it. Thanks for that explanation :)

Do you have access to Maya 2017?
I can see pyside2uic is included with Maya's Python site-packages as its own package. You might be able to copy this module into your own site-packages (just for testing purposes).

mottosso:
> Do you have access to Maya 2017?

I do, but I'm looking to include tests for this and need it accessible somehow via the Docker image.

fredrikaverpil:
For completion... I managed to find the built package and manually copy it into my site-packages. Then I also copied the built binary. Details [here](https://github.com/fredrikaverpil/pyside2-wheels/issues/44#issuecomment-247280856).

mottosso:
Ok, let's get this on the road for testing as-is, then continue working on it through more issues and PRs.

Any objections?

fredrikaverpil:
No objections!

PR #136, opened 2016-09-09
Fix #135

### LINKED ISSUE #135: Remapping doesn't work correctly with python-qt5 library
# Problem

An error occurs when loading the 0.41 version of Qt.py whilst having the [_python-qt5_](https://github.com/pyqt/python-qt5) version of `PyQt5` on the `PYTHONPATH`. The error is:

```python
  File "<nofile>", line 1, in <module>
  File "P:\path\to\Qt.py\Qt.py", line 265, in <module>
    init()
  File "P:\path\to\Qt.py\Qt.py", line 245, in init
    binding = binding()
  File "P:\path\to\Qt.py\Qt.py", line 92, in pyqt5
    add(PyQt5, "__qt_version__", QtCore.QT_VERSION_STR)
  File "P:\path\to\Qt.py\Qt.py", line 78, in add
    remap(object, name, value)
  File "P:\path\to\Qt.py\Qt.py", line 64, in remap
    "%s.%s" % (object.__name__, name))
AttributeError: Cannot override existing name: PyQt5.__qt_version__
```

The root of the problem is because this PyQt5 package already [sets the qt version variable](https://github.com/pyqt/python-qt5/blob/master/PyQt5/__init__.py) in that particular library and it conflicts with Qt.py's remapping.

Also see:
- [Original mention of the error](https://github.com/mottosso/Qt.py/pull/130#issuecomment-246041258)
- [Indication of the problem](https://github.com/mottosso/Qt.py/pull/130#issuecomment-246050776)

mottosso:
Make an exception for https://github.com/pyqt/python-qt5.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -31,7 +31,7 @@ Usage:
 import os
 import sys
 
-__version__ = "0.4.1"
+__version__ = "0.4.2"
 
 # All unique members of Qt.py
 __added__ = list()
@@ -72,10 +72,10 @@ def remap(object, name, value, safe=True):
     setattr(object, name, value)
 
 
-def add(object, name, value):
+def add(object, name, value, safe=True):
     """Identical to :func:`remap` and provided for readability only"""
     __added__.append(name)
-    remap(object, name, value)
+    remap(object, name, value, safe)
 
 
 def pyqt5():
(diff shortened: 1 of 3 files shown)


mottosso:
Let me know if this works for you, @BigRoy.

BigRoy:
This seems to work! :)

Qt.py
```diff
@@ -68,14 +68,20 @@ def remap(object, name, value, safe=True):
             raise AttributeError("%s != 'module': Cannot alter "
                                  "anything but modules" % object)
 
-    __remapped__.append(name)
+    elif hasattr(object, name):
+        # Keep track of modifications
+        __modified__.append(name)
+
+    if name not in __added__:
```
mottosso: This is a minor bugfix; before, members in `__added__` were also in `__remapped__`.

PR #138, opened 2016-09-15
Fix #137

### LINKED ISSUE #137: Cannot override existing name
The error produced when Qt.py is attempting to override members that isn't officially part of a binding, but for one reason or another is there anyway.

```bash
AttributeError: Cannot override existing name: PySide.QtWidgets # 
# Error: AttributeError: Cannot override existing name: PySide.QtWidgets # 
```

### Solution

Offer the user an option to disable this check.

```bash
$ export QT_ALLOW_OVERRIDES=true
$ python -c "import Qt"
```

### Alternative

Another option is to disable run-time checking of this kind altogether, and save it for testing-only. These things do not change at run-time, and are unlikely to change across bindings and versions of bindings.

mottosso:
Disable run-time testing of attribute overrides.

See original issue #137.

### DIFF
--- a/Dockerfile-py2.7
+++ b/Dockerfile-py2.7
@@ -19,6 +19,8 @@ RUN pip install nose nosepipe
 
 # Enable additional output from Qt.py
 ENV QT_VERBOSE true
+ENV QT_IGNORE_OVERRIDES true
+
 # Xvfb
 ENV DISPLAY :99
 
--- a/Dockerfile-py2.7
+++ b/Dockerfile-py2.7
@@ -19,7 +19,7 @@ RUN pip install nose nosepipe
 
 # Enable additional output from Qt.py
 ENV QT_VERBOSE true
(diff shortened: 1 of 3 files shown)


fredrikaverpil:
Um. I'm not clear on when this issue is triggered. From your example, PySide.QtWidgets must've been created by the user somehow?

fredrikaverpil:
Since I've never seen this issue I was just curious when you hit it.

mottosso:
> I'm not clear on when this issue is triggered.

You and me both.

I never expected to experience it, but it happened and when it did I had no clue about how to find the cause. That feeling is what compelled me to disable it at run-time.

Overall it should be harmless. If anything is overwritten in a vanilla session, it would still be caught by our tests.

fredrikaverpil:
Ok, then this looks good to me. :) 👍

Issue #139, opened 2016-09-16
QHeaderView.setResizeMode in PySide2


finmacmini:
Hey Marcus,

So I've been using your Qt.py shim to allow compatibility between PySide and PySide2 -- which has been working great -- thanks for your work. The only thing I noticed that was missing (for our scripts) was the deprecated, but still present in Qt 5 QHeaderView.setResizeMode. I assume PySide2 decided not to include it since it's deprecated.

I added a quick monkey patch to my code to solve the issue (since setSectionResizeMode is the same as setResizeMode).
```python
if Qt.__binding__ in ('PySide2', 'PyQt5'):
    Qt.QtWidgets.QHeaderView.setResizeMode = Qt.QtWidgets.QHeaderView.setSectionResizeMode
```

Ideally this would be handled by Qt.py itself. I looked through you're code and I wasn't sure the best way to implement it according to your design (and subsequently issue a pull request). So I thought I'd contact you to see if it's easy to add.

Anyway, just thought I'd ask.

Cheers,
Mark

mottosso:
Thanks @finmacmini!

This particular incompatibility has been a popular topic (see links below) and in a nutshell the only reason it isn't in Qt.py itself is because an obscure problem were found in monkey patching the way you illustrate, pull-request [#112](https://github.com/mottosso/Qt.py/pull/112) has more information on this.

Due to the popularity, it's possible we may have to take another look at working around or further prove that it does actually cause problems, but at the moment the current recommended approach is listed in [CAVEATS.md](https://github.com/mottosso/Qt.py/blob/master/CAVEATS.md#qtwidgetsqheaderviewsetresizemode).

- https://github.com/mottosso/Qt.py/pull/112
- https://github.com/mottosso/Qt.py/issues/124
- https://github.com/mottosso/Qt.py/issues/109

You can help by either thinking of an alternative workaround or implementation, or somehow finding a way to prove that the obscure error is isolated to this one occasion.

mottosso:
Shortly after typing this up, an idea occurred to me.

Let me know what you think about this.

```python
from Qt import QtWidgets, QtCompat
app = QtWidgets.QApplication(sys.argv)
view = QtWidgets.QTreeWidget()
header = view.header()
QtCompat.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
```

Where `QtCompat` is an added module containing this and other compatibility functionality, away from other modules so as to not have to monkey patch anything.

[...]

PR #140, opened 2016-09-17
Implement #139

mottosso:
This introduces a radical change to counter #139, #109, #112 and #96.

The change involves moving all added members of Qt.py to it's own individual module, `QtShim`.

```python
# Before
import Qt
Qt.setSectionResizeMode()
Qt.__binding__

# After
from Qt import QtShim
QtShim.setSectionResizeMode()
QtShim.__binding__ == "PyQt5"
```

- See [Example](https://github.com/abstractfactory/Qt.py/blob/master/CAVEATS.md#workaround-1)

It also counteracts our current Achilles heel, which is that at some point one of our additions are bound to conflict with native existing members of the binding. For example, if one day PySide decides to add `__qt_version__` to their offering, but have a different type for it, like `tuple`, whereas we have `str`

So I'm on the fence about whether this is the right way to do it. The disadvantage is a more complex code base; the reader must now understand how `self` plays a part, and how `QtShim` even makes it into the `Qt` namespace.

So let me know what you think.

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -254,10 +254,21 @@ AttributeError: 'PySide.QtGui.QHeaderView' object has no attribute 'setSectionRe
 
 ##### Workaround
 
-Use a conditional.
+Use compatibility wrapper.
 
 ```python
 # PySide2
+>>> from Qt import QtWidgets, QtShim
+>>> app = QtWidgets.QApplication(sys.argv)
+>>> view = QtWidgets.QTreeWidget()
+>>> header = view.header()
+>>> QtShim.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
+```
+
+Or a conditional.
+
+```python
+# PyQt5
 >>> from Qt import QtWidgets, __binding__
 >>> app = QtWidgets.QApplication(sys.argv)
 >>> view = QtWidgets.QTreeWidget()
@@ -267,4 +278,3 @@ Use a conditional.
 ... else:
 ...   header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
 ```
-
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -258,11 +258,11 @@ Use compatibility wrapper.
 
 ```python
 # PySide2
->>> from Qt import QtWidgets, QtShim
+>>> from Qt import QtWidgets, QtCompat
 >>> app = QtWidgets.QApplication(sys.argv)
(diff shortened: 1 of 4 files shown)


fredrikaverpil:
This only applies to members we've added ourselves, right?

Could we just make a list of them here before merging, for completeness?
I'm wondering if `QtShim` is the best name though.

mottosso:
> This only applies to members we've added ourselves, right? (so for instance, QtWidgets is not changed but e.g. load_ui is moved?

Yeah.

Remap as per usual, and make our own space for our own members.

> Could we just make a list of them here before merging, for completeness?

Of course, list is in the [README](https://github.com/abstractfactory/Qt.py/blob/d96c7764313e7a634c1536797d21be2156b1b028/README.md#documentation).

> I'm wondering if QtShim is the best name though.

The name is up for grabs. `QtShim` is bad because it has no obvious connection to `Qt.py`. I wanted to name it `Qt.py`, but of course couldn't. `QtCompat` was another option that crossed my mind, but wasn't sure since `load_ui` isn't necessarily a compatibility function. Got anything in mind?

fredrikaverpil:
> Of course, list is in the README.

Ah, excellent. Make it easier to come up with a good name for this.

Hm. How about something like

    from Qt import Custom

I mean.. what are the chances that... hm. I'll have to sleep on it.

mottosso:
How was your sleep? :)

How about..?

```python
# After
from Qt import QtCompat
QtCompat.setSectionResizeMode()
QtCompat.__binding__ == "PyQt5"
```

I think we'll need the `Qt`-prefix on whichever name we choose, to make it blend better with other calls. In the middle of a block of code, it can be difficult to tell where `Custom` or something else without this prefix comes from.

```python
button = QtWidgets.QPushButton("Button")
Custom.someOtherFunction()  # Is this related to Qt?
```

fredrikaverpil:
> `from Qt import QtCompat`

Yep. Sounds good. Let's go for it.
Does this backwards-breaking feature that we should also hit v1.0.0?

mottosso:
It doesn't break backwards compatibility, the curent members still work. Once we hit 1.0, we can remove them.

See the `_maintain_backwards_compatibility` function at the bottom.

fredrikaverpil:
> See the _maintain_backwards_compatibility function at the bottom.

Ah, that's perfect.

> Once we hit 1.0, we can remove them.

Sounds good!

This all looks excellent to me.

Issue #141, opened 2016-09-18
Add pyside2uic to container image


fredrikaverpil:
The current Docker container doesn't have pyside2uic which makes it difficult to implement and test the --compile option.

PR #142, opened 2016-09-18
Add pyside2-uic (executable) and pyside2uic (module)

fredrikaverpil:
Fix https://github.com/mottosso/Qt.py/issues/141

### DIFF
--- a/Dockerfile-py2.7
+++ b/Dockerfile-py2.7
@@ -1,5 +1,8 @@
 FROM ubuntu:16.04
 
+# Needed to install pyside2-tools without issues
+ENV DEBIAN_FRONTEND noninteractive
+
 RUN apt-get update && \
     apt-get install -y \
         software-properties-common && \
@@ -12,7 +15,11 @@ RUN apt-get update && \
         python-pyqt5 \
         python-pyside \
         python-pyside2 \
+        pyside2-tools \
         xvfb
+
+# Make pyside2uic availble for Python 2.x
+RUN cp -avr /usr/lib/python3/dist-packages/pyside2uic /usr/local/lib/python2.7/dist-packages
 
 # Nose is the Python test-runner
 RUN pip install nose nosepipe
--- a/tests.py
+++ b/tests.py
@@ -263,6 +263,7 @@ def test_meta_add():
     assert "MyAttr" not in Qt.__remapped__
     assert "MyAttr" not in Qt.__modified__
 
+
 def test_meta_remap():
     """Qt.remap() appends to __modified__"""
     import types
@@ -275,7 +276,6 @@ def test_meta_remap():
     assert "MyAttr" not in Qt.__added__, Qt.__added__
     assert "MyAttr" in Qt.__remapped__
     assert "MyAttr" not in Qt.__modified__
-
 
 
 def test_meta_add_existing():
--- a/tests.py
+++ b/tests.py
@@ -327,7 +327,7 @@ def test_pyside2uic():
         # Write compiled ui into Python file
         py_filename = 'output.py'
         with open(py_filename, 'w') as py_file:
-            pyside2uic.compileUi(ui_file, py_file, indent=0)
+            pyside2uic.compileUi(ui_file, py_file)
             import output
             # ui_form = output.Ui_Form()  # Create object
         os.remove(py_filename)
--- a/tests.py
+++ b/tests.py
@@ -315,8 +315,16 @@ def test_pyside2uic():
     """pyside2-uic"""
 
     import pyside2uic
+    from cStringIO import StringIO
 
     with open(self.ui_qwidget, 'r') as ui_file:
+        # Read compiled UI into string
+        pyfile_output = StringIO()
+        pyside2uic.compileUi(ui_file, pyfile_output, indent=0)
+        print(pyfile_output.getvalue())
+        exec pyfile_output.getvalue()
+
+        # Write compiled ui into Python file
         py_filename = 'output.py'
         with open(py_filename, 'w') as py_file:
             pyside2uic.compileUi(ui_file, py_file, indent=0)
--- a/tests.py
+++ b/tests.py
@@ -315,9 +315,9 @@ def test_pyside2uic():
     """pyside2-uic"""
 
     import pyside2uic
-    from cStringIO import StringIO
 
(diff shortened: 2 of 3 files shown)


fredrikaverpil:
I realize I need to add a test to this. Will try to do that later this evening.

Dockerfile-py2.7
```diff
@@ -12,8 +15,12 @@ RUN apt-get update && \
         python-pyqt5 \
         python-pyside \
         python-pyside2 \
+        pyside2-tools \
         xvfb
 
+# Make pyside2uic availble for Python 2.x
+RUN cp -avr /usr/lib/python3/dist-packages/pyside2uic /usr/local/lib/python2.7/dist-packages
```
mottosso: Why are the tools installed for Python 3? :O

fredrikaverpil: I really don't know. I asked [the author](https://launchpad.net/~thopiekar) of the pyside2-tools apt package the same thing.

mottosso: Okay, well that's fine. This solution should work, they're Python 2 compatible at least?

fredrikaverpil: That I don't know yet (but assume so). Let me try it out via tests.

fredrikaverpil:
I added a test which compiles an the already existing ui in tests.py (qwidget.ui) into a .py file (output.py):

```python
# -*- coding: utf-8 -*-

# Form implementation generated from reading ui file 'qwidget.ui'
#
# Created: Sun Sep 18 13:22:20 2016
#      by: pyside2-uic  running on PySide2 2.0.0~alpha0
#
# WARNING! All changes made in this file will be lost!

from PySide2 import QtCore, QtGui, QtWidgets

class Ui_Form(object):
	def setupUi(self, Form):
		Form.setObjectName("Form")
		Form.resize(235, 149)
		self.gridLayout = QtWidgets.QGridLayout(Form)
		self.gridLayout.setObjectName("gridLayout")
		self.lineEdit = QtWidgets.QLineEdit(Form)
		self.lineEdit.setObjectName("lineEdit")
		self.gridLayout.addWidget(self.lineEdit, 0, 0, 1, 1)

		self.retranslateUi(Form)
		QtCore.QMetaObject.connectSlotsByName(Form)

	def retranslateUi(self, Form):
		Form.setWindowTitle(QtWidgets.QApplication.translate("Form", "Form", None, -1))
```

I'm both importing this and I'm creating the Ui_Form class. But I'm not sure on how I should run this... and assert its functionality. Any suggestions?

fredrikaverpil:
I'm not sure on how to assert this functionality. Any suggestions?

mottosso:
> I'm not sure on how to assert this functionality. Any suggestions?

Compile something you are familiar with and know the outcome of. Then just `==` them.

Have a look at [this](https://github.com/mottosso/Qt.py/blob/master/tests.py#L189) for example.

fredrikaverpil:
Jotting down this here since I can't find anything on the `pyside2uic.compileUi` class anywhere online:

```python
def compileUi(uifile, pyfile, execute=False, indent=4, from_imports=False):
    """compileUi(uifile, pyfile, execute=False, indent=4, from_imports=False)

    Creates a Python module from a Qt Designer .ui file.

    uifile is a file name or file-like object containing the .ui file.
    pyfile is the file-like object to which the Python code will be written to.
    execute is optionally set to generate extra Python code that allows the
    code to be run as a standalone application.  The default is False.
    indent is the optional indentation width using spaces.  If it is 0 then a
    tab is used.  The default is 4.
    from_imports is optionally set to generate import statements that are
    relative to '.'.
    """

```

fredrikaverpil:
> Compile something you are familiar with and know the outcome of. 

The problem is I've never compiled anything before so I'm not familiar with what the output should be like.

In this case, I think we should be able to assume the output is proper and that it all works. But I'd like someone to confirm that.

tests.py
```diff
@@ -311,6 +311,28 @@ def test_import_from_qtwidgets():
     assert QPushButton.__name__ == "QPushButton", QPushButton
 
 
+def test_pyside2uic():
```
mottosso: Wait, what are we testing here? Are we testing Qt.py, or are we testing pyside2-tools?

We only need to test what we add to the table. In this case, `python -m Qt --compile my_ui.ui` and the `--convert` option.

fredrikaverpil: I'm just testing pyside2uic generally. You can compile via it so I figured that was a good test. I'm **not** testing Qt.py's `--compile` or `--convert`.

fredrikaverpil: Should I just remove my test then?

This PR was just all about adding pyside2-tools to the Dockerfiles.

mottosso: There's no need to include test for this. We can expect the PySide team to test their own tools.

fredrikaverpil: ...since I was copying pyside2uic from Python 3 to Python 2 I figured it would be good to test that. Perhaps that should've been more clear in my test.

mottosso: Ah, but I think that's testing the Docker image within which we run tests for Qt.py. The tests don't belong amongst tests for Qt.py. We're basically producing tests for this particular distribution of pyside2-tools that we got from some random PPA for this particular OS. It doesn't prove anything. We should make sure what we install to run our tests are correct, and do that external from these tests.

fredrikaverpil: Ok, I'll just remove the test then. And we should be good to go and merge this if you feel it looks allright.

mottosso: It would be allright, but [the addition you made](https://github.com/mottosso/Qt.py/pull/142/files#diff-d5856ea6ea00e434586174344f638824R22), is it still uncertain whether or not that works?

We should find that out first, so we know whether we can rely on that the errors it gives us is related to Qt.py, and not the way we installed it.

fredrikaverpil:
I could do a hard reset to my first commit: https://github.com/mottosso/Qt.py/pull/142/commits/410d1f8966a2c30a1de314dfedf6b50a9e3212dd

or we could just merge this as-is.

mottosso:
> I could do a hard reset to my first commit: 410d1f8

It's an option, but we would lose out on the conversation. No harm one way or the other, I'll leave it up to you.

fredrikaverpil:
I'll just merge this :)

PR #150, opened 2016-10-21
fix a bug of load_ui function for pyside

sol87:
If there are custom widgets in ui files, we must register them before loader.load() in pyside. So I make a new function for that.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -29,6 +29,7 @@ Usage:
 """
 
 import os
+import re
 import sys
 import shutil
 
@@ -195,7 +196,7 @@ def _pyside2():
     _remap(QtCore, "QStringListModel", QtGui.QStringListModel)
 
     _add(PySide2, "__binding__", PySide2.__name__)
-    _add(PySide2, "load_ui", lambda fname: QtUiTools.QUiLoader().load(fname))
+    _add(PySide2, "load_ui", _pyside_loadui)
     _add(PySide2, "translate", lambda context, sourceText, disambiguation, n: (
         QtCore.QCoreApplication(context, sourceText,
                                 disambiguation, None, n)))
@@ -227,7 +228,7 @@ def _pyside():
         pass
 
     _add(PySide, "__binding__", PySide.__name__)
-    _add(PySide, "load_ui", lambda fname: QtUiTools.QUiLoader().load(fname))
+    _add(PySide, "load_ui", _pyside_loadui)
     _add(PySide, "translate", lambda context, sourceText, disambiguation, n: (
         QtCore.QCoreApplication(context, sourceText,
                                 disambiguation, None, n)))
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks @sol87! Could you make an example or test for this functionality as well? Have a look in the [`tests.py`](https://github.com/mottosso/Qt.py/blob/master/tests.py#L162) file for how you could write one.

mottosso:
And also take a gander at the [error message](https://travis-ci.org/mottosso/Qt.py/builds/169445813), looks like something went out the window with PySide2 for both Python 2 and 3.

See the [developer guide](https://github.com/mottosso/Qt.py#developer-guide) for how you could run tests locally, of keep committing to your repository to test here.

fredrikaverpil:
Hi @sol87 and thanks for this!

Regarding adding custom widgets support, I did implement that into `load_ui` back in an old [PR](https://github.com/mottosso/Qt.py/pull/81) which we collectively decided to close.

You can view the full `load_ui` implementation I was proposing here: https://github.com/fredrikaverpil/Qt.py/blob/3daabe95bff0e09454595899543a3ab1d75c4e70/Qt.py

However, this was not merged. We came to the conclusion that this would add functionality to Qt.py which wasn't maintained by PySide/PyQt/Qt but instead had to be maintained by us. For end-users this would mean relying on features which were actually not developed in co-operation with neither the PySide development team or the PyQt development team. And the same goes for adding the "base_instance" argument.

We actually [defined](https://github.com/mottosso/Qt.py/pull/81#issuecomment-242671548) some of our guidelines based on that discussion:

> Henceforth, Qt.py is about linking together commonalities between PySide2 and other bindings, and documenting differences, so as to foster a mindful developer of cross-compatible software.
>
> This is in contrast to wrapping all functionality that differs PySide2 with our own ad-hoc implementations, which we believe will cause more problems than it solves long-term.

This should've been added to the [development guidelines](), which @mottosso already mentioned, but now when I glance through it, I can't see it being mentioned so we should probably fix that.

Okay... back to my story. Since I personally set out to add a base instance argument to load_ui, I decided it would be useful to at least add examples of how load_ui could be extended to support this (although not making load_ui) support it out of the box, and therefore steering users in a more compatible direction in terms of development. These examples are part of Qt.py testing and reside here: https://github.com/mottosso/Qt.py/tree/master/examples/load_ui

I propose that instead of adding custom widgets support to Qt.py we add another example, where it is shown how such functionality can be added to load_ui.

Please let me know what you think.

sol87:
Sure. I will make a example file soon.

sol87:
Hi @fredrikaverpil . I have checked your hack. It seems cool. I believe you make this decision based on good reasons :)

mottosso:
Well put, @fredrikaverpil, and thanks for understanding @sol87!

Issue #152, opened 2016-10-29
Enforce PySide2 API


melevine:
While I think this could be considered a feature request, you can also take this as a question.  In the README under portability, you describe reasons for not removing member from existing modules.

> Qt.py does not hide members from the original binding. This can be problematic if, for example, you accidentally use a member that only exists PyQt5 and later try running your software with a different binding.

When I read that I immediately think that is a great argument for enforcing a common API.  You're probably not going to hit 100% compatibility, but you can probably get it close.  A similar project QtPy seems to approach it in that manner.  Here's an example, if I'm developing an application in a Qt4 environment using Qt.py, It's very easy to accidentally write code that won't work in a Qt5 environment.

```
>>> import Qt
>>> Qt.QtWidgets.QWidget
<class 'PyQt4.QtGui.QWidget'>
>>> Qt.QtGui.QWidget
<class 'PyQt4.QtGui.QWidget'>
```

If I want to write even _simple_ code that will work under both Qt4 and Qt5 I have to be very careful to check which module a given class lives in under PySide2.  This is not only more error-proned because it requires the developer to make a decision, it makes converting older code harder because PySide/PyQt4 code implicitly has classes in the wrong modules.  (Assuming we're saying that the PySide2 is correct).

Anyway, I'd certainly love to see a stricter api enforcement, but it would be good to get a better explanation as to why this is a bad idea.  My organization is also in the VFX industry and we're trying to decide whether to go with QtPy or Qt.py and this is a big sticking point for me.

mottosso:
Thanks for this, @melevine.

I've also been giving this some thought since the inception of Qt.py and I think you're right. 

I can see this happen, with one notable obstacle.

```python
QtWidgets = QtGui
```

This is what happens currently, which is really easy from an implementation standpoint and also eliminates any possibility of (1) missing something and (2) overlooking some of the magic taking place in the C++ glue that makes `QtWidgets` into `QtWidgets`.

The downside is, as you noticed, it includes members that are not supposed to be part of PySide2.

The way I've been meaning to approach it is like how QtPy approached it - which is to create a new namespace and install members individually.

```python
QtWidgets = types.ModuleType("QtWidgets")
QtWidgets.QWidget = QtGui.QWidget
```

I think this could work.

[...]

PR #153, opened 2016-10-30
Cosmetics

### LINKED ISSUE #148: QtCompat translate.
Hey guys, it's me again.

Think there's a bug when adding the translate function to QtCompat.

`_add(PySide2, "translate", lambda context, sourceText, disambiguation, n: (
        QtCore.QCoreApplication(context, sourceText,
                                disambiguation, None, n)))`

It should be 'QtCore.QCoreApplication.translate' right? Missing the translate...

If I change it I get an argument type error and I haven't dug into that...

Thanks for your time.

mottosso:
I was about to have a look at #152 when I got lost reading the source.

I've attempted to make the file more readable to both newcomers and future selves; with the aim to have it read in a top-to-bottom order and be perfectly understandable. If it still isn't, then that's a bug.
- Environment variables put on-top
- QtCompat more explicitly defined on-top
- Lambdas expanded
- Clarify project and development rules (As per https://github.com/mottosso/Qt.py/pull/150#issuecomment-255310565)

I also took the liberty to resolve #148 on the broken `.translate()`, but it still hasn't got any tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Qt.py was born to address the growing needs in these industries for the developm
 - [Development goals](#development-goals)
   - [Support co-existence](#support-co-existence)
   - [Keep it simple](#keep-it-simple)
-  - [No bugs](#no-bugs)
+  - [No wrappers](#no-wrappers)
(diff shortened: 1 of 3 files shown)


fredrikaverpil:
Nice!

Issue #162, opened 2016-11-22
Universal wheel


fredrikaverpil:
Edit 1: posted this prematurely (hotkey on commute...), please refresh!
Edit 2: structured up findings into one single post

### Goal

Offer one single, universal, wheel for Qt.py: `Qt.py-0.6.6-py2.py3-none-any.whl`.

### Background

I'm maintaining an internal wheel "registry" at my company and I recently realized there are one wheel for each major version of Python:

```
Qt.py-0.6.6-py2-none-any.whl
Qt.py-0.6.6-py3-none-any.whl
```

If you attempt to install e.g. the py3 version into Python 2.x you'll get this:

```bash
pip2 install Qt.py-0.6.6-py3-none-any.whl
Qt.py-0.6.6-py3-none-any.whl is not a supported wheel on this platform.
```

So far, I haven't looked into the pip deployment at all, but there seems to be a possibility to have Qt.py use one and the same wheel for either version, which would be nice. An example of another open source package doing this is [virtualenv](https://github.com/pypa/virtualenv). Here you can see only one wheel is available and it's called `virtualenv-15.1.0-py2.py3-none-any.whl`.

### Implementation

It seems we need to do two things:

1. Add `setup.cfg` with the following contents:

```
[bdist_wheel]
universal=1
```

2. Change [this line](https://github.com/mottosso/Qt.py/blob/master/.travis.yml#L26) of the Travis PyPi deployment into:

```
distributions: "sdist bdist_wheel"
```

Sources:
* https://wheel.readthedocs.io/en/latest/
* http://www.dougalmatthews.com/2016/Sep/01/automate-publishing-to-pypi-with-pbr-and-travis/
*  https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/

___

@mottosso I believe you have a better overall understanding (than me) on how the PyPi deployment works and how to test it out. Would you like to have a look at this? (there's absolutely no hurry on my part on getting this implemented)

fredrikaverpil:
Hm. It seems like all we have to do is add `setup.cfg` with the following contents:

```
[bdist_wheel]
universal = 1
```

Please see [here](https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/) where he explains:

> The first part will make wheel build a universal wheel file (e.g. attrs-15.1.0-py2.py3-none-any.whl) and you won’t have to circle through virtual environments of all supported Python versions to build them separately.

[...]

PR #163, opened 2016-11-24
Feature/universal wheel

fredrikaverpil:
Referring back to issue #162 - here's how we could avoid wheel fragmentation of Qt.py since it's a pure Python module.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -69,7 +69,7 @@ self.__remapped__ = list()  # Members copied from elsewhere
 self.__modified__ = list()  # Existing members modified in some way
 
 # Below members are set dynamically on import relative the original binding.
-self.__version__ = "0.6.6"
+self.__version__ = "0.6.7-test-01"
 self.__qt_version__ = "0.0.0"
 self.__binding__ = "None"
 self.__binding_version__ = "0.0.0"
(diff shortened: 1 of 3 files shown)


mottosso:
Ok, so the version picked up by PyPI (the irreversible and forcible unique identifier) is the one from `setup.py`.

As we test, we'll (1) set this manually to something temporary that we can increment, like `0.6.7-test-01` and (2) merge it, (3) make a release, called the same as the temporary version. Once that's done, Travis will attempt to automatically upload it to PyPI, and from there we can install it and see whether it works.

It's a bit of a terrible way of testing, but I can't see any other way. Once we've merged a couple of times and we see that it's working, we can remove the tags from the repo, along with the releases on GitHub and set the version back to something sensible.

Qt.py
```diff
@@ -69,7 +69,7 @@
 self.__modified__ = list()  # Existing members modified in some way
 
 # Below members are set dynamically on import relative the original binding.
-self.__version__ = "0.6.6"
+self.__version__ = "0.6.7-test-01"
```
mottosso: You can set it here, or in setup.py. Either is ok.

fredrikaverpil: Yup, it was done in 15a6eab. Looks ok?

fredrikaverpil:
> It's a bit of a terrible way of testing

Indeed, but I don't see any other way either.

So, are we ready to merge?
Would you like to do the honors? 😄

mottosso:
It's all yours. :)

PR #165, opened 2016-11-24
Fix QtCompat attributes

### LINKED ISSUE #164: Issues with querying binding versions
I've been on a really old version of Qt.py for a long time and I just noticed that now when I upgraded to Qt.py 0.6.6 that I get this:

```python
>>> from Qt import QtCompat
>>> print(QtCompat.__binding_version__)
0.0.0
>>> print(QtCompat.__qt_version__)
0.0.0
```

When peeking into the code, it seems these attributes are simply never remapped (just their default 0.0.0 values are set).

Before I embark on creating a PR to fix this, was there a specific reason this got changed (or perhaps it was just forgotten when setting up QtCompat)?

fredrikaverpil:
Fix #164

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -69,7 +69,7 @@ self.__remapped__ = list()  # Members copied from elsewhere
 self.__modified__ = list()  # Existing members modified in some way
 
 # Below members are set dynamically on import relative the original binding.
-self.__version__ = "0.6.6"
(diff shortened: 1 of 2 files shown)


mottosso:
Great work, could you version-up, and make a new release mentioning what changed?

fredrikaverpil:
Sure thing!

PR #172, opened 2016-12-15
fix QT_TESTING test 

sol-ansano-kim:
Hello

We are trying to unify our Qt UI tools to Qt.py.
It seems testing of QT_TESTING in _remap() function always returns True as it is tested against None but the bool() call converts None to False.
Here is my simple fixing.

If possible, could you specify more explicitly the expected or valid values for QT_VERBOSE, QT_TESTING, like on/off or 0/1? I think this would make things clearer.

Thanks,
Sol Kim.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -69,7 +69,7 @@ self.__remapped__ = list()  # Members copied from elsewhere
 self.__modified__ = list()  # Existing members modified in some way
 
 # Below members are set dynamically on import relative the original binding.
-self.__version__ = "0.6.8"
+self.__version__ = "0.6.9"
 self.__qt_version__ = "0.0.0"
 self.__binding__ = "None"
 self.__binding_version__ = "0.0.0"
--- a/Qt.py
+++ b/Qt.py
@@ -124,7 +124,7 @@ def _remap(object, name, value, safe=True):
 
     """
 
-    if QT_TESTING is not None and safe:
+    if QT_TESTING and safe:
         # Cannot alter original binding.
         if hasattr(object, name):
             raise AttributeError("Cannot override existing name: "


mottosso:
Thanks @sol-ansano-kim, that's a great catch.

It seems the `if` statement got left behind when `QT_TESTING` got translated to a `bool()`. You are entirely correct.

Would you mind incrementing the version for this as well? Then I can make a new release.

On a separate note, `QT_TESTING` is both undocumented and isn't for public use. I'd be happy to make it public if you could let me know how you are looking to use it?

sol-ansano-kim:
No, not at all.

About QT_TESTING usage, 
If QT_TESTING means prevention overwriting of existing member also, 
when some tool or software include another version Qt.py already(like houdini 16), we can`t  guarantee to two remapped function are fully identical and which one remains last. Almost cases False will be OK, but I think sometime we maybe want to warn or ban it. 

But the QT_TESTING means only for Qt.py testing, I think we can choice other way for prevention.

mottosso:
> when some tool or software include another version Qt.py already(like houdini 16), we can`t guarantee to two remapped function are fully identical and which one remains last.

What should happen, is that only one of them should ever get imported. If Houdini does this upon loading, then your code should reuse their version.

> Almost cases False will be OK, but I think sometime we maybe want to warn or ban it.

I think you're right. There is another ongoing issue on preventing accidental use that you might be interested in:

- https://github.com/mottosso/Qt.py/issues/152

In any case, this PR fixes an important bug and I'd be happy to merge it.

Could you go [here](https://github.com/mottosso/Qt.py/pull/172/commits/6ccc4f4c1f8ecef7876a8949364cc29d894603a9#diff-cd1d7c0a982bce680146d5b248151256R72) and increment version `0.6.8` to `0.6.9`? That way I can merge and then make a new release.

Thanks @sol-ansano-kim!

sol-ansano-kim:
> If Houdini does this upon loading, then your code should reuse their version.

Yes, you`re right. My problem was that my code had read another module which load Qt.py.

> Could you go here and increment version
Sorry, I fogot it.I did it.

Thanks.

mottosso:
I also forgot something.

> If possible, could you specify more explicitly the expected or valid values for QT_VERBOSE, QT_TESTING, like on/off or 0/1? I think this would make things clearer.

This. What do you think about a comment on each variable?

In a nutshell, the value is simply that they exist in the environment. For example, each of these are valid.

```bash
$ export QT_TESTING=exists
$ export QT_TESTING=1
$ export QT_TESTING=True
$ export QT_TESTING=anything
```

If it doesn't exist, it's `False`.

sol-ansano-kim:
I just think it can be misunderstood, for example QT_TESTING=0 after set QT_TESTING=1.
However as you said, it`s un-public and we won`t use it usually, so I think that it has no problem.

mottosso:
`QT_TESTING=0` is also `True`, simply because the variable exists (and is always a string, in Python).

```bash
$ set MY_VARIABLE=0
$ python
>>> import os
>>> bool(os.getenv("MY_VARIABLE"))
True
```

Ok, let's leave it for now. Thanks for bringing it up!

PR #174, opened 2016-12-28
CGRU project added (http://cgru.info/)

timurhai:
It is not a company.
But it is a tools pack for CG too.
Main tool is Afanasy - free and open-source render farm manager.


mottosso:
Thanks @timurhai! 

Do you think this would be better suited for 'Projects using Qt.py' section?

timurhai:
Hi!
I think that yes.
But some studios can use in-house tools even w/o any name.
May be better to have both sections, and projects and studios.
Also famous studios can attract more developers ))

mottosso:
Both could work, one for the company and one for the product.

PR #178, opened 2017-01-29
Move "distributions" out of "on" part of "deploy"

greatestape:
See the URL below for docs on where travis expects to find "distributions":
https://docs.travis-ci.com/user/deployment/pypi/#Uploading-different-distributions

See issue #162


mottosso:
Thanks for this, I'll give this a try on the next release!

fredrikaverpil:
Thanks @greatestape !
Was this changed recently?

mottosso:
Ok, the only way we can find out whether this works is to try. So for the next release, let's keep an eye out.

Issue #53, opened 2016-06-03
Should there be a wrapper around sip/shiboken?


fredrikaverpil:
Originally asked by Pierre Augeard:

> What would be the wrapper around sip/shiboken?
> 
> For example:
> - http://nathanhorne.com/pyqtpyside-wrap-instance/
> - https://github.com/rgalanakis/practicalmayapython/blob/master/src/chapter5/qtshim.py

mottosso:
> Originally asked by Pierre Augeard:

Where? :O

fredrikaverpil:
> Where

Over at the Maya beta forums, but now I see he's interacting here. Ping @pipoun

fredrikaverpil:
@pipoun - could you please provide a use case for this request/question?

pipoun:
```py
from Qt import QtGui, binder
import maya.OpenMayaUI


def get_maya_mainwindow():
    """Get the main Maya window

    :return: :class:`QtGui.QMainWindow` instance of the top level Maya window
    """
    ptr = maya.OpenMayaUI.MQtUtil.mainWindow()
    if ptr is not None:
        # wrap_instance (pep8), wrapInstance (shiboken) or wrapinstance (sip)
        return binder.wrap_instance(long(ptr), QtGui.QMainWindow)
```

fredrikaverpil:
@pipoun do you ever use this for anything except returning the main Maya window?

mottosso:
A portable approach would be preferable. However..

> Would you consider this approach

This returns the script editor for me.

```python
from Qt import QtWidgets
window = QtWidgets.QApplication.activeWindow().objectName()
print(window)
```

Unless I have another window active, in which case that window is returned.

There is this.

```python
from Qt import QtWidgets
window = {w.objectName(): w for w in QtWidgets.qApp.topLevelWidgets()}["MayaWindow"]
print(window)
```

pipoun:
I just wanted to focus on the fact that `sip.wrapinstance` and `shiboken.wrapInstance` have not the same spelling nor the exact same behavior and it should be considered to have a wrapper to unify all of this ?

> @pipoun do you ever use this for anything except returning the main Maya window when you want to parent your UI to Maya?

There are other use cases than getting a main window as a `QMainWindow` instance.

`QtWidgets.qApp.topLevelWidgets()` is very slow in some cases compared to `MQtUtil.mainWindow` (when there are many Qt widgets init'd), see the implementation :
[...]

PR #180, opened 2017-01-30
Exposing select functions through a new c_binding module

boglin:
A possible approach to addressing #53

- Addition of Qt.c_binding module with functions:
    - wrapInstance
    - getCppPointer (the compliment of wrapInstance)
    - delete
- Graceful handling of the case when the binding is not available.


mottosso:
Thanks for this @boglin. 

Could we put these members in the QtCompat module, and give them a snake_case name?

That way they would comply with other wrappers already in there.

> Graceful handling of the case when the binding is not available.

In your experience,  when would such a binding not be available? I'm very inexperienced with sip and shiboken and thought that they would have had to be available to make a binding work.

mottosso:
I've addressed some of the usecases in the #53. Could you have a look and let me know if this wrapper provides unique value?

mottosso:
Superseded by #190

Issue #177, opened 2017-01-25
__path__ member on Qt module


osstony:
Thank you for creating and continuing to support this project.

I've updated to the 1.0.0.b1 release and ran into a minor issue. I've used the \_\_path__ property on the PySide module in the past and was hoping to continue to do so. The previous version of Qt.py allowed for this by using Qt.\_\_path__. 

I'm hoping this could be made possible again. If setting Qt.\_\_path__ doesn't feel appropriate as it's not truly the path to the Qt.py module then perhaps a member could be added called \_\_binding_path__, which would be set to the bound module's \_\_path__ variable.

*Edited to fix double underscore markdown issue.

mottosso:
Certainly, @osstony. Just to clarify, you are looking for the path to the original binding, and not the path to Qt.py?

I like the `__binding_path__` idea. Feel free to make pull request with this, and increment the version to b2.

osstony:
That is correct, @mottosso.

I'll add it in and get a pull request up soon, thanks!

fredrikaverpil:
Shouldn't `__path__` just work if it exist in all the bindings?

```python
>>> import PySide2
>>> PySide2.__path__
['/usr/local/lib/python2.7/site-packages/PySide2']
```

```python
>>> import PyQt5
>>> PyQt5.__path__
['/usr/local/lib/python3.5/site-packages/PyQt5']
```

```python
>>> import PySide
>>> PySide.__path__
['/usr/local/lib/python2.7/site-packages/PySide']
```

osstony:
Yep, it should work but __path__ is not an available member on the Qt module any longer as of 1.0.0.b1. It was available in previous versions when the Qt module was effectively being set to the binding module.

fredrikaverpil:
Hm. It could also be kind of weird if Qt.__path__ didn't return the path to `Qt.py`.

Technically, Qt.py isn't replacing itself with the desired binding anymore.

Perhaps it would be better to actually put `__path__` on `Qt.QtCompat.__path__`...?

osstony:
Yeah I definitely agree. Adding `__binding_path__` would allow easy access to the path of the binding package and not add unexpected results like `__path__` would.

fredrikaverpil:
> Yeah I definitely agree. Adding __binding_path__ would allow easy access to the path of the binding package and not add unexpected results like __path__ would.

👍

mottosso:
What are your thoughts on this?

```python
import Qt
[...]

PR #182, opened 2017-02-08
Implement #177

mottosso:
#177

Alternatively.

```python
import Qt
binding = __import__(Qt.__binding__)
binding.__path__
```

Which do you feel is most intuitive?


mottosso:
See discussion for why the hold status at https://github.com/mottosso/Qt.py/issues/177

Issue #179, opened 2017-01-30
mixedCase naming convention


mottosso:
Currently, `load_ui` is snake_case but Qt and it's bindings are all mixedCase.

I originally chose this convention so as to conform with PEP8 but in retrospect it doesn't seem appropriate as it intermixed two conventions in one library. 

This issue then is about conforming new members of QtCompat to members of other submodules, e.g. QtCore and QtGui.

E.g. `loadUi`

fredrikaverpil:
Hm. I thought it was good not to confuse `load_ui` with e.g. `PyQt4.uic.loadUi` as the behaviour is not 100% identical.

I like the snake case :) primarily because that makes it easy to spot what's part of Qt.py and what's not. But it's not something I would fight over. You decide :)

mottosso:
Good point, good point. It's worth taking into account.

My motivation for wanting mixedCase everywhere is part cleanliness but
mainly discoverability. At this point, *some* members of QtCompat are
mixedCase, such as `setSectionResizeMode` whereas others are snake case.
Which means it isn't enough to remember what a member was called, but also
which case convention it has. It adds unessecary cognitive load.

Let's think about it.

fredrikaverpil:
Yes, I see the dilemma.

I know what the pep says, but let's not get too tangled up in that. The pep is really just guidelines in this case. We're going past that here and need to define our own rules.

Right now I think it would be very weird to rename `setSectionResizeMode` to `set_section_resize_mode`. This wouldn't make any sense.

We could very well conform to mixedCase, such as renaming `load_ui` into `loadUi`. The `translate` function is already named correctly for both snake_case and mixedCase.

But when talking about the functions within `Qt.py`, `tests.py`, `run_tests.py`, I would propose we should keep it snake_case. This all means the following;
* User-facing functions are mixedCase
* Internal Qt.py functions are snake_case

fredrikaverpil:
Also, we could rename `load_ui` into something which cannot easily be confused with `loadUi`(which exists). We could call it e.g. `loadUiFile` (it *does* require a filename at all times).

bpabel:
I'm not sure you need to change the naming, since it's essentially replacing that function and the namespace will be different.

mottosso:
Ok, sounds like we're in agreement on this convention.

```python
from Qt import QtCompat
QtCompat.loadUi(...)
QtCompat.setSectionResizeMode(...)
```

[...]

PR #183, opened 2017-02-09
Implement #179

mottosso:
This makes the naming convention across original and added members uniform.

- #179

```python
from Qt import QtCompat

# Before
QtCompat.load_ui(...)

# After
QtCompat.loadUi(...)
```

The convention is this:

1. mixedCase across the board
2. snake_case for dunder names

As dunder names are unique to Python, they remain [aligned with PEP8](https://www.python.org/dev/peps/pep-0008/#id35), though it's open for discussion whether these should be dunder names at all. E.g. `__binding_version__` could instead be `bindingVersion`.

Thoughts?

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -63,7 +63,7 @@ import types
 import shutil
 import importlib
 
-__version__ = "1.0.0.b2"
+__version__ = "1.0.0.b3"
 
 # Enable support for `from Qt import *`
 __all__ = []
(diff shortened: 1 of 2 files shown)


mottosso:
This is backwards compatible, `QtCompat.load_ui` will continue to be available but is to be considered deprecated.

fredrikaverpil:
I'm now using `QtCompat.loadUi` and it works fine for me so far. 👍

fredrikaverpil:
We'll need to update the examples: https://github.com/mottosso/Qt.py/tree/master/examples/load_ui

mottosso:
That's right, could you take care of it?

fredrikaverpil:
Sure, I won't be able to take a look at it until next week though.

mottosso:
That's allright. There's no rush.

Issue #188, opened 2017-03-28
Override to Disable Setting of API2 for PyQt4


dgovil:
So this is a minor request but currently the `_pyqt4()` function tries to set the sip api to v2 by default.
I was thinking we could add a env var to disable this behavior. By default the env variable would be disabled so that the current behavior is preserved but when it is enabled it would disable the api change.

Something like `QT_USE_API1` which if set to 1, will disable that behavior.

This would let Qt.py be included in bootstrap files without breaking existing code. Otherwise it makes porting legacy code hard.

I can make a pull request if there's no opposition.

Cheers

dgovil:
See pull request: https://github.com/mottosso/Qt.py/pull/189

mottosso:
Thanks @dgovil.

I might not fully understand your use case, but it appears to me that enabling the use of sip v1 would really *hinder* porting of legacy code. Not facilitate it.

For example, the compatibility matrix for sip v1 looks like this.

| Sip  | PyQt4 | PyQt5 | PySide | PySide2
|:-----|:-------:|:------:|:-------:|:---------:
| v1   | X         |           |              |
| v2   | X         | X         | X          | X

Where every binding except PyQt4 uses it by default. Which means that, if you were to enable sip v1 with Qt.py, Qt.py would be usable *only* with PyQt4.

If you could take me through the steps you envision in porting of legacy code with sip v1 enabled, maybe that would provide more some clarity.

dgovil:
Our current situation is that we have lots of code that is not versioned specific to whether it's being consumed by an api1 or api2 environment. Part of this is because they either don't use api1 only features like qvariant or that they cast variables to a Python type for example with `str(Foo)` so would work regardless of API version.

The problem comes if you try and load qt.py in an environment that expects API v1 for even a single script and it suddenly falls apart when it tries to access API v1 features. That makes it hard to let code that uses pyqt4 directly coexist with code that uses qt.py instead till we port everything over at once.

Having an environment variable switch would let us determine when to force the API switch over completely. 

In my experience it's easier to have the API set to v1 and start porting to v2 then the other way around.

Totally agree that it is not something someone should be doing during regular usage, which is why I think the default should be to stay on api2, but I think modifying someone's global environment should also be optional.

mottosso:
Thanks @dgovil, that makes sense.

Onto the [implementation](https://github.com/mottosso/Qt.py/issues/188#issuecomment-290138148).

PR #189, opened 2017-03-29
Added QT_PRESERVE_API to prevent changing of API versions

dgovil:
During the porting of legacy code to Qt.py the changing of API versions
causes issues. By optionally preserving it, we can control that behavior.

This is primarily an issue when porting legacy code to use Qt where we are in an environment where importing Qt in bootstraps could cause issues when it changed the api.

https://github.com/mottosso/Qt.py/issues/188

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -71,7 +71,7 @@ __all__ = []
 # Flags from environment variables
 QT_VERBOSE = bool(os.getenv("QT_VERBOSE"))
 QT_PREFERRED_BINDING = os.getenv("QT_PREFERRED_BINDING", "")
-QT_SIP_API_HINT = int(os.getenv("QT_SIP_API_HINT") or 2) # Set using OR 2 in case env is set but empty string
+QT_SIP_API_HINT = os.getenv("QT_SIP_API_HINT")
 
 # Reference to Qt.py
 Qt = sys.modules[__name__]
@@ -764,20 +764,28 @@ def _pyqt4():
     """Initialise PyQt4"""
     
     import sip
-    try:
-        sip.setapi("QString", QT_SIP_API_HINT)
-        sip.setapi("QVariant", QT_SIP_API_HINT)
-        sip.setapi("QDate", QT_SIP_API_HINT)
-        sip.setapi("QDateTime", QT_SIP_API_HINT)
-        sip.setapi("QTextStream", QT_SIP_API_HINT)
-        sip.setapi("QTime", QT_SIP_API_HINT)
-        sip.setapi("QUrl", QT_SIP_API_HINT)
-    except AttributeError as e:
-        raise ImportError(str(e))
-        # PyQt4 < v4.6
-    except ValueError as e:
-        # API version already set to a different version
-        raise ImportError(str(e))
+
+    # Validation of env var. Prevents an error if the env var is invalid since it's just a hint.
+    api_hint = 2
+    if QT_SIP_API_HINT in ('1', '2'):
+        api_hint = int(QT_SIP_API_HINT)
+    elif QT_SIP_API_HINT:
+        _log("Warning: QT_SIP_API_HINT is not a valid value of 1 or 2: [%s]" % (QT_SIP_API_HINT))
+
+    api_names = ('QString', 'QVariant', 'QDate', 'QDateTime', 'QTextStream', 'QTime', 'QUrl')
+
+    for api_name in api_names:
+        try:
+            sip.setapi(api_name, api_hint)
+        except AttributeError as e:
+            raise ImportError(str(e))
+            # PyQt4 < v4.6
+        except ValueError as e:
+            # API version already set to a different version
+            if QT_SIP_API_HINT:
+                _log("Warning: API '%s' has already been set to %s" % (api_name, sip.getapi(api_name)))
+            else:
+                raise ImportError(str(e))
 
     import PyQt4 as module
     _setup(module, ["uic"])
--- a/Qt.py
+++ b/Qt.py
@@ -783,7 +783,7 @@ def _pyqt4():
         except ValueError as e:
             # API version already set to a different version
             if QT_SIP_API_HINT:
-                _log("Warning: API '%s' has already been set to %s" % (api_name, sip.getapi(api_name)))
+                print("Warning: API '%s' has already been set to %s" % (api_name, sip.getapi(api_name)))
             else:
                 raise ImportError(str(e))
 
--- a/Qt.py
+++ b/Qt.py
@@ -762,30 +762,39 @@ def _pyqt5():
 
 def _pyqt4():
     """Initialise PyQt4"""
-    
+
     import sip
 
-    # Validation of env var. Prevents an error if the env var is invalid since it's just a hint.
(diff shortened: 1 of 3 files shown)


mottosso:
Thanks @dgovil.

As this switch is so specific to PyQt4 and sip, I'd prefer the variable to have a name more akin to this.

1. `QT_SIP_API_V1`
2. `QT_SIP_API_V1_FOR_PYQT4`
3. `QT_PYQT4_SIP_API_V1`

Though I'm not too keen on either of those. (1) Doesn't work as it implies V1 for all bindings, and it can't work for Qt 5 bindings. (2) Is awefully long and ugly, difficult to remember. (3) is a little repetitive, and difficult to remember.

What else could we use?

dgovil:
Yeah that was my dilemma too on what a good name is. Sorry I should have provided my thinking on it too. Overtime is frying my brain, haha.

The reason I thought of `QT_PRESERVE_API` would be that it's global to not change the API version, regardless of the binding. Technically it would only apply to PyQt4 but I figured it was short/generic enough while still being clear?

## Option 1: Variables with API_V1 in the name

My problem with implying `API_V1` in the env var name would be that it implies the api is being set to v1. I'm not sure if you wanted to explicitly set the version to 1. I felt maybe that would be too many more lines of code for little 

## Option 2: Declaring the API version as the value of the variable

Another alternative could be a variable called `QT_PYQT4_API` which is set to a default of 2 if unset, but instead of the current hardcoded setting of the api to 2 it would set it to the value of this variable. So it would still be 2 by default but would atleast give external control over it.

-------

Finally though the problem would be if someone wanted to set specific parts of the api to 2. Like maybe they only want QString to be 2. I don't know why they would but I'm sure some mad person has found a use for that. In that case I think it would be prudent to give them the flexibility to tell Qt.py to not change the api. 

That's why I thought `QT_PRESERVE_API` was the most flexible option?

To be clear, I think the default should definitely be to set the api to v2 as you have it now, but the option to not set it would be great.

Anyway that was just my thinking on the matter if that helps. I'll defer to your judgement on the naming of course.

mottosso:
Thanks @dgovil those are good points.

In that case, my only concern with `QT_PRESERVE_API` is that it requires the reader to know what version is default for a binding. For us it's obvious that they are all on 2, except PyQt4. But it's only obvious if you've done your research.

How about `QT_SIP_API_HINT=1`?

It suggests an attempt at setting the API, but won't abort if failing. As the API can only be set in PyQt4, and can only be set once, I would expect others to read it as "unless it's already been set, set it to `X`" where `X` is the version specified, in this case 1, and default to 2.

> Finally though the problem would be if someone wanted to set specific parts of the api to 2.

I think it's safe to draw a line here. For such (rare) situations, adjusting the Qt.py source shouldn't be a problem.

dgovil:
Yeah `QT_SIP_API_HINT=1` works. 
How would you envision it working? I was thinking its default value would be 2 (if unset) and then `Qt.py` tries to set it to whatever API hint is provided.

If it fails to set the API version to the hint, would it raise an exception as it does now or just log and continue?

mottosso:
> How would you envision it working?

It does introduce a few circumstances..

I'd imagine this to be the case, letting a hint be a hint rather than forceful. What do you think?

**CASE 1: No hint, `sip=1`**

```bash
$  export QT_PREFERRED_BINDING=PyQt4
$ python
...
>>> import sip
>>> sip.setapi("QString", 1)
>>> import Qt
ImportError: API 'QString' has already been set to version 1
```
**CASE 2: No hint, `sip=2`**

```bash
$  export QT_PREFERRED_BINDING=PyQt4
$ python
...
>>> import sip
>>> sip.setapi("QString", 2)
>>> import Qt
>>> # OK
```

**CASE 3: hint=1, sip=1**

```bash
$ export QT_SIP_API_HINT=1
$ python
...
>>> import sip
>>> sip.setapi("QString", 1)
>>> import Qt
>>> # OK
```

**CASE 4: hint=2, sip=1**

```bash
$ export QT_SIP_API_HINT=2
$ python
...
>>> import sip
>>> sip.setapi("QString", 1)
>>> import Qt
Warning: API 'QString' has already been set to version 1
>>> # OK
```

**CASE 5: hint=2, sip=2**

```bash
$ export QT_SIP_API_HINT=2
$ python
...
>>> import sip
>>> sip.setapi("QString", 2)
>>> import Qt
>>> # OK
```

**CASE 6: hint=1, sip=2**

```bash
$ export QT_SIP_API_HINT=1
$ python
...
>>> import sip
>>> sip.setapi("QString", 2)
>>> import Qt
Warning: API 'QString' has already been set to version 2
>>> # OK
```

dgovil:
Okay let me know how you feel about this implementation. I tried it with each of the Cases above and it worked as expected.  Additionally I added some validation to the envvar so that it's harder to set an unexpected value.

If it's fine with you, I'll write up some tests for it as well.

mottosso:
Thanks @dgovil! I'll have a look at this tomorrow. 

As for tests, yes please. Remember that each test runs independently in its own Python interpreter and with its own environment. That should help you write the tests. It means you can start several QApplication instances and customize the environment as you please in each test.

dgovil:
So I feel dirty about this, but why does `assert_raises` only work in Python 2? I assumed it would work for both but guess not, so the test coverage is a bit dicey.

mottosso:
> So I feel dirty about this, but why does assert_raises only work in Python 2?

It's a limitation of `nose`, it simply isn't implemented for 2.

Qt.py
```diff
@@ -761,22 +762,30 @@ def _pyqt5():
...
+    for api_name in api_names:
+        try:
+            sip.setapi(api_name, api_hint)
+        except AttributeError as e:
+            raise ImportError(str(e))
+            # PyQt4 < v4.6
+        except ValueError as e:
+            # API version already set to a different version
+            if QT_SIP_API_HINT:
+                _log("Warning: API '%s' has already been set to %s" % (api_name, sip.getapi(api_name)))
```
dgovil: Yeah I agree it's not the prettiest but I can't think of a good way to consolidate it further.
As for printing always instead of logging, I think you're right. It is an important enough message that it shouldn't be invisible.

I'll make that change later today.

dgovil:
Changed the _log to a print

mottosso:
I made a few cosmetic tweaks and added a test for the warning, along with 1 functional change.

The change is that when the `QT_SIP_API_HINT` is invalid, such as `=blargh` it throws an `ImportError` whereas before it logged it. I think this is the expected behaviour when configuring this variable and think it's safe to assume that if it exists it is never meant to be invalid. This would immediately show the *developer* that there's a mistake.

I also made the warning come from `stderr` as opposed to `stdout` which happens with `print()`.

Finally, given your position, there's some legal cruft I have to consider. If you could take a look at the (default) CLA and click the sign button, that'd secure Qt.py from any future trouble. Let me know if there are any issues.

dgovil:
Cool, the changes look like nice improvements. I also just agreed to the CLA.

mottosso:
Sweet beans, this is done! Thanks @dgovil!

PR #196, opened 2017-04-06
Baseinstance support for all bindings

dgovil:
This adds support for the baseinstance argument across all 4 bindings. 
It's based on the implementation [here](https://gist.github.com/cpbotha/1b42a20c8f3eb9bb7cb8) but removes any parameters that don't directly map between bindings.

It's not the prettiest implementation, so I'd love feedback on ways it could be better. 
I also broke from style a bit and used double underscore so that it doesn't get added to the Qt namespace. Let me know if you'd prefer single underscore to be inline with the other 'hidden' functions.

Also I'm not sure how to write a test for this. Do you want one? I'm not sure if you want to package a dummy ui file with it and load it on to the instance.


fredrikaverpil:
@dgovil awesome, code looks very clean. Could you please add tests for this?

mottosso:
@dgovil see [this test](https://github.com/mottosso/Qt.py/blob/1bd56298a5a8c45ff84d09189f0236fd99a991f2/tests.py#L93) for an example of how to test without external files.

Ideally though, I'd like the loader to take a string of the full ui in addition to a file path. Maybe for another PR.

dgovil:
I'm not 100% sure if it's possible with the UILoader to take a string input directly. Does uic allow that? I didn't see it in the docs.

As for tests, I can definitely add some. I'm going to be away from my PC for a week though but if noone beats me to it, I'll get it done then.

mottosso:
> I'm not 100% sure if it's possible with the UILoader to take a string input directly. Does uic allow that? I didn't see it in the docs.

It likely doesn't, it'd have to either (1) wrapped or (2) copy/pasted from source, as it's pure Python anyway.

Wrapped is likely the more suitable option, I'll post an example of current thoughts, but would stress that its implementation is likely best suited for a separate issue and PR.

```python
# Psuedo, untested
def loadUi(source):
  if os.path.isfile(source):
    return UiLoader.loadUi(source)
  else:
    with tempfile.NamedTemporaryFile(delete=False) as f:
      f.write(source)
    try:
      return UiLoader.loadUi(f.name)
    finally:
      os.remove(f.name)
```

dgovil:
@mottosso I've added a test for baseinstance loading. It's fairly simple but it checks to see if the ui was loaded onto the instance correctly.

mottosso:
Thanks @dgovil, I think there is more to test for with this functionality.

Try and break it in various ways and ensure you've gotten the corner cases. We also need tests that check for differences across bindings such that we may document them properly.

Off the top of my head, we need to check:

1. Overwriting; if a widget, signal, property or slot already exists, is it overwritten, augmented, ignored?
2. Does signals created in Qt Designer work once loaded?
3. What happens when an invalid .ui file is passed?
4. What happens when a really large (2 gb) .ui file is passed?
4. What happens when a .ui file from an older/newer version of Qt is passed?

And finally, don't forget about PEP8. Those line lengths look beyond 80 characters wide.

dgovil:
Fair enough. Are you okay with me making a separate test file for these? Defining the classes needed to test those scenarios isn't easy/possible with the current tests.py script.

I'll still have the functions to call these classes in the tests.py module, but I'd need to put the classes for testing in another file.

Would you also be okay with me including a test.ui file rather than embedding the xml in the py file?

Also as to Test 4 in your list for the large UI file, I have no idea how I would even generate a ui file that large, plus I think it would be more of annoyance to anyone checking out this repo to have such a large file solely for testing.

But the others make sense.

Let me know if you're fine with the separate ui/py file and I can get started.

mottosso:
Sure, give it a try.

My concern with multiple files is that it invites complexity. Ideally your tests would be as small and concise as possible, and separate files per ui enables you to write arbitrarily large ui's, the same can be said for tests themselves if they are being organised into separate files.

If you can somehow avoid this trap, then I'd be fine with the splitting. Otherwise I'd prefer sticking to a single file and keeping it small.

dgovil:
How do you feel about taking the improvements/implementation from the `qtpy` project?
https://github.com/spyder-ide/qtpy/blob/master/qtpy/uic.py

Their loadUi implementation is slightly bigger (when looking at just the code, minus the license and comments) but supports more features and is also production tested since `qtpy` is used quite frequently. It also covers a few more edge cases. 

The 3-clause BSD + MIT license covering that portion of the code would not impose anything new on top of your existing MIT license and `qtpy` is already MIT anyway.

bpabel:
I like that their implementation supports the customwidgets argument; though, they don't have tests for it.  And I'm pretty certain it doesn't match the way PyQt's `loadUi` handles custom widgets defined in UI files via the "Promote Widget" feature.

In PyQt, if you had a custom widget at `package.module.MyWidget`, you would create whatever the base of that widget was in Designer and Promote it.  The convention for PyQt was to use the `Header File` field to list the module, but with the old header file semantics, so you would enter it as `package/module.h` with a class name of `MyWidget` and `PyQt4.uic.loadUi` would convert that to `package.module.MyWidget`.  The qtpy implementation looks like it expects the header file field to just have the actual module (i.e. `package.module`) and doesn't do the header path conversion that PyQt4 does.

bpabel:
I just checked and PyQt4 supports both ways of specifying module path, either with the normal python syntax (`package.module`) or the header file syntax (`package/module.h`)

dgovil:
Thanks for the info @bpabel and looking into how PyQt handles it.

I think I'll try taking their implementation then, even if it isn't 1:1 identical to PyQts implementation. Can always build on top of it I think but this would get Qt.py to parity with qtpy atleast.

mottosso:
All we need now are some tests!

dgovil:
Yup. I should have them by this weekend. Crazy crunch has kept me from the PC much.

dgovil:
So I've added a few tests that I think should cover most of the ground suggested.

> Overwriting; if a widget, signal, property or slot already exists, is it overwritten, augmented, ignored?
So this is a hard one. To test overwriting of anything I need to add a layout to the top level widget. When loading the ui file on, PyQT4/5 will complain there is already a layout. PySide(2) will not however. I've traced this back to a difference in PySide and PyQt, where PyQt cares when you redefine a layout for an existing object but PySide does not.
 
I have the test solely running for PySide and PySide2 since I figure it's still good to test the behavior, but it's a binding difference as far as my tests show.

> Does signals created in Qt Designer work once loaded?

Added a test for this

> What happens when an invalid .ui file is passed?

Added a test for this

> What happens when a really large (2 gb) .ui file is passed?

Not sure how to do this test other than procedurally generating a ui file and  I think it might be overkill.

> What kind of paths are supported? Backslash vs. forward slash, network drive (Windows, Linux, Osx) vs local path.

Is there a good way to test this? AFAIK, travis runs on linux so a windows path would be invalid anyway. That said I've tested on both Linux and Windows and it works fine. I just don't see a good way to write a meaningful test here.

>What happens when an invalid path is passed?

Added a test for this

>What happens when a .ui file from an older/newer version of Qt is passed?

So the ui file doesn't actually store Qt version info. There appears to be a schema version tag, but I can't find documentation or an older version. The schema appears to not have changed in a while as far back as I can tell.

-----------------------

Anyway I know the test coverage isn't great, but I think it covers enough to see if it's working, albeit just barely.

Another change I've made is that all bindings now call to `__loadUi` which then dispatches to the requisite implementation. I think this will make it easier to keep the implementations in sync if someone adds features in the future to bring more parity between bindings.

mottosso:
Thanks @dgovil, that's some stellar work. Having a look as soon as I'm able.

Qt.py
```diff
@@ -839,6 +837,45 @@ def _log(text):
         sys.stdout.write(text + "\n")
 
 
+def __loadUi(fname, baseinstance=None):
+    """
+    Dynamically load a user interface from the given ``uifile``.
```
mottosso: We'll need to move this documentation into the [README](https://github.com/mottosso/Qt.py#documentation) somehow, maybe think about a different layout since there is a quite a bit of it.

Alternatively, link to the PySide2 documentation for it as they should match, and highlight that the `custom` argument is excluded.

Qt.py
```diff
@@ -990,6 +1027,180 @@ def _install():
...
+    # THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    # IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    # FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    # THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    # DEALINGS IN THE SOFTWARE.
+
+    class __UiLoader(Qt._QtUiTools.QUiLoader):
+        """
```
mottosso: Skip the newline here, consult the [Napoleon Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for reference.

Qt.py
```diff
@@ -990,6 +1027,180 @@ def _install():
...
+        interface in a base instance.
+        
+        Unlike :class:`~Qt._QtUiTools.QUiLoader` itself this class does not
+        create a new instance of the top-level widget, but creates the user
+        interface in an existing instance of the top-level class if needed.
+        
+        This mimics the behaviour of :func:`PyQt4.uic.loadUi`.
+        """
+
+        def __init__(self, baseinstance, customWidgets=None):
```
mottosso: Get rid of the `customWidgets` argument, it isn't supported.

tests.py
```diff
@@ -1,5 +1,5 @@
 """Tests that run once"""
-
+import contextlib
```
mottosso: Order the imports by length please. :)

mottosso:
Ok, that's a little verbose, but let's give that a try.

To give you some idea of what I'm seeing, I made this.

untitled project

Qt.py
dgovil: Do you have a link to the PySide2 documentation? I am unable to find it

Qt.py
dgovil: customWidgets is passed in by this.
I should probably move it to a class method instead and then remove the argument

`customWidgets = __get_custom_widgets(fname)`

dgovil:
@mottosso Okay I think that should do it. Sorry for the many commits, was jumping between computers this weekend so had to use git as my sync point.

mottosso:
Don't worry about it, thanks for your patience with these micro issues. I think Hound is using the flake8 linter, you could see those issues in real time if you use it locally. Having a look asap.

dgovil:
What are your settings for flake8? I get more warnings than what Hound is configured to report.

mottosso:
That's ok! I haven't configured Hound at all and am also getting more warnings than it does. I've got all linters I can get my hands on running on full blast over here.

tests.py
```diff
@@ -112,6 +138,90 @@ def test_load_ui_returntype():
...
+    else:
+        raise RuntimeError("Loaded an invalid xml")
+    finally:
+        app.exit()
+
+
+if binding("PySide") or binding("PySide2"):
+    def test_load_ui_overwrite_widget():
+        """Checks to make sure ui file widgets supersede previous ones
+        Unfortunately this test only runs in PySide, because PyQt doesn't
```
mottosso: As funny as it sounds, we need to replicate the PyQt behaviour here.

It's important that what works with PyQt, works with Qt.py is PyQt is really what we are using to guide us. Warts and all. We wouldn't want our users to end up in a situation where "It worked before!" as they were using PySide but switched to PyQt.

mottosso:
I've removed the custom widgets feature.

Have a look at try and replicate the PyQt "bug" if loadUi and we're nearly there.

dgovil:
@mottosso So PyQt raises an odd exception type and behavior that I cannot match exactly. So for now I just intercept the condition at the head of the loadUi function and raise my own RuntimeError.

That should make it consistent between bindings. Let me know if you think I should implement it differently but I figure this is the most bang for the buck.

mottosso:
I think we are a go! Merging shortly, great work!

PR #205, opened 2017-05-24
PyCharm autocomplete support

dgovil:
This lets PyCharm autocomplete based on either the systems available PyQt5 or PySide2 bindings.
If you try and do the same with any Qt4 bindings it doesn't seem to like it, maybe because there are so many overlapping ones, but it seems fine with just these two

Image of Autocomplete

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -45,6 +45,15 @@ import importlib
 
 __version__ = "1.0.0.b4"
 
+def path():
+    print("Foo")
+
+# For Autocomplete support inside of packages like PyCharm.
+# Tricks PyCharm into believing these namespaces are imported
+if False:
+    from PySide2 import QtWidgets, QtGui, QtCore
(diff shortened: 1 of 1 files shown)


Qt.py
```diff
@@ -45,6 +45,12 @@
 
 __version__ = "1.0.0.b4"
 
+# For Autocomplete support inside of packages like PyCharm.
+# Tricks PyCharm into believing these namespaces are imported
+if False:
+    from PySide2 import QtWidgets, QtGui, QtCore
+    from PyQt5 import QtWidgets, QtGui, QtCore
```
dgovil: Of course Hound will fail. Is there a way to make it ignore a section of code?

ZimosTD:
@dgovil Thank you for this! I went looking for this because I'm taking your Maya Udemy course. Can't wait for the next one.

PR #206, opened 2017-05-24
Binding State Enums

dgovil:
This adds flags to `Qt.py` that allow end users to check the binding more easily.
Rather  than comparing strings in every piece of code like so:

```python
if Qt.__binding__ == 'PySide2': doSomething()
```

They can now simply do

```python
if Qt.PYSIDE2: doSomething()
```

This in turn makes for less error prone code since string comparisons have the increased risks of typos not caught till runtime.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1088,10 +1088,10 @@ def _install():
 _install()
 
 # Setup Binding Enum states
-Qt.PYSIDE2 = Qt.__binding__ == 'PySide2'
-Qt.PYQT5 = Qt.__binding__ == 'PyQt5'
-Qt.PYSIDE = Qt.__binding__ == 'PySide'
-Qt.PYQT4 = Qt.__binding__ == 'PyQt4'
+Qt.isPySide2 = lambda: Qt.__binding__ == 'PySide2'
+Qt.isPyQt5 = lambda: Qt.__binding__ == 'PyQt5'
+Qt.isPySide = lambda: Qt.__binding__ == 'PySide'
+Qt.isPyQt4 = lambda: Qt.__binding__ == 'PyQt4'
 
 """Augment QtCompat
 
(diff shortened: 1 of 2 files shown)


mottosso:
Thanks @dgovil, I'm no fan of this particular syntax.

How about this?

```python
if Qt.__binding__ == Qt.PYSIDE2:
  doSomething()
```

Or this?

```python
if Qt.IsPySide2:
  doSomething()

if Qt.ISPYSIDE2:
  doSomething()

if Qt.IS_PYSIDE2:
  doSomething()
```

fredrikaverpil:
Hey this is nice @dgovil - thanks :)

I also think it is better to use prefixes somehow, as this is what Qt does to both functions and constants.

I believe the last time we discussed naming, we opted to follow the type of naming used throughout of Qt/PySide2. And if we were to follow that I would say it should be `Qt.isPySide2()` for a function or `Qt.IsPySide2` for a constant.

dgovil:
So I've renamed them to be prefixed and camel cased them to be in line with the Qt functions.
I think that's cleaner.

fredrikaverpil:
Nice!
This all looks great to me.

@mottosso what do you think?

mottosso:
Sorry to be picky, but why is this a function? To me this has all the hallmarks of a constant, and constants in Qt are CamelCase. https://doc.qt.io/qt-5/qt.html

dgovil:
That's fine. I can switch them to constants. Performance would be minorly
better too. I'll put an update in a couple hours when I'm back to my pc

dgovil:
@mottosso, forgot to update the thread, but they're now using CamelCased
constant values

mottosso:
I think it looks good! Real slick work with the one test for all four bindings. :) Thanks @dgovil!

PR #186, opened 2017-03-11
Implement #185

mottosso:
This implements #185, a loosening of the strictness in such a way that (1) still only PySide2 members are present, but (2) members are not required.

This should enable use in less than ideal environments, such as certain OSs that lack various submodules or submodule members, such as QtOpenGL, or applications that may or may not expose the entirety of a binding, such as the lack of QtHelp under Houdini.

The result should be software that is robust and reliable, with a Qt.py that is more forgiving where needed.

**To test**

```bash
# Via pip
$ pip install git+https://github.com/abstractfactory/Qt.py.git@implement185
$ python -c "import Qt"

# Via git
$ git clone https://github.com/abstractfactory/Qt.py.git
$ cd Qt
$ git checkout implement185
$ python -c "import Qt"
```

Or copy/paste [this file](https://raw.githubusercontent.com/abstractfactory/Qt.py/implement185/Qt.py).

Let me know what you think!

Issue #185, opened 2017-03-08
[beta] Optional submodules


mottosso:
It's discussion time.

At the moment, in the current latest commit of this repository, lies the 1.0.0 beta of Qt.py. In it, we have implemented #152 via #173 which means Qt.py contains a subset of members from PySide2.

For the past few months, I and others have tested this and it have worked well. Up to a point.

The problem lies in the strict requirement on availability. Meaning, for Qt.py to even import, these submodules must exist in the original binding.

```python
__all__ = [
    "QtGui",
    "QtCore",
    "QtWidgets",
    "QtNetwork",
    "QtXml",
    "QtHelp",
]
```

Which in most cases is fine. They do exist, most of the time. But not always.

`QtHelp` does not exist in Houdini 16 (at the time of this writing), `QtXml` does not exist in the `python-pyqt5` package of Ubuntu 14.10 and god knows what else is missing from other packages.

The result is a fragile Qt.py, that "sometimes works, sometimes doesn't" depending on factors often outside of your control.

### Possible solution

What I've been considering is making some submodules required, and some optional. `QtGui`, `QtCore` and `QtWidgets` are a given, whereas others may be optional.

It would break the current mantra "If it works with Qt.py, it works on any binding" in that it may work on a platform with support for, say, `QtXml` but then break on platforms without it. On the other hand it would be more forgiving on less-often-used submodules, such as `QtHelp`.

Another option is to give up on strict enforcement, and keep things are they are. The most forgiving, but also least admitting to potential threats of cross-binding compatibility.

tomwod:
hey.  imo i would prefer optional over strict if it a binary choice.  
how about by default be strict, as is, which is explicit.  but allow optional submodules via environment setting - so I can explicitly acknowledge that something is missing in a context that I'm responsible for?

mottosso:
Thanks @tomwod, that's a good suggestion.

davvid:
I would very much like these sometimes-available module to be exposed by Qt.py.  It would be better to not need any environment variables or configuration whatsoever.

It's a good idea to push the burden of dealing with sometimes-available modules to the app developer rather than the global environment.  Apps can then selectively use features[1] and handle the ImportError app-side when importing the sometimes-available module from Qt.py fails.

[...]

PR #207, opened 2017-05-25
Add site customization of _common_members

MHendricks:
We make use of optional Qt libraries at Blur Studio, Qsci and QtCore.QJsonDocument. It would be nice to be able to opt-into being able to use those modules. Ideally in a way where we don't have to maintain our own fork of Qt.py. It also means that we don't need to define imports of these modules outside of the Qt object.

To make this happen I made Qt.py attempt to import the python module Qt_site_config. If it can import the module it calls Qt_site_config.update_common_members(_common_members) and updates _common_members with the returned value.


fredrikaverpil:
Thanks @MHendricks for this!
Looks like a nice addition to Qt.py to me. I see you've included a test. Well done :)

I'm not too fond of underscores in the module name (`Qt_site_config`), although according to PEP8, this is fine:

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged
>
> - [PEP8#package-and-module-names](https://www.python.org/dev/peps/pep-0008/#package-and-module-names)

However, since we are conforming to CamelCase to make Qt.py blend into Qt land. Could you please expand on any reasoning behind this or if you would mind changing it into e.g. `QtSiteConfig`?

A bit on documentation:

- Let's add some info on this feature to the README
- Let's also add this to the list of examples [here](https://github.com/mottosso/Qt.py/blob/master/examples/README.md)

But other than these minor things, thanks for this nice contribution!

@mottosso what do you think?

mottosso:
Hi @MHendricks,

There was a change made recently, March this year, to enable optional submodules to become part of Qt.py, without causing trouble for those that did not have the implementation available locally.

- https://github.com/mottosso/Qt.py/pull/186

I think this would be the preferable method of adding these members and you are welcome to do so. In practice, anything present in `_common_members` is available via `import Qt` if the submodule is available locally.

fredrikaverpil:
@mottosso so what you're saying is we should just add e.g. `Qsci` and `QtCore.QJsonDocument` directly to Qt.py's `_common_members` itself?

Previously, I think we've only discussed members which we know are included in the more regular type of compilation of Qt. In this case I believe that e.g. `Qsci` and `QtCore.QJsonDocument` require custom compilation settings for these members to be available. Am I correct?

This means these members are most likely not available for the "common user"... and I wonder if these should really be added to the `_common_members`...

mottosso:
> In this case I believe that e.g. Qsci and QtCore.QJsonDocument require custom compilation settings for these members to be available. Am I correct?

That's right, this will still work.

There's reasoning behind how and why in the [PR](https://github.com/mottosso/Qt.py/issues/186) and [original issue](https://github.com/mottosso/Qt.py/issues/185), but in a nutshell, Qt.py should encompass all of PySide2 which should encompass all of Qt in some shape or form. Currently the various distributions of PySide2 vary greatly, with even basic submodules missing, like `QtHelp` missing from Houdini specifically. This would enable us to grow Qt.py whilst still allowing PySide2 to catch up.

That said, I'm not expecting us to add any submodules that aren't related to PySide2.

MHendricks:
I also don't like underscores, so I'm happy to rename the module to QtSiteConfig. I can also make the other changes requested.

> That said, I'm not expecting us to add any submodules that aren't related to PySide2.

Heh, along those lines I need to make use of QVariant from PyQt so we have access to the type enums `PyQt5.QtCore.QVariant.Date` for example. At this point we have some bindings that are PyQt specific and we want to use Qt.py to transition to Maya 2017. While we are able to switch to sip api v2 for QVariant we sill need those type enums. I can add code similar to the examples in #185 in the application to support this, but I was planning to use QtSiteConfig to add it that way.

Do you want to proceed with the direction of this pull request, or should I make a new one that just adds the Qsci and QJsonDocument to the existing _common_members?

mottosso:
Have you considered using those members via `QtCore.QMetaType`?

```python
from PyQt4 import QtCore
asset QtCore.QVariant.Date == QtCore.QMetaType.QDate
```

- Reference - https://doc.qt.io/qt-5/qvariant-obsolete.html

> Do you want to proceed with the direction of this pull request, or should I make a new one that just adds the Qsci and QJsonDocument to the existing _common_members?

I'm not sure yet, on one hand there isn't any harm to existing functionality by introducing this functionality but on the other it may open the floodgates to code not usable by anyone other than the original author if code written with this "extension" in place gets into the wild.

For example, if the developers at your studio are first introduced to Qt.py when extended in this manner are later expecting to write software on their own or to help others write software, then if there is suddenly this "hidden addition" that only the original author knows about it may create more issues than it solves.

I think the ideal scenario was if you could find away to migrate your use of `QVariant` to `QMetaType`, and add other PySide2-native submodules you require to Qt.py.

`QMetaType` currently isn't available in Qt.py nor PySide2, that's in the [process of changing](https://wiki.qt.io/PySide2_Missing_Bindings) and you'd be welcome to add it in which case it would become available to PyQt bindings and to PySide2 once they finish implementing it. Then the question remains whether we should implement backwards compatibility to PySide(1) where it is missing also.

MHendricks:
I talked with our Qt developer and it looks like we should be able to migrate to QMetaType and away from QVariant.

I went ahead and addressed @fredrikaverpil comments in case we end up merging this pull request.

fredrikaverpil:
> I went ahead and addressed @fredrikaverpil comments in case we end up merging this pull request.

Thanks @MHendricks :)

@mottosso you decide which route to take here. Either way is fine by me, really. I just reacted to the fact that these particular modules are _most often_ not part of PySide2 and probably never will be in the most popular DCC apps. On the other hand, if you need to use them you might already be aware of this fact.

On a slightly different note, I like the idea of being able to customize Qt.py without actually having to maintain code changes to the Qt.py file itself. The question here is: would it really be useful to be able to customize e.g. the `_common_members` dictionary?

Perhaps the best way forward would be to just add these members to Qt.py but make inline comments about modules which are not compiled by default?
Or perhaps define a new dictionary called `_optional_members` which could hold e.g. `Qsci`. Then we'd merge the `_common_members` and `_optional_members` dictionaries into a `_members` dictionary which would be used in the end. Or perhaps this is just overthinking this problem?

mottosso:
Hi @MHendricks, I had a look at the documentation you've made and gave the implementation a try, I think it's solid.

I made a few tweaks to the API, namely:

1. `update_common_members` was renamed to `update_members`
2. Argument `_common_members` was renamed to `members`
3. `update_members` no longer needs to return
4. No use of `global` in Qt.py

I've also gone ahead and added `QMetaType` for your convenience and updated part of the documentation and examples.

Let me know what you think.

mottosso:
Was going to add some motivation for the changes but ran out of time; I found some so here goes!

I renamed `update_members` as it contained an implementation detail of Qt.py, coming from the variable `_common_members`. Like with exposing any implementation detail, odds are it'll change at some point which would make the name of this function rather nonsensical. `update_members` is user-facing and logical as it's members of Qt.py.

I removed the return statement, as it didn't seem needed and removing it simplified the method in which the functions needed to be written.

As the function no longer needed to return, it was no longer necessary to assign it in the `Qt.py` file, which meant we didn't require the use of `global`.

Finally, I renamed the private function `_add_site_members()` to `_apply_site_config()` as it not only added but potentially modified and removed members as well and felt this better reflected its true intention.

MHendricks:
I wondered if you could just modify the passed in dict. One of the reasons I left it in there is that it made it clear to understand that the function was modifying _common_members by looking at the code. However, I dislike using global if I can avoid it, so I'm happy with the changes. I went ahead and fixed a few broken links from the rename.

Everything looks good to me.

fredrikaverpil:
Headsup, guys. It seems `QtCore.QMetaType` was for some reason removed in more recent builds of PySide2:

- https://github.com/mottosso/Qt.py/pull/227#issuecomment-323254809
- https://wiki.qt.io/PySide2_Missing_Bindings

fredrikaverpil:
@MHendricks @mottosso Headsup, guys. It seems `QtCore.QMetaType` was for some reason removed in more recent builds of PySide2:

- https://github.com/mottosso/Qt.py/pull/227#issuecomment-323254809
- https://wiki.qt.io/PySide2_Missing_Bindings

PR #209, opened 2017-06-01
Fix #208

### LINKED ISSUE #208: Qt.QtCompat.loadUi was broken in Nuke with Qt.py release 1.0.0b4
We've been running Qt.py 1.0.0b3 in production for some time now without any issues. When I gave PR https://github.com/mottosso/Qt.py/pull/190 a spin, I noticed I got errors in Nuke 9.0v9.

When digging further, it seems that the issue stems from changes implemented in Qt.py 1.0.0b4.

### Error

```
# Result: Traceback (most recent call last):
  File "<string>", line 72, in <module>
  File "<string>", line 69, in __init__
  File "/Users/fredrik/miniconda3/envs/Qtdev_py27/lib/python2.7/site-packages/Qt.py", line 1028, in _loadUi
    Qt.QtCore.QMetaObject.connectSlotsByName(widget)
RuntimeError: Internal C++ object (PySide.QtGui.QMainWindow) already deleted.
```

screen shot 2017-05-31 at 08 20 47

### Reproducible code 

This reproducible code runs fine in e.g. Maya but will crash in Nuke when using Qt.py 1.0.0b4:

#### Run from Nuke script editor

```python
import tempfile
import os
import io

from Qt import QtWidgets
from Qt import QtCompat

TEMPDIR = tempfile.mkdtemp()
tempfile = os.path.join(TEMPDIR, "qmainwindow.ui")

with io.open(tempfile, "w", encoding="utf-8") as f:
    f.write(u"""\
<ui version="4.0">
 <class>MainWindow</class>
 <widget class="QMainWindow" name="MainWindow">
  <property name="geometry">
   <rect>
    <x>0</x>
    <y>0</y>
    <width>223</width>
    <height>140</height>
   </rect>
  </property>
  <property name="windowTitle">
   <string>MainWindow</string>
  </property>
  <widget class="QWidget" name="centralwidget">
    <property name="geometry">
     <rect>
      <x>70</x>
      <y>40</y>
      <width>80</width>
      <height>16</height>
     </rect>
    </property>
    <property name="text">
     <string>PushButton</string>
    </property>
   </widget>
  </widget>
  <widget class="QMenuBar" name="menubar">
   <property name="geometry">
    <rect>
     <x>0</x>
     <y>0</y>
     <width>223</width>
     <height>17</height>
    </rect>
   </property>
  </widget>
  <widget class="QStatusBar" name="statusbar"/>
 </widget>
 <resources/>
 <connections/>
</ui>
""")


class MyWindow(QtWidgets.QMainWindow):
        super(MyWindow, self).__init__(parent)
        self.main_widget = QtCompat.loadUi(tempfile)
        self.setCentralWidget(self.main_widget)

my_window = MyWindow()
my_window.show()
```

### Install the problematic Qt.py version

    pip install Qt.py==1.0.0b4

@mottosso I'm labeling this as a bug. Do you have any ideas on why this may be happening or should I check with Foundry Support?

fredrikaverpil:
This fixes #208.

I tested this with Nuke 9.0v9 (where the issue is hit), Nuke 10.0v6, Nuke10.5v4 and Nuke 11 beta. I also tested this with Maya/PySide2.

After having read about [this (fixed) PySide bug](https://bugreports.qt.io/browse/PYSIDE-10), it seems you can avoid the garbage collection by calling [`widget.parentWidget()`](http://doc.qt.io/qt-5/qwidget.html#parentWidget). 🤔

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -897,7 +897,7 @@ def _loadUi(uifile, baseinstance=None):
                     self, uifile, *args, **kwargs)
 
                 # Prevent RuntimeError in Nuke 9.0v9 with PySide 1.0.9
-                parent = widget.parentWidget()
+                widget.parentWidget()
 
                 return widget
 
--- a/Qt.py
+++ b/Qt.py
@@ -897,7 +897,7 @@ def _loadUi(uifile, baseinstance=None):
                     self, uifile, *args, **kwargs)
 
                 # Prevent RuntimeError in Nuke 9.0v9 with PySide 1.0.9
-                widget.parentWidget()
+                parent = widget.parentWidget()
 
                 return widget
 
--- a/Qt.py
+++ b/Qt.py
@@ -893,8 +893,13 @@ def _loadUi(uifile, baseinstance=None):
                 etree = ElementTree()
                 etree.parse(uifile)
 
-                return Qt._QtUiTools.QUiLoader.load(
+                widget = Qt._QtUiTools.QUiLoader.load(
                     self, uifile, *args, **kwargs)
+
+                # Prevent RuntimeError in Nuke 9.0v9 with PySide 1.0.9
+                parent = widget.parentWidget()
+
+                return widget
 
             def createWidget(self, class_name, parent=None, name=""):
                 """Called for each widget defined in ui file
--- a/Qt.py
+++ b/Qt.py
@@ -896,7 +896,7 @@ def _loadUi(uifile, baseinstance=None):
                 widget = Qt._QtUiTools.QUiLoader.load(
                     self, uifile, *args, **kwargs)
 
-                # Workaround for PySide 1.0.9, see issue #209
+                # Workaround for PySide 1.0.9, see issue #208
                 widget.parentWidget()
 
                 return widget
--- a/Qt.py
+++ b/Qt.py
@@ -897,7 +897,7 @@ def _loadUi(uifile, baseinstance=None):
                     self, uifile, *args, **kwargs)
 
                 # Prevent RuntimeError in Nuke 9.0v9 with PySide 1.0.9
-                parent = widget.parentWidget()
+                Qt.QtCompat._loadUi_temp = widget.parentWidget()
 
                 return widget
 
@@ -935,6 +935,7 @@ def _loadUi(uifile, baseinstance=None):
 
         widget = _UiLoader(baseinstance).load(uifile)
         Qt.QtCore.QMetaObject.connectSlotsByName(widget)
+        del Qt.QtCompat._loadUi_temp
 
         return widget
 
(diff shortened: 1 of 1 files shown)


fredrikaverpil:
@mottosso perhaps the object name `Qt.QtCompat._loadUi_temp` is too ambiguous?

Qt.py
```diff
@@ -893,9 +893,14 @@ def load(self, uifile, *args, **kwargs):
                 etree = ElementTree()
                 etree.parse(uifile)
 
-                return Qt._QtUiTools.QUiLoader.load(
+                widget = Qt._QtUiTools.QUiLoader.load(
                     self, uifile, *args, **kwargs)
 
+                # Prevent RuntimeError in Nuke 9.0v9 with PySide 1.0.9
+                Qt.QtCompat._loadUi_temp = widget.parentWidget()
```
mottosso: That's no good, I think it means your window will stick around forever.

fredrikaverpil: Hm. Even if I delete it [here](https://github.com/fredrikaverpil/Qt.py/blob/d07a1e998261ad8006d8a8d52ef9533f2c96cbca/Qt.py#L938)?

    del Qt.QtCompat._loadUi_temp

mottosso: Ah, didn't catch that. No, that should take care of it.

In that case I'm less keen on having Nuke specific code in Qt.py. If we could find a way of reproducing the bug in isolation and resolve that, that would be ideal.

fredrikaverpil: This is most likely not Nuke-specific. I even got the workaround from the [PySide bug tracker](https://bugreports.qt.io/browse/PYSIDE-10).

If we were to repro this outside of Nuke we'd better try it in PySide 1.0.9. Not sure how I could most easily install that old version. Perhaps the only solution is to compile it from scratch:
https://github.com/pyside/PySide/releases/tag/1.0.9

:dizzy_face:

Any other ideas?

mottosso: Ok, in that case something like this should do the trick.

```python
                # Workaround for PySide 1.0.9, see issue #209
                Qt.QtCompat._loadUi_temp = widget.parentWidget()
```

Are we sure we need to store the variable, could we not just call `parentWidget()`? If not, could we store it in a local variable, e.g. `_workaround = ...`

What I'm looking for is our future selves not reading this and wonder what's what.

fredrikaverpil:
> What I'm looking for is our future selves not reading this and wonder what's what.

Exactly. :)

> could we not just call `parentWidget()`?

I was sure just calling `parentWidget()` broke the fix, but now I see it all works fine.

fredrikaverpil:
Ok, that's it. I'm done.

Issue #151, opened 2016-10-24
Obsolete QPixmap attributes


fredrikaverpil:
`QtGui.QPixmap.grabWindow` -> `QtGui.QScreen.grabWindow`
`QtGui.QPixmap.grabWidget` -> `QtWidgets.QWidget.grab`

It seems `grabWindow` can be remapped just fine. But I don't think `grabWidget` can be remapped to `grab` as it takes different arguments.

More info: http://doc.qt.io/qt-5/qpixmap-obsolete.html

On a related note, it doesn't seem like `QtGui.QScreen` has been implemented in PySide2 yet.

MHendricks:
I think QtCompat is going to grow a lot if we add more support for obsolete members, and there are other issues like changed return values that will need to be addressed.

Also, if we keep adding methods to QtCompat its going to be hard to keep track of what objects a function is for. For example `QtCompat.setSectionResizeMode` is for QHeaderView, but you have to read the Qt.py docs or look at the code to figure that out. 

Do we want to define a more strict naming convention? `QtCompat.QHeaderViewSetSectionResizeMode`, or possibly add a namespace object like QtCompat `QtCompat.QHeaderView.setSectionResizeMode`

<br>

I'm working on porting existing PyQt4 code to Qt.py so we can use the code in both Qt4 and Qt5 and have ran into several obsolete method names and different return values. I'm sure I will run into others as I continue as the links below show many obsolete methods.

These fit with the existing QtCompat workaround:
`QtGui.QMouseEvent.angleDelta` -> `QtGui.QMouseEvent.delta`
`QtGui.QHeaderView.setMovable` -> `QtWidgets.QHeaderView.setSectionsMovable`

However, `QtWidgets.QFileDialog`'s `getOpenFileName`, `getOpenFileNames`, and `getSaveFileName` for PyQt4 return `selectedFilename`, while PyQt5, PySide, and PySide2 all return `(selectedFilename, selectedFilter)`. For now, I've made compatibility functions that ensure that PyQt4 returns a empty string for the selected filter and added it to a module I have access to. This probably can go into QtCompat as only PyQt4 needs patched.

Docs:
http://doc.qt.io/qt-5/qwheelevent-obsolete.html#delta
http://doc.qt.io/qt-5/qheaderview-obsolete.html#setMovable

mottosso:
> possibly add a namespace object like `QtCompat.QHeaderView.setSectionResizeMode`

I think that's a good idea. +1

[...]

PR #210, opened 2017-06-19
QtCompat namespacing

MHendricks:
#151 

I've added a framework for adding a class namespace for QtCompat functions. This allows gives you a clear indication of what type of QObject you should be calling this function on.

## Naming
These objects should have a unique name so they don't conflict with real Qt objects, but the name should make it easy to identify what type QObject they are intended to work on. Currently I'm adding a t after the Q so `QFileDialog` becomes `QtFileDialog`, and `QHeaderView` becomes `QtHeaderView`. Some other ideas I have are 'QFileDialogCompat', 'QFileDialogC', 'CFileDialog', 'CompatQFileDialog', but I'm open to any suggestions.

## Implementation
I am using type to create the classes dynamically. For example this is how I'm creating the PySide implementation of the QtHeaderView class. It has 4 methods.
```python
attrs = dict(
    sectionResizeMode = Qt._QtWidgets.QHeaderView.resizeMode,
    setSectionResizeMode = Qt._QtWidgets.QHeaderView.setResizeMode,
    sectionsMovable = Qt._QtWidgets.QHeaderView.isMovable,
    setSectionsMovable = Qt._QtWidgets.QHeaderView.setMovable
)
# type(name, bases, dict)
Qt.QtCompat.QtHeaderView = type('QtHeaderView', tuple([_QtCompat]), attrs)
```
You can use these like the existing QtCompat functions by passing in the QHeaderView as the first argument `QtHeaderView.setSectionsMovable(header, False)`. However, because of the _QtCompat implementation you can also create a instance of QtHeaderView by passing in the header so you don't need to pass header into every function.
```python
headerComp = QtHeaderView(header)
headerComp.setSectionsMovable(False)
```
I think this can be useful especially if you have to call several compatibility functions on a item. However, I have found a couple of drawbacks. For static functions like QFileDialog.getOpenFileName you have to wrap the function in staticmethod or when you call this function on a instance it will pass the arguments wrong.

```python
attrs = dict(
    getOpenFileName = staticmethod(Qt._QtWidgets.QFileDialog.getOpenFileName),
    getOpenFileNames = staticmethod(Qt._QtWidgets.QFileDialog.getOpenFileNames),
    getSaveFileName = staticmethod(Qt._QtWidgets.QFileDialog.getSaveFileName),
)
Qt.QtCompat.QtFileDialog = type('QtFileDialog', tuple([_QtCompat]), attrs)
```
### Alternative
If we don't want to include the instancing feature, we can just pass object instead of _QtCompat.
```python
Qt.QtCompat.QtFileDialog = type('QtFileDialog', tuple([object]), attrs)
```

### DIFF
--- a/examples/QtSiteConfig/main.py
+++ b/examples/QtSiteConfig/main.py
@@ -33,10 +33,15 @@ def test():
     win.setWindowTitle(title)
 
     # Verify that our simple remapping of QWidget.windowTitle works
-    assert QtCompat.QWidget.windowTitleTest(win) == title
+    assert QtCompat.QWidget.windowTitleTest(win) == title, \
+        "Non-decorated function was added to QtCompat.QWidget"
     # Verify that our decorated remapping of QWidget.windowTitle works
     check = 'Test: {}'.format(title)
-    assert QtCompat.QWidget.windowTitleDecorator(win) == check
+    assert QtCompat.QWidget.windowTitleDecorator(win) == check, \
+        "Decorated function was added to QtCompat.QWidget"
+
+    # Suppress 'app' imported but unused warning
+    app
 
 
 if __name__ == '__main__':
--- a/examples/QtSiteConfig/main.py
+++ b/examples/QtSiteConfig/main.py
@@ -25,6 +25,19 @@ def test():
         # Suppress 'Qt.QtCore' imported but unused warning
         QtCore
 
+    # Test _compatibility_members is applied correctly
+    title = 'Test Widget'
+    from Qt import QtWidgets, QtCompat
+    app = QtWidgets.QApplication(sys.argv)
+    win = QtWidgets.QWidget()
+    win.setWindowTitle(title)
+
+    # Verify that our simple remapping of QWidget.windowTitle works
+    assert QtCompat.QWidget.windowTitleTest(win) == title
+    # Verify that our decorated remapping of QWidget.windowTitle works
+    check = 'Test: {}'.format(title)
+    assert QtCompat.QWidget.windowTitleDecorator(win) == check
+
 
 if __name__ == '__main__':
     test()
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -264,7 +264,7 @@ Use compatibility wrapper.
 >>> app = QtWidgets.QApplication(sys.argv)
 >>> view = QtWidgets.QTreeWidget()
 >>> header = view.header()
->>> QtCompat.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
+>>> QtCompat.QHeaderView.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
 ```
 
 Or a conditional.
@@ -281,6 +281,10 @@ Or a conditional.
 ...   header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
 ```
 
+Note: Qt.QtCompat.setSectionResizeMode is a older way this was handled and has been left in for now, but this will likely be removed in the future.
+
+<br>
+<br>
 
 #### QtWidgets.qApp
 
--- a/Qt.py
+++ b/Qt.py
@@ -671,7 +671,8 @@ _compatibility_members = {
     "pyside2": {
         "QHeaderView": {
             "sectionsClickable": "_QtWidgets.QHeaderView.sectionsClickable",
-            "setSectionsClickable": "_QtWidgets.QHeaderView.setSectionsClickable",
+            "setSectionsClickable":
+                "_QtWidgets.QHeaderView.setSectionsClickable",
             "sectionResizeMode": "_QtWidgets.QHeaderView.sectionResizeMode",
             "setSectionResizeMode":
                 "_QtWidgets.QHeaderView.setSectionResizeMode",
@@ -687,7 +688,8 @@ _compatibility_members = {
     "pyqt5": {
         "QHeaderView": {
             "sectionsClickable": "_QtWidgets.QHeaderView.sectionsClickable",
-            "setSectionsClickable": "_QtWidgets.QHeaderView.setSectionsClickable",
+            "setSectionsClickable":
+                "_QtWidgets.QHeaderView.setSectionsClickable",
             "sectionResizeMode": "_QtWidgets.QHeaderView.sectionResizeMode",
             "setSectionResizeMode":
                 "_QtWidgets.QHeaderView.setSectionResizeMode",
--- a/Qt.py
+++ b/Qt.py
@@ -654,6 +654,76 @@ _misplaced_members = {
     }
 }
 
+""" Compatibility Members
+
+This dictionary is used to build Qt.QtCompat objects that provide a consistent
+interface for obsolete members, and differences in binding return values.
+
+{
+    "binding": {
+        "classname": {
+            "targetname": "binding_namespace",
+        }
+    }
+}
+"""
+_compatibility_members = {
+    "pyside2": {
+        "QHeaderView": {
+            'sectionResizeMode': '_QtWidgets.QHeaderView.sectionResizeMode',
+            'setSectionResizeMode': \
+                '_QtWidgets.QHeaderView.setSectionResizeMode',
+            'sectionsMovable': '_QtWidgets.QHeaderView.sectionsMovable',
+            'setSectionsMovable': '_QtWidgets.QHeaderView.setSectionsMovable',
+        },
+        "QFileDialog": {
+            'getOpenFileName': '_QtWidgets.QFileDialog.getOpenFileName',
+            'getOpenFileNames': '_QtWidgets.QFileDialog.getOpenFileNames',
+            'getSaveFileName': '_QtWidgets.QFileDialog.getSaveFileName',
+        },
+    },
+    "pyqt5": {
+        "QHeaderView": {
+            'sectionResizeMode': '_QtWidgets.QHeaderView.sectionResizeMode',
+            'setSectionResizeMode': \
+                '_QtWidgets.QHeaderView.setSectionResizeMode',
+            'sectionsMovable': '_QtWidgets.QHeaderView.sectionsMovable',
+            'setSectionsMovable': '_QtWidgets.QHeaderView.setSectionsMovable',
+        },
+        "QFileDialog": {
+            'getOpenFileName': '_QtWidgets.QFileDialog.getOpenFileName',
+            'getOpenFileNames': '_QtWidgets.QFileDialog.getOpenFileNames',
+            'getSaveFileName': '_QtWidgets.QFileDialog.getSaveFileName',
+        },
+    },
+    "pyside": {
+        "QHeaderView": {
+            'sectionResizeMode': '_QtWidgets.QHeaderView.resizeMode',
+            'setSectionResizeMode': '_QtWidgets.QHeaderView.setResizeMode',
+            'sectionsMovable': '_QtWidgets.QHeaderView.isMovable',
+            'setSectionsMovable': '_QtWidgets.QHeaderView.setMovable',
+        },
+        "QFileDialog": {
+            'getOpenFileName': '_QtWidgets.QFileDialog.getOpenFileName',
+            'getOpenFileNames': '_QtWidgets.QFileDialog.getOpenFileNames',
+            'getSaveFileName': '_QtWidgets.QFileDialog.getSaveFileName',
+        },
+    },
+    "pyqt4": {
+        "QHeaderView": {
+            'sectionResizeMode': '_QtWidgets.QHeaderView.resizeMode',
+            'setSectionResizeMode': '_QtWidgets.QHeaderView.setResizeMode',
+            'sectionsMovable': '_QtWidgets.QHeaderView.isMovable',
+            'setSectionsMovable': '_QtWidgets.QHeaderView.setMovable',
+        },
+        "QFileDialog": {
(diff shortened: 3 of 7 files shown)


MHendricks:
By the way, this is not ready for merging, it's just to get feedback on the concept before I continue working on it.

mottosso:
Thanks for this @MHendricks, I'll be having a closer look today but overall I think it looks fab!

mottosso:
Sorry about the delay on getting into this, pushing a deadline at the moment and will try and have a look this weekend if not early next week!

MHendricks:
No problem, I'm probably not going to get a chance to work on this any next week anyway.

I think the code will need updated to work with #212, but I definitely think 212 will help address the issue with code duplication between all the bindings that the current pull request seems to lead to.

mottosso:
Hi @MHendricks, I'm having a look at this now.

> These objects should have a unique name so they don't conflict with real Qt objects

What was your original thinking behind this?

Hosted under a different namespace, I'd imagine giving them their original name should be fine, and introduce less of a surprise to users and to those reading it afterwards.

Unless of course you prefer importing into the global namespace.

```python
from Qt.QtWidgets import *
from Qt.QtCompat import *
```

I'd discourage this in general, but in this case it actually works out quite well, presuming the user wishes to use our wrappers in place of the originals. They would simply get overwritten with the compatibility wrappers here.

> However, because of the _QtCompat implementation you can also create a instance of QtHeaderView by passing in the header so you don't need to pass header into every function.

That's a handy feature. But I wonder whether it isn't a little bit too complex, both for us to maintain and for the user to memorise.

How about we complete this in a few stages.

1. We stick to the current implementation of mapping only wrapped members to the compatibility class, but move it into a namespace each, e.g. `QtCompat.QHeaderView`
2. Once all wrapped members are there, we can have a look at including non-wrapped members too, such that the `QtCompat` members can be used entirely in-place of their originals.

MHendricks:
First of all, #212 is great, I'm going to use that method to solve my issues with code duplication between the bindings.

I'm working on updating my pull request with the latest code changes.

> in this case it actually works out quite well, presuming the user wishes to use our wrappers in place of the originals. They would simply get overwritten with the compatibility wrappers here.

The _QtCompat objects currently don't reproduce all functionality of the original classes, so `from Qt.QtCompat import QHeaderView` would make it hard to access the original functionality of the QHeaderView class or make a new instance of the class. This may not be a problem if you are using a .ui file, but if you want to create a widget in code it would be much harder.

```python
from Qt.QtCore import Qt
from Qt.QtWidgets import QHeaderView
from Qt import QtCompat
h = QHeaderView(Qt.Horizontal)
QtCompat.QHeaderView.setSectionsMovable(False)
```
If someone(aka me) insists on not using namespaces, this seems reasonable and makes it clear what you are doing. I'm going to update my pull request to use the same name as the original Qt class. That way we don't have to enforce a naming convention. It also makes it compatible with using _QtCompat objects as drop in replacements for the original non-Qt.py class.

>> However, because of the _QtCompat implementation you can also create a instance of QtHeaderView by passing in the header so you don't need to pass header into every function.

> That's a handy feature. But I wonder whether it isn't a little bit too complex, both for us to maintain and for the user to memorise.

Until we make the QtCompat objects a drop in replacement for the original Qt class I think we can drop this feature. So far I know that staticmethods cause problems, and the only reason I know that is because I had a use case to test it with QFileDialog.

While I'm going to remove the instancing feature for now, I think I'm going to keep the _QtCompat class as a baseclass without any custom code. This way if you get into a odd case where you need to use isclassof(cls, _QtCompat) you can do that.

mottosso:
Thanks @mhenricks, diving in tomorrow.

MHendricks:
I've re-done the QtCompat implementation to work with the latest code and in a manner similar to #212 with the changes discussed above. 

http://doc.qt.io/qt-5/obsoleteclasses.html I haven't added all the bindings yet as I would like to get feedback on the implementation so far.

1. My biggest issue with the current code is that there will be a lot of duplication between PySide2/PyQt5 and PySide/PyQt4 in _misplaced_members. All the ways I can think of to reduce the duplication will cause `_build_compatibility_members()` to become more complicated, and probably make it harder for users to track down what is going on when modifying _misplaced_members.

2. Do we need to worry about adding these modules to the _none binding? I  have some test code setup and commented out, but it requires some modifications to `_build_compatibility_members()` to work so it can deal with the original methods not existing.

MHendricks:
Posting this code here for future reference in case we end up wanting to re-add the instantiate functionality of the _QtCompat class.

```python
class _QtCompat(object):
    """ Baseclass for QtCompat namespace objects.

    This class is used as a baseclass for compatibility wrappers used to
    provide a consistent interface across all bindings. For example QFileDialog
    has several staticmethods that have different return values in PyQt4. It is
    also used to handle obsolete methods that were renamed between Qt4 and Qt5.

    Usage:
        header = Qt.QtWidgets.QHeader(...)
        # You can use this class without instantiation by passing the QObject
        # as the first argument.
        ret = Qt.QtCompat.QtHeaderView.sectionsMovable(header)

        # You can also instantiate the class with the QObject to avoid having
        # to pass it every time.
        cmpHeader = Qt.QtCompat.QtHeaderView(header)
        ret = cmpHeader.sectionsMovable()
    """
    def __init__(self, inst):
        data = type(self).__dict__
        # Bind all callables so they use inst as self instead of self.
        # This allows the user to use this class like any other
        for key in data:
            # Note: When defining subclasses, any C++ staticmethods need to
            # to be wrapped in the the staticmethod decorator.
            # See PyQt5 bindings for QFileDialog
            if callable(data[key]):
                setattr(self, key, types.MethodType(data[key], inst))
```

mottosso:
Thanks @MHendricks.

Sorry about this, but I'm not liking the complexity of this PR. I think it solves a narrow problem with too much overhead. The problem being a concern about the growing size of `QtCompat` when taking more members into consideration.

Another thing that worries me is the potential for obscure and difficult to debug errors in the end-user code when using these automatically generated classes. What I like about the current approach is that the `QHeaderView` members of `QtCompat` is mere aliases of the original. It leaves very little room for additional issues to appear.

I'll need some additional feedback on this one. @fredrikaverpil and @dgovil could I get your thoughts on this?

MHendricks:
> Another thing that worries me is the potential for obscure and difficult to debug errors in the end-user code when using these automatically generated classes. What I like about the current approach is that the QHeaderView members of QtCompat is mere aliases of the original. It leaves very little room for additional issues to appear.

https://github.com/blurstudio/Qt.py/commit/009e252755baf29ad176b15980c418a63ce3959f Here is a quick stab at hard coding all the classes instead of defining them dynamically. I also removed the _QtCompat class. We can add it later if we need it.

> The problem being a concern about the growing size of QtCompat when taking more members into consideration.

Yeah, if we include all of the obsolete members its going to get very long. If we wanted to make it complete I don't see a way around it. Perhaps once we define the structure of adding namespaces, we only add them as needed in separate pull requests. It still has the potential to grow very large over time.

MHendricks:
One advantage to the dictionary driven system is that it can be modified using the QtSiteConfig system with a little modification. That would allow users to add the items they need without forcing us to implement all of the obsolete members, just the most common or weird ones like QFileDialog. 
I guess that could be accomplished with a hard coded approach by adding a QtSiteConfig function call after the binding code is run that would allow users to add namespace classes to Qt.QtCompat.

dgovil:
Personally I like the dictionary based method. It has essentially the same chance of issues as the hardcoded system. I believe it also doesn't preclude having hardcoded classes for ones that specifically need them. Plus, like Mike mentions, it would allow for easier extensibility.

What module would this be under? QtCompat? I think as long as the original class is still available under QtWidgets without modification, I don't think there's much harm in extending them under QtCompat. It would be up to the discretion of the developers which they'd use, but I can see it being quite convenient.

MHendricks:
> What module would this be under? QtCompat? I think as long as the original class is still available under QtWidgets without modification, I don't think there's much harm in extending them under QtCompat. It would be up to the discretion of the developers which they'd use, but I can see it being quite convenient.

Yeah, these compatibility modules exist under QtCompat and do not modify the original classes. The QFileDialog implementation for PyQt4 uses a custom decorator function that calls the original binding then changes the returned value to match the other bindings(makes it a tuple of two strings). Otherwise these simply remap the existing functions to a consistent name.

MHendricks:
I've been looking at some of the  obsolete members. Classes like QRect, QColor, QDate, QIcon, QBitmap, QTreeView, QImageWriter, QImageIOHandler all are listed, but their suggested replacements all exist in PySide and PyQt4, so for them I don't think we should provide compatibility classes.

Other things like QImage.setAlphaChannel seem like they would require too much implementation to make a compatibility wrapper.

At this point we have a defined framework for adding more compatibility classes for special cases. Adding compatibility classes can be added as they are found necessary in separate pull requests.

mottosso:
I think it looks good. I'll try and take a closer look within a day or two.

fredrikaverpil:
@MHendricks I just read through your initial post in this PR and also went through the current code. The current code looks elegant to me 👍  and I like how the dictionary falls in line with the other member dictionaries, allowing for greater readability and grokability overall.

My editor highlighted [the `cls` variable](https://github.com/blurstudio/Qt.py/blob/6a4bc3868741bbcc11edb684b3d24052db96f5b6/Qt.py#L886) which surprised me a bit since I actually haven't seen that before. When I checked up I see that it's mentioned in [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments). I don't think you're using this "reserved" keyword "correctly" so may I just be a nitpick and suggest we just call this variable something else, like `compat_class` perhaps?

**Question:** should we allow [QSiteConfig](https://github.com/mottosso/Qt.py/tree/master/examples/QtSiteConfig) to modify the `_compatibility_members` and – while we are at it – the `_misplaced_members` dictionaries?
Right now, it is only allowing to modify the `_common_members` dict ([here](https://github.com/blurstudio/Qt.py/blob/6a4bc3868741bbcc11edb684b3d24052db96f5b6/Qt.py#L747)).
Since this facility is already in place, and the dicts are what they are I see no reason why we shouldn't include these. Meaning;

```python
QtSiteConfig.update_members(_misplaced_members)
QtSiteConfig.update_members(_compatibility_members)
```

MHendricks:
I'll update cls to compat_class.

I think adding QtSiteConfig support for _compatibility_members is a good idea. I think we need to modify the values passed into update_members if we are going to use a single function so its easy to know what type of modifications you need to make. Something like:

```python
QtSiteConfig.update_members(_common_members, 'common')
QtSiteConfig.update_members(_misplaced_members, 'misplaced')
QtSiteConfig.update_members(_compatibility_members, 'compatibility')
```

Qt._build_compatibility_members would require being able to update the list of decorators passed to it. Otherwise you wouldn't be able to handle what I had to do for QFileDialog in PyQt4.

`QtSiteConfig.update_compatibility_decorators(binding, decorators)` could be called at the start of `Qt._build_compatibility_members`.

MHendricks:
Implementation committed, I will try to get tests, examples and documentation updated tomorrow.

I changed the signature of `QtSiteConfig.update_members` to take a second step argument. I did not make this change backwards compatible to keep the code simple, but if you think its necessary, I can add support for it.

Also, I changed the binding keys used in _misplaced_members and _compatibility_members to match the same case as the original module names for consistency when needing to reference them in QtSiteConfig.update_members.

fredrikaverpil:
In the case of #218 it seems that `Qt.QtCompat.loadUi` doesn't exist, perhaps due to that the `PySide.QtUiTools.QUiLoader` is missing to the PySide2 distribution in Houdini (?).

So, [this line](https://github.com/blurstudio/Qt.py/blob/39d157c0d084325a943226568ef370d9549331b0/Qt.py#L1402) would cause issues.

Perhaps we should address this in this PR as well?

MHendricks:
I'll add a `hasattr(Qt.QtCompat, 'loadUi')` check for [that line](https://github.com/blurstudio/Qt.py/blob/39d157c0d084325a943226568ef370d9549331b0/Qt.py#L1402). The binding specific code already handles if `Qt.QtCompat.loadUi` is created.

MHendricks:
Documentation and unit tests have been added.

You can now modify `_common_members`, `_misplaced_members` and `_compatibility_members` using `QtSiteConfig.update_members`.

You can modify the decorators used when building QtCompat namespace classes using `QtSiteConfig.update_compatibility_decorators`. This function is optional and you can omit it if you don't need to modify it.

I've also updated the decoration dictionary to allow you to provide multiple compatibility methods while using the same Qt method. I did this largely to support my unit test, but I can see it being useful in some cases.

examples/QtSiteConfig/main.py
```diff
@@ -25,6 +25,19 @@ def test():
         # Suppress 'Qt.QtCore' imported but unused warning
         QtCore
 
+    # Test _compatibility_members is applied correctly
+    title = 'Test Widget'
+    from Qt import QtWidgets, QtCompat
+    app = QtWidgets.QApplication(sys.argv)
```
fredrikaverpil: You can also supress it like this, I presume:

    _ = QtWidgets.QApplication(sys.argv)

fredrikaverpil:
This all looks great to me. Good job! 👍

mottosso:
Hi guys, sorry for my tardiness on this. You know the story; busy, no time, solar eclipse, yadda yadda.

I say let's try this out. I've incremented the version so we can install it and test things in production. Great job @MHendricks, especially on taking the time to include documentation, something so easily overlooked.

Issue #211, opened 2017-06-21
Binding specific replacement mappings as a dictionary, similar to _common_members


Ahuge:
This is a feature request.

I would love if all of the per binding tomfoolery that is done in the "_pyside", "_pyqt4", etc functions was global as a constant mapping dictionary. 
This would allow other packages relying on Qt.py to know the extent of changes being made.

I have a PR incoming for this.

mottosso:
> tomfoolery

Hah! I learnt a new expression today. :)

Thanks @Ahuge, as discussed this would be a fantastic enhancement to the Qt.py project.

Ahuge:
I have a part two with this that I would like to implement.

Adding support for reassignments to custom functions, for example:

{
    "pyqt4": {
        "uic.loadUi": "QtCompat.loadUi"
    }
}

The problem being that QtCompat.loadUi maps to the custom _loadUi function.


I will create a branch and PR tonight but my thought is that we could support the following "mapping"

{
    "pyqt4": {
        "uic.loadUi": ["QtCompat.loadUi", _loadUi]
    }
}

Where if the value is a list|tuple, we are saying that we are actually replacing the functionality of <key> with <value[1]> and you call it with <value[0]>

So it's more of a replacement than a moved mapping.


I'd love to hear your thoughts on this

mottosso:
There's a similar request [here](https://github.com/mottosso/Qt.py/pull/210), with a similar method of achieving results.

My thoughts in that thread is that the added benefits might not be worth the added complexity. I think what make it more easy to judge is if we could turn things around, and work top-to-bottom - looking first at the feature you are looking to support with these smaller enhancements. That way we'd have some way of justifying the necessary complexity.

What do you think about making a PR with (failing) tests representing the final product you are looking to implement?

Ahuge:
This was solved in #285

PR #212, opened 2017-06-21
Binding specific replacement mappings as a dictionary, similar to _common_members.

Ahuge:
This is a PR that aims to solve #211.

You can see in my first commit that I had been silently ignoring the missing modules, or setting dummy ones. Let me know if that is desired functionality, I doubt it is though.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -840,7 +840,7 @@ def _pyside2():
         Qt.QtCompat.loadUi = _loadUi
 
     if hasattr(Qt, "_QtCore"):
-        Qt.__binding__ = Qt._QtCore.qVersion()
+        Qt.__qt_version__ = Qt._QtCore.qVersion()
 
     # Replacements using _common_replacements
     _set_common_replacements("pyside2")
--- a/Qt.py
+++ b/Qt.py
@@ -605,8 +605,6 @@ _common_replacements = {
             "QtCore.QStringListModel",
         "_QtWidgets.QHeaderView.setSectionResizeMode":
             "QtCompat.setSectionResizeMode",
-        "_QtCore.qVersion":
-            "__qt_version__",
         "_QtCore.QCoreApplication.translate":
             "QtCompat.translate",
         "_QtCore.Property":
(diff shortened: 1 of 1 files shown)


Ahuge:
Ack Travis has failed. Digging into now

Ahuge:
Setting up docker locally for myself. I don't want to deal with installing all the bindings manually if I don't have to. 

I might not get to solving all the failed tests right now, might have to be in 18 hours or so.

mottosso:
> I might not get to solving all the failed tests right now, might have to be in 18 hours or so.

Take your time @Ahug. Thanks for the effort.

Ahuge:
Updated it but even on 67710d40265b9382d7abfc5b43b1822b199bf289 there appears to be 6 failing cases in my local docker environment.

I am pretty sure there is no test case regression on my part.

Ahuge:
Had a small misstyping in my PySide2 bindings. Should be all good now, waiting on Travis

mottosso:
Thanks @Ahuge! Having a look.

mottosso:
About the `_setattr` and friends, I feel that add quite a bit of complexity to the benefit we're reaping here. 

Do you think we could get away with a plain `setattr` if we didn't have to deal with `QHeaderView.setResizeMode`? It looks like the dynamic nature of these functions could be reduced if we could assert that each replacement is accessing exactly one member of one submodule, e.g. `QHeaderView` of `QtWidgets`.

Ahuge:
Yeah I agree about the complexity,

I will spin up a commit just going the one level deep in the morning and move setResizeMode back into the binding function.

Thanks

mottosso:
I gave it a go, let me know what you think.

About QHeaderView and other more complex reassignments, there's a [sibling PR](https://github.com/mottosso/Qt.py/pull/210) about namespacing those members. I think that could go well with this simplification!

Ahuge:
Cool yeah that looks great! Thanks for doing that. 

The sibling PR looks neat as well, I think it could help with the reassignments too. 

Thanks for being patient with me and for all your help

fredrikaverpil:
I've been away for some time and are now looking through what I've missed... I just wanted to chime in and say this looks very neat. Very nicely done, @Ahuge @mottosso !

PR #190, opened 2017-03-29
Implement wrapInstance and getCppPointer (#59)

mottosso:
This adds a wrapper for `wrapInstance` and `getCppPointer` from `shiboken2` and automatically unifies the differences with `shiboken` and `sip` for both Python 2 and 3.

| Attribute               | Returns    
|:------------------------|:-----------
| `QtCompat.wrapInstance(addr=long, type=QObject)` | `QObject` |
 `QtCompat.getCppPointer(object=QObject)`    | `long`    

- See #53 for details.
- To **test**, copy/paste [this file](https://raw.githubusercontent.com/abstractfactory/Qt.py/implement59/Qt.py)

**Usage**

```python
import sys
from Qt import QtCompat, QtWidgets
app = QtWidgets.QApplication(sys.argv)
button = QtWidgets.QPushButton("Hello world")
pointer = QtCompat.getCppPointer(button)
widget = QtCompat.wrapInstance(long(pointer), QtWidgets.QWidget)
assert widget == button
app.exit()
```

**Maya Example**

This works for both 2016 and 2017.

```python
import sys
from maya import OpenMayaUI
from Qt import QtCompat, QtWidgets
pointer = OpenMayaUI.MQtUtil.mainWindow()
widget = QtCompat.wrapInstance(long(pointer), QtWidgets.QWidget)
assert isinstance(widget, QtWidgets.QWidget)
```

### Important

This addition requires `sip`, `shiboken` or `shiboken2` to be available on your system. If not found, Qt.py will still import successfully, but these members will not be available.

In such cases, here is a Qt-only version and guaranteed cross-compatible version of the above.

```python
from Qt import QtWidgets
app = QtWidgets.QApplication.instance()
widget = {o.objectName(): o for o in app.topLevelWidgets()}["MayaWindow"]
```

The same pattern may be applied to any and all uses of `sip`, `shiboken` and `shiboken2`, as discussed in-depth at #53.

Enjoy!

PR #214, opened 2017-06-26
wrapInstance and getCppPointer 

dgovil:
Doing a merge of https://github.com/mottosso/Qt.py/pull/190 to the latest version of `Qt.py`

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -9,6 +9,7 @@ There are cases where Qt.py is not handling incompatibility issues.
 - [QtGui.QRegExpValidator](#qtguiqregexpvalidator)
 - [QtWidgets.QHeaderView.setResizeMode](#qtwidgetsqheaderviewsetresizemode)
 - [QtWidgets.qApp](#qtwidgetsqapp)
+- [QtCompat.wrapInstance](#qtcompatwrapinstance)
 
 <br>
 <br>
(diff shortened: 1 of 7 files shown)


dgovil:
I am unable to resolve these hound errors. The long name is being tripped up because long doesn't exist in Python3 and the indentation errors are because of the sources use of lambdas. I could unindent but I feel that looks worse?

dgovil:
Lesson learned, Py2 has a different scoping system than Py3. Declaring long inside a function causes py2 to error. Has to be declared in the global scope.

Anyway tests are now passing. 

As for the remaining issues:
* Hound: I don't know how to address these indentation errors without affecting the line length or making it more unreadable. Thoughts?

* The license/cla test says I haven't signed it but when I click details it says I have (and I have for previous PR's so not sure why its failing)

mottosso:
Hey @dgovil, there's something not right about this merge, it appears to merge another version than the latest. 

[Here's](https://github.com/abstractfactory/Qt.py/blob/ec26a563cccc291fe8ebc1d6002f14fb6bd2ab5f/Qt.py#L786) what the latest look like, where the wrapInstance line length issue has already been addressed, and [here's](https://github.com/dgovil/Qt.py/blob/7f0272f0cfb859ae956ed8f6588bf1322b27fa89/Qt.py#L752) what this merge looks like.

I also realised that I thought this was originally your PR, when it was actually I who made it. Sorry for asking you to do this merge for me, how lazy of me!

I finished up the original PR now and will release in a few mins.

dgovil:
Odd not sure how it picked up the older version since I pulled latest. But glad you merged the original! thanks!

PR #216, opened 2017-07-12
Tbttfox load ui

tbttfox:
The default empty layout object `bool(QMainWindow().layout())` evaluates to `True` in PySide and `False` in PyQt4.  So I added a check to see if an existing layout is empty.


tbttfox:
Sorry, Git noob here. Somehow, I got my accounts mixed up (hence the failed cli), and trying to clean things up made it worse. Just gonna close this and try again.

PR #217, opened 2017-07-13
Ignore the default empty layout objects when loading a .ui file

tbttfox:
Lets try this again, this time without the phantom user:

The default empty layout object bool(QMainWindow().layout()) evaluates to True in PySide and False in PyQt4. So I added a check to see if an existing layout is empty.


tbttfox:
Ok, I'm getting a Travis build failure on the PyQt implementations. From what I can track down, calling `QMainWindowLayout::count()` emits the warning `qWarning("QMainWindowLayout::count: ?");`, and that warning is what's causing the failure.  However, I haven't the foggiest idea how to either fix it, or have the test ignore that warning.

mottosso:
Thanks @tbttfox.

I've tracked down the original commit from which this clause got added, I think it'd be helpful to future readers if we included a comment on what it's doing there. 

- https://github.com/mottosso/Qt.py/commit/e4a736ab1d9ae214f4caaca20147ed3c83d2925c#diff-cd1d7c0a982bce680146d5b248151256R861

The reason being to mimic PyQt throwing an exception. Which now that I think of it sounds a little backwards. It's PyQt who's supposed to mimic PySide2.

Do you think it's possible to have PyQt *support* this feature, as opposed to PySide inheriting shortcomings of PyQt? It might not be, in which case this would be fine.

tbttfox:
From what I can see, PyQt does support this feature. Nosepipe is just confused by the qwarning (that doesn't actually break anything), and I'm not familiar enough with nosepipe to tell it that the warning is expected, because it's not a normal python error.

And FYI:
My test passes all cases if I remove the layout existence check (the other tests fail, of course). So I'm guessing the act of checking for a layout on a new QMainWindow instance creates the layout. PyQt4 then emits that qwarning when the ui is loaded.

MHendricks:
From my testing for both PyQt and PySide:
- QMainWindow().layout() returns a QLayout with a count of 0.
- QDialog().layout() returns None.
- QDockWidget().layout() returns a QLayout with a count of 2. It stores the close and float buttons used internally.

And if you call bool(widget.layout()) or if widget.layout():
- QMainWindow returns **False in PyQt, and True for PySide**
- QDialog returns False in both PyQt and PySide
- QDockWidget returns **True in both PyQt and PySide**

This is consistent in both Qt4 and and Qt5 versions.

This means that due to the artificial check, you can't use `Qt.QtCompat.loadUi` on QDockWidget at all, and **only in PySide/2 you cant use loadUI on QMainWindow.**

I've done some testing and **PyQt4/5 do not error** if you use `PyQt5.uic.loadUI` on a Widget that already has a layout assigned to it.  I used the code from the unit test to verify that. I only have access to PySide2 from inside Maya so I haven't verified this in PySide.

Qt's c++ code for QLayout raises a qWarning 'QLayout: Attempting to add QLayout "" to QMainWindow "", which already has a layout' I'm assuming this goes to stderr, and nose detects it as a error. 

This is a c++ output, so I'm not sure if you can redirect its stderr output away from nose, but I think this is the original reason for raising the RuntimeError in `Qt.QtCompat.loadUi`. Just so the unittests didn't error out.

Here is a simple example of the problem, in this case it was run with PyQt4 from a windows command prompt:
```python
>>> import sys
>>> from Qt import QtWidgets
>>> app = QtWidgets.QApplication(sys.argv)
>>> win = QtWidgets.QMainWindow()
>>> layout = QtWidgets.QVBoxLayout(win)
QLayout: Attempting to add QLayout "" to QMainWindow "", which already has a layout
>>>
```
No error was raised, but "QLayout: Attempting ..." is written to the command prompt, not python. I haven't seen it in Maya, but I assume its there just not being shown.

Is it possible to make nose ignore stderr from c++ for these specific tests? Or make Qt not log qWarnings? I don't think the check added in e4a736a is needed at all and it actually breaks things.

MHendricks:
I did a little more searching and I think we could control the qWarning output using this function already exposed in Qt.py

http://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler

Some example code that looks promising: https://stackoverflow.com/a/25681472

mottosso:
Thanks for the excellent investigatory notes, @MHendricks.

> I don't think the check added in e4a736a is needed at all and it actually breaks things.

Would you be able to take the lead on resolving that? I don't use `loadUi` myself so it's hard for me to find out about these kinds of problems.

MHendricks:
It may be a few days before I can get to this (production support gets in the way). You don't need to test with loadUi, just attempt to create a layout and pass a parent widget(that already has a layout) into its constructor like I did in the code sample above.

I'll try to find some time to give it a try, or I can help @tbttfox debug it if he has time.

tbttfox:
Ok, so @MHendricks and I got together and got this working, but there's some extra stuff we found out

In the Qt4 bindings, the message handler is called [`qInstallMsgHander`](http://doc.qt.io/qt-4.8/qtglobal.html#qInstallMsgHandler)
In the Qt5 bindings, it's called [`qInstallMessageHandler`](http://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler)
In both cases, they take exactly 1 function as an argument... however, that function requires a different signature between 4 and 5 (Two args in Qt4, vs. Three args in Qt5)

So, for now, we've added `qInstallMessageHandler` as a misplaced member, and it's up to the implementation to handle the different signatures. For the tests we just grab `*args` and parse from there.

There is probably a bigger discussion to be had on how to handle the message handlers, but I think we can do that in a separate PR.

fredrikaverpil:
I'm sorry for the silence here. I'm very busy at work and my wife is expecting in about a week... so I'm not sure exactly when I'll be able to review this.

tbttfox:
I pulled some commits from @MHendricks.
The layout existence check in `_loadUi` was removed. Having that check was causing failures on Linux where the code would run just fine otherwise. The test checking for this was removed.

Tests for running loadUi on `QMainWindow` `QDialog` and `QDockWidget` were added.

MHendricks:
There is some inconsistency between the windows and Linux builds of Qt4 when dealing with QDockWidgets.

```python
>>> from Qt.QtWidgets import QDockWidget
>>> dw = QDockWidget()
>>> dw.layout().count()
2
>>> dw.layout().isEmpty()
True
```
For QDockWidgets on Windows, isEmpty() returns True, but on Linux it returns False. I'm not sure why the layout existence check in _loadUi was added in the first place so I removed it.

QMainWindow has a default .layout() object and return False when .layout().isEmpty() is called on a new instance. This seems consistent for both Linux and Windows.

QDialog and QWidget do not have a default layout.

Also, thanks @tbttfox for cleaning up the unfinished mess I left on my test repo.

mottosso:
> For QDockWidgets on Windows, isEmpty() returns True, but on Linux it returns False.

Maybe we could add a `QtCompat.isEmpty(my_widget)` to account for the differences?

> I'm not sure why the layout existence check in _loadUi was added in the first place so I removed it.

If you select a line here on GitHub and press the "Blame" button, it'll tell you the commit message for that line. In this case, I found:

> Added exception to PySide loadUi if there is an existing layout to mimic PyQt4 behavior
 [reference](https://github.com/mottosso/Qt.py/blame/51778140f6aa590a295280976df737b4dac94615/Qt.py#L882)

Not sure what this means given the changes in this PR. @dgovil do you remember what issues you where having before that commit? Does this resolve those issues?

> QtCore.qInstallMessageHandler

Not sure how this slipped in here, there hasn't been any mention of this so far? I'm happy to keep it though, it looks well thought out and documented.

dgovil:
@mottosso It seems PyQt4 would error if you tried to load a ui file on to a widget that already had a layout assigned, which was the reason for that warning. 

However I've recently discovered that it really depends on the PyQt version whether that is an issue or not. Some versions do not care while others do.

I'm not sure what's the best way to proceed here? IIRC the version that the tests download will fail, but the version we have at work does not. Not sure what's entirely different and noone seems to have answered me

tbttfox:
So I added a test to see that error. And it wasn't an actual error, it was just nosepipe freaking out about a `qWarning` again. I wrapped the test in a message handler, and everything passes!

tbttfox:
After some thought, we decided that `qInstallMessageHandler` should be moved to `QtCompat` because of the handler signatures are different, and the types passed to the handlers are different.

And with that, I think this beast is finally ready, pending review.

mottosso:
Well done everyone. I think it looks great. I've bumped the version and will merge and release once the tests do one final round.

mottosso:
Merging this, sorry for the delay!

PR #225, opened 2017-07-31
Revert "QtCompat namespacing"

mottosso:
I noticed that this breaks backwards compatibility with `QtSiteConfig` and that there are no signs of compatibility not breaking again once we continue to add members to it.

- [Before](https://github.com/mottosso/Qt.py/blob/52e3c073c1baeef2c871a6e546a93458e84590a8/Qt.py#L611)
    `QtSiteConfig.update_members(_common_members)`
- [After](https://github.com/mottosso/Qt.py/blob/2196efcaa9f8a13dab55d38433b4f14f7da6d957/Qt.py#L748)
`QtSiteConfig.update_members(_common_members, 'common')`

In practice, anyone who has used `QtSiteConfig` for anything will have their code broken. That's no good, the last thing I want is for users to fear updating their Qt.py release and for that we must maintain *complete* backwards compatibility until a new major release (i.e. 2.0) is made.

@MHendricks I'd suggest we either remove any alteration to QtSiteConfig, or find a way of altering it in a way that doesn't break backwards compatibility.

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -264,7 +264,7 @@ Use compatibility wrapper.
 >>> app = QtWidgets.QApplication(sys.argv)
 >>> view = QtWidgets.QTreeWidget()
 >>> header = view.header()
->>> QtCompat.QHeaderView.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
+>>> QtCompat.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
 ```
 
 Or a conditional.
@@ -281,10 +281,6 @@ Or a conditional.
 ...   header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
 ```
 
-Note: Qt.QtCompat.setSectionResizeMode is a older way this was handled and has been left in for now, but this will likely be removed in the future.
-
-<br>
-<br>
 
 #### QtWidgets.qApp
 
--- a/Qt.py
+++ b/Qt.py
@@ -43,8 +43,7 @@ import types
 import shutil
 import importlib
 
-
-__version__ = "1.1.0.b3"
+__version__ = "1.1.0.b2"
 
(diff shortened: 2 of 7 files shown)


MHendricks:
What about adding a TypeError check? I think this should work for backwards compatibility.

```python
def _apply_site_config():
    try:
        import QtSiteConfig
    except ImportError:
        # If no QtSiteConfig module found, no modifications
        # to _common_members are needed.
        pass
    else:
        # Update _common_members with any changes made by QtSiteConfig
        try:
            QtSiteConfig.update_members(_common_members, 'common')
        except TypeError:
            # Handle the old method signature so we don't break backwards
            # compatibility with previous versions of QtSiteConfig
            # TODO: add warning.warn
        QtSiteConfig.update_members(_misplaced_members, 'misplaced')
        QtSiteConfig.update_members(_compatibility_members, 'compatibility')
```

MHendricks:
I've added a #226 pull request with a modified version of the code posted above.

fredrikaverpil:
> I noticed that this breaks backwards compatibility with QtSiteConfig

@mottosso whoa, good catch.

Looks like @MHendricks's suggestion could work.

mottosso:
For the sake of new users accidentally pulling the latest backwards incompatible version and to avoid misleading users with the current documentation, I'll revert this for now and then we can re-add it once we've found a fix.

fredrikaverpil:
@MHendricks: Since @mottosso reverted #210, would you mind either creating a new PR with the code from commit 2196efcaa9f8a13dab55d38433b4f14f7da6d957 or add that code to the existing #226 PR and we can work together to make the QtCompat namespacing make it into Qt.py?

MHendricks:
@fredrikaverpil #226 is up to date with the master and the QtCompat Namespace code has been added.

fredrikaverpil:
@MHendricks many thanks!

PR #226, opened 2017-07-31
Make QtSiteConfig.update_members backwards compatible

MHendricks:
This pull request address the problems raised in #225 by adding backwards compatibility.

I've updated the the call to `QtSiteConfig.update_members` to attempt to call the new two argument structure, and if that fails, it logs a message and calls `update_members` passing the single argument.

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -264,7 +264,7 @@ Use compatibility wrapper.
 >>> app = QtWidgets.QApplication(sys.argv)
 >>> view = QtWidgets.QTreeWidget()
 >>> header = view.header()
->>> QtCompat.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
+>>> QtCompat.QHeaderView.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
 ```
 
 Or a conditional.
@@ -281,6 +281,10 @@ Or a conditional.
 ...   header.setSectionResizeMode(QtWidgets.QHeaderView.Fixed)
 ```
 
+Note: Qt.QtCompat.setSectionResizeMode is a older way this was handled and has been left in for now, but this will likely be removed in the future.
+
+<br>
+<br>
 
 #### QtWidgets.qApp
 
--- a/Qt.py
+++ b/Qt.py
@@ -1129,7 +1129,7 @@ def _pyqt4():
 
     decorators = {
         "QFileDialog": {
-            "getOpenFileName":_standardizeQFileDialog,
+            "getOpenFileName": _standardizeQFileDialog,
             "getOpenFileNames": _standardizeQFileDialog,
             "getSaveFileName": _standardizeQFileDialog,
         }
--- a/Qt.py
+++ b/Qt.py
@@ -43,7 +43,8 @@ import types
 import shutil
 import importlib
 
-__version__ = "1.1.0.b2"
+
+__version__ = "1.1.0.b3"
 
 # Enable support for `from Qt import *`
 __all__ = []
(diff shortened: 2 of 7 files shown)


mottosso:
Thanks @MHendricks, the approach is a bit heavy handed I think. What do you think about something like this?

```python
def apply(members):
  members["common"].pop("QtCore")
  members["compatibility"]["PyQt4"]["QFileDialog"] = {
    "getOpenFileName": "QtWidgets.QFileDialog.getOpenFileName",
    "getOpenFileNames": "QtWidgets.QFileDialog.getOpenFileNames",
    "getSaveFileName": "QtWidgets.QFileDialog.getSaveFileName",
  }
```

This would introduce a new function, whilst keeping the old one for compatibility. This function would take a single member where the user is the diving into which "step" to affect. This'd enable us to keep adding to it without breaking compatibility in the future.

Could alternatively make it..

```python
def apply(**members):
```

To communicate that there is a variable number of keyword arguments being applied to it and that the user is responsible for managing changes.

Note that I also removed the `_` from keys as that's really more of an implementation detail and probably shouldn't be exposed to the user. We can add this ourselves during the actual remapping, like we are [with misplaced members](https://github.com/mottosso/Qt.py/blob/2196efcaa9f8a13dab55d38433b4f14f7da6d957/Qt.py#L843).

MHendricks:
So, if QtSiteConfig.apply exists it gets called and should take a arbitrary number of keyword arguments. We can add more keyword arguments as needed.
We would also still call `QtSiteConfig.update_members(_common_members)` though this is optional now and would be removed if we ever go to version 2.

Or, what about adding new optional functions for each new feature we want to add QtSiteConfig control over?

```python
if 'update_members' in QtSiteConfig:
if 'update_misplaced_members' in QtSiteConfig:
	QtSiteConfig.update_misplaced_members(members=_misplaced_members)
if 'update_compatibility_members' in QtSiteConfig:
	QtSiteConfig.update_compatibility_members(members=_compatibility_members)
```

This would give us more flexibility as we find more options we want to add to QtSiteConfig. Even with this approach I think we should recommend using (**kwargs ) when implementing these functions in QtSiteConfig so we have the flexibility to add kwargs to the functions down the road without backwards compatibility issues.

mottosso:
> Or, what about adding new optional functions for each new feature we want to add QtSiteConfig control over?

I think that sounds good, it's just a little verbose. Both to implement and to use.

> Even with this approach I think we should recommend using (**kwargs ) when implementing these functions in QtSiteConfig so we have the flexibility to add kwargs to the functions down the road without backwards compatibility issues.

Let's not. It'll save the day today but I can see it becoming increasingly confusing once one `QtSiteConfig` comes into contact with multiple versions of Qt.py. The only reason I'd opt for `**kwargs` in the case of `apply()` is because interacting with arbitrary keywords would be the only way of interacting with it, so the user would have no choice but to implement safeguards against conflicts into his own code. With specific functions for specific things, arbitrary arguments is a little ambiguous.

Ok, let's try the `QtSiteConfig.update_compatibility_members(members=_compatibility_members)` approach. And do keep `members=` in there. Maintaining backwards compatibility should be a matter of good design, not afterthoughts.

Also take a moment to reflect on whether the data layout of the dictionaries are to your liking. Because we are exposing these via what is essentially an API, we can never modify them again.

MHendricks:
I modified the decorator dictionary to a nested structure similar to the other dictionaries. I think it also is easier to understand and more readable than the old structure I was using. 

```python
    decorators = {
        "QFileDialog": {
            "getOpenFileName": _standardizeQFileDialog,
            "getOpenFileNames": _standardizeQFileDialog,
            "getSaveFileName": _standardizeQFileDialog,
        }
    }
    _build_compatibility_members('PyQt4', decorators)
```

I didn't add `members=` to `QtSiteConfig.update_members(_common_members)` as that was not part of the original QtSiteConfig, but both `QtSiteConfig.update_misplaced_members` and `QtSiteConfig.update_compatibility_members` pass members as a keyword argument.

The "_" requirement has been removed as well.

mottosso:
Ok, let's give this a try!

Personally, I'm no fan of the complexity of this API and expect us to use it as a point of transition into a better suited, alternative API once more members have been implemented using it.

PR #223, opened 2017-07-27
Fix #213

mottosso:
This'll take care of that pesky bug that's been appearing several times over the past month. :)

PR #229, opened 2017-08-09
Repair #223

mottosso:
This was undone in a recent pull request for reasons I cannot fathom.

Thanks @mnemyx for reporting, https://github.com/mottosso/Qt.py/commit/e2e4c423cca32a6d1dbcc56b7965faf142fa1182

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1237,7 +1237,8 @@ def _install():
             setattr(our_submodule, member, their_member)
 
     # Backwards compatibility
-    Qt.QtCompat.load_ui = Qt.QtCompat.loadUi
+    if hasattr(Qt.QtCompat, 'loadUi'):
+        Qt.QtCompat.load_ui = Qt.QtCompat.loadUi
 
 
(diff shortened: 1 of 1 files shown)


mnemyx:
Thank you! Are you planning on keeping this as part of 1.1.0.b2 pre-release? Currently, that tag won't actually include this noted maintenance fix.

mottosso:
Ah I've put myself in a tricky situation; the fix is now made alongside other work-in-progress features and I'd rather not make a release until they are relatively stable.

Are you able to work with this as-is for the time being? I'd expect a week or two for the new changes to settle, but it may take longer. Let me know otherwise and I'll see whether I can pull some git-fu to make this happen.

mottosso:
Ok, the state of the feature I mentioned was quite stable and I've made a new release for it, including this fix.

mnemyx:
Ah thank you again! We were fine with it as-is, but this works even better!

PR #231, opened 2017-08-15
Update setSectionResizeMode section

carlosfelgarcia:
Hi,

There is an error in the solution of the setSectionResizeMode, QtCompat does not have a QHeaderView function.

Cheers,

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -264,7 +264,7 @@ Use compatibility wrapper.
 >>> app = QtWidgets.QApplication(sys.argv)
 >>> view = QtWidgets.QTreeWidget()
 >>> header = view.header()
->>> QtCompat.QHeaderView.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
+>>> QtCompat.setSectionResizeMode(header, QtWidgets.QHeaderView.Fixed)
 ```
 
 Or a conditional.


OEP:
I think this is fairly new functionality (in 1.1.0b3) but they left `QtCompat.setSectionResizeMode` for backwards-compatibility.

fredrikaverpil:
@carlosfelgarcia may I ask which versions of the following you're using:

- Python, Qt.py, binding (such as PySide2) - and if you're using e.g. Maya of a certain version.

carlosfelgarcia:
Hi,
@OEP Yeah you may be right, I'm using 1.1.0b1, although looking at the code of b3 there is no reference of QtCompat supporting QHeaderView as a member.
@fredrikaverpil I'm using with Python 2.7.11, Qt.py 1.1.0b1 and Maya 2017 (PySide2)

mottosso:
@carlosfelgarcia Are you referring to this: https://github.com/mottosso/Qt.py/blob/master/CAVEATS.md#workaround-1

`QHeaderView` is implemented [here](https://github.com/mottosso/Qt.py/blob/adb29579e8e031744d56d3a75284faa7b9e662b6/Qt.py#L673) and should be accessible to you.

```bash
$ python
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] on win32
>>> from Qt import QtCompat
>>> QtCompat.QHeaderView
<class 'Qt.QHeaderView'>
>>> from Qt import __version__
>>> __version__
'1.1.0.b3'
```

carlosfelgarcia:
@mottosso Yes, you are right I can see it now using b3 version. I guess you can reject this pull request. Sorry for the confusion.

mottosso:
Not a problem @carlosfelgarcia, thanks for flagging things you find off!

PR #234, opened 2017-08-18
Testing Qt.py with VFX Platform CY2018

fredrikaverpil:
This is a work in progress and is not ready to merge.

Blocking:

- [x] https://bugreports.qt.io/projects/PYSIDE/issues/PYSIDE-557
- [x] https://github.com/mottosso/Qt.py/pull/240

To do:

- [x] Update `DOCKER.md` with CY2018 specification table
- [x] Update `README.md`
- [x] Update bash scripts to build, run and test using Docker
- [x] Address tests errors


fredrikaverpil:
### Python 2.7 error

```
======================================================================
FAIL: .wrapInstance doesn't need the `base` argument
  File "/usr/local/python2.7/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/usr/local/python2.7/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/workdir/Qt.py/tests.py", line 501, in test_implicit_wrapInstance
    assert widget != button
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
Order: 'PySide2'
Trying PySide2
'QtCore.QMetaType' was missing.
--------------------- >> end captured stdout << ----------------------
```

### Python 3.x error

```
======================================================================
FAIL: Doctest: test_caveats.test_11_qtcompat_wrapinstance
  File "/usr/local/python3.6/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/python3.6/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/usr/local/python3.6/lib/python3.6/doctest.py", line 2199, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for test_11_qtcompat_wrapinstance
  File "/workdir/Qt.py/test_caveats.py", line 175, in test_11_qtcompat_wrapinstance
File "/workdir/Qt.py/test_caveats.py", line 191, in test_11_qtcompat_wrapinstance
Failed example:
    widget == button
Expected:
    False
Got:
    True
```

fredrikaverpil:
Here we go: https://codereview.qt-project.org/gitweb?p=pyside/pyside-setup.git;a=commit;h=8aca0dec116c6b321373f5636b34036530ac6e4b

fredrikaverpil:
Hm. I wonder if we should now be using Autodesk's [`qtdeclarative` fork](https://github.com/autodesk-forks/qtdeclarative/tree/adsk-contrib/vfx/5.6.1) as well. The VFX Platform reference doesn't mention it, but it's sitting there in Autodesk's repository.

I get this when I use the latest commits from Qt5 (5.6 branch) with Autodesk-modified [`qtbase`](https://github.com/autodesk-forks/qtbase/tree/adsk-contrib/vfx/5.6.1) and [`qtx11extras`](https://github.com/autodesk-forks/qtx11extras/tree/adsk-contrib/vfx/5.6.1):

```
/opt/rh/devtoolset-4/root/usr/bin/g++ -c -include .pch/Qt5Quick -pipe -O2 -std=c++1z -fvisibility=hidden -fvisibility-inlines-hidden -fno-exceptions -Wall -
W -D_REENTRANT -fPIC -DQT_NO_MTDEV -DQT_NO_LIBUDEV -DQT_NO_TSLIB -DQT_NO_LIBINPUT -DQT_NO_URL_CAST_FROM_STRING -DQT_NO_INTEGER_EVENT_COORDINATES -DQT_BUILD_
QUICK_LIB -DQT_BUILDING_QT -DQT_NO_CAST_TO_ASCII -DQT_ASCII_CAST_WARNINGS -DQT_MOC_COMPAT -DQT_USE_QSTRINGBUILDER -DQT_DEPRECATED_WARNINGS -DQT_DISABLE_DEPR
ECATED_BEFORE=0x050000 -DQT_NO_EXCEPTIONS -D_LARGEFILE64_SOURCE -D_LARGEFILE_SOURCE -DQT_NO_DEBUG -DQT_GUI_LIB -DQT_QML_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -
DQT_NETWORK_LIB -DQT_CORE_LIB -I. -I. -I../../include -I../../include/QtQuick -I../../include/QtQuick/5.6.3 -I../../include/QtQuick/5.6.3/QtQuick -I/workdir
/qt5/qtbase/include/QtGui/5.6.1 -I/workdir/qt5/qtbase/include/QtGui/5.6.1/QtGui -I../../include/QtQml/5.6.3 -I../../include/QtQml/5.6.3/QtQml -I/workdir/qt5
/qtbase/include/QtCore/5.6.1 -I/workdir/qt5/qtbase/include/QtCore/5.6.1/QtCore -I/workdir/qt5/qtbase/include -I/workdir/qt5/qtbase/include/QtGui -I../../inc
lude/QtQml -I/workdir/qt5/qtbase/include/QtNetwork -I/workdir/qt5/qtbase/include/QtCore -I.moc -isystem /usr/include/libdrm -I/workdir/qt5/qtbase/mkspecs/li
nux-g++ -o .obj/qquicktextinput.o items/qquicktextinput.cpp
items/qquicktextinput.cpp:60:43: fatal error: QtGui/private/qinputcontrol_p.h: No such file or directory
compilation terminated.
make[3]: *** [.obj/qquicktextinput.o] Error 1
make[3]: *** Waiting for unfinished jobs....
items/qquicktextnode.cpp: In member function 'QSGGlyphNode* QQuickTextNode::addGlyphs(const QPointF&, const QGlyphRun&, const QColor&, QQuickText::TextStyle
, const QColor&, QSGNode*)':
items/qquicktextnode.cpp:100:42: error: 'class QFontEngine' has no member named 'isSmoothlyScalable'
             preferNativeGlyphNode = !fe->isSmoothlyScalable;
                                          ^
make[3]: *** [.obj/qquicktextnode.o] Error 1
make[3]: Leaving directory `/workdir/qt5/qtdeclarative/src/quick'
make[2]: *** [sub-quick-make_first-ordered] Error 2
make[2]: Leaving directory `/workdir/qt5/qtdeclarative/src'
make[1]: *** [sub-src-make_first] Error 2
make[1]: Leaving directory `/workdir/qt5/qtdeclarative'
make: *** [module-qtdeclarative-make_first] Error 2
```

fredrikaverpil:
@mottosso so, now I need to do some checking in our tests whether a new PySide2 is used or an older one.

I'm thinking we could add a new function to QtCompat which parses the `PySide2.__build_date__` into a `datetime` format:

```python
import datetime

# Just example code...
def pyside2_build_date():
    try:
        datetime_object = datetime.strptime(
            PySide2.__build_date__[:PySide2.__build_date__.rfind('+')],
            '%Y-%m-%dT%H:%M:%S')
    except (AttributeError):
        # PySide2 version which doesn't have the attribute
        datetime_object = None
    return datetime_object
```

In our tests, we could then do something like:

```python
import datetime

pyside2_build_date = QtCompat. pyside2_build_date()
if pyside2_build_date is None:
    # old version is being used
elif pyside2_build_date <= SOME_DATE:
    # pyside2 is older or equal to SOME_DATE
...
...
```

This way we can utilize date checks against PySide2 builds in tests, caveats more grokable, I think. What do you think?

fredrikaverpil:
I ended up not adding anything to `QtCompat`. Instead I added a `_pyside2_commit_date()` function to `tests.py`. So, inside `tests.py`, we can now query the git commit date of PySide2. Right now, I'm just querying whether this attribute exists or not but in the future we can compare dates more explicitly.

I don't think I can put if conditions in the caveats code snippets and then also put a return value, can I?
Like this I mean...

```python
>>> if hasattr(Qt, '__build_commit_date__'):
...     widget == button
False
>>> else:
...     widget == button
True
```

Anyway, I ended up adding `untested` and a note instead to [the erroring caveat](https://github.com/mottosso/Qt.py/pull/234/files#diff-915339f1306346f3f70f708409018718R323).

Now, we're getting a lot of [membership errors with Python 2.7](https://travis-ci.org/mottosso/Qt.py/jobs/268693884#L620). But all other platforms/combos are green/OK. Not sure how we should proceed here.

build_membership.py
```diff
@@ -147,6 +163,216 @@ def test_{binding}_members():
...
+                "QXmlSimpleReader",
+            ],
+
+            "QtTest": [
+                # missing from PySide
+                "QTest",
+            ],
+        }
+    elif _pyside2_commit_date() <= datetime.datetime(2017, 8, 25):
+        exclude_dict = {}
```
fredrikaverpil: @mottosso okay so I made a fresh dictionary for this new version of PySide2. Not sure if this is how we should do this... please advise. 😀

build_membership.py
```diff
@@ -147,6 +163,216 @@ def test_{binding}_members():
...
+                "QXmlSimpleReader",
+            ],
+
+            "QtTest": [
+                # missing from PySide
+                "QTest",
+            ],
+        }
+    elif _pyside2_commit_date() <= datetime.datetime(2017, 8, 25):
+        excluded_members = {}
```
fredrikaverpil: @mottosso okay so I made a fresh and empty dictionary for this new version of PySide2 which of course causes [lots of members](https://travis-ci.org/mottosso/Qt.py/jobs/268746864#L622) to be reported as missing on all bindings. Not sure if this is how we should do this... please advise. 😀

mottosso:
Hey @fredrikaverpil,

> I'm thinking we could add a new function to QtCompat which parses the PySide2.__build_commit_date__ into a datetime format:

Remind me again why we need conversion on this? The way I see it, it's a version number like any other and would think we'd compare it 1-to-1.

```python
if PySide2.__build_date__ == SOME_DATE:
  ...
```

> Now, we're getting a lot of membership errors with Python 2.7. 

I think we need to take a step back here. These membership tests should guarantee that *all* members included in Qt.py (which is a subset of PySide2) are available in *all* other bindings.

At the moment, that's being tested by (1) extracting *all* members of PySide2 and comparing them to members of other bindings, with some exceptions. That may have worked initially, when PySide2 closely resembled PySide(1) but has since grown to include lots of other things too which is why you're seeing all of these additional exceptions we'd need to exclude.

So I think we need to turn things around. How about we, rather than extract members from PySide2 and compare against that, compare directly against our `_common_members` dictionary? Each of those should exist in all bindings, which is exactly what this test is all about.

That way we (1) don't have to build a membership list, as it's already built which means that (2) we can write the test amongst all other tests and reduce complexity.

fredrikaverpil:
> I'm thinking we could add a new function to QtCompat which parses the PySide2.build_commit_date into a datetime format:
> > Remind me again why we need conversion on this? The way I see it, it's a version number like any other and would think we'd compare it 1-to-1.

We won't need this anymore in `Qt.QtCompat`. I just added this to `tests.py` and `build_membership.py`. We'll need it because PySide2 changes, see [here](https://github.com/fredrikaverpil/Qt.py/blob/d45af8c84b500b0f38a9bc4616a785b2f7597ad8/tests.py#L518).

> So I think we need to turn things around.

Would you mind taking a stab at this, as it sounds like you have an idea on how this could be improved?
If not, I'll look into it later this week, or when possible.

fredrikaverpil:
Note to self: try adding Python 2.6, see what happens.

fredrikaverpil:
Keeping fingers crossed now that we won't get any membership errors...

fredrikaverpil:
Whohooo!! 🎉 💪 

I'll fix the remaining stuff for this PR later tonight (mainly documentation).

fredrikaverpil:
So.. I wonder. Should we stop testing for CY2017?
Those are failing now.

fredrikaverpil:
I'ts because of a newer PySide2 version, which causes new modules to be supported: https://github.com/mottosso/Qt.py/pull/234/commits/80fabd1aa4089695a57bd774cca8dc4234d02758

fredrikaverpil:
Actually, I think it'd be better to keep everything and modify the membership test to exclude the new modules added for each "VFXPLATFORM" version of our tests. That would also be sort of a documentation of what gets added to each iteration of our "testing suite".

fredrikaverpil:
## Common member changes; topics for discussion

This PR is getting ready to be merged. But...

We really haven't talked about how to deal with the fact that PySide2 seems to be picking up speed and is adding new members - and how Qt.py should reflect this change.


### Background

Right now, we have a script in place (`run_membership.sh`) which will generate the common members of the bindings used in our CY2017 Docker test container. The binding versions used in the container doesn't get updated. Instead we add a new container each year (around time of SIGGRAPH) where binding versions may or may not get updated (which is what I'm doing here in this PR). So in this PR I'm adding a more recent version of PySide2 and the membership script detects new members, previously not common across bindings.

So how do we deal with a change of common members?
The way I see it, we have some options depending how we want to play this.


### Clarifying our decisions on this in the project goals

Either we just go with what's the latest and greatest and just include these new common members in Qt.py. Or we do something different. The drawbacks of going with the latest common bindings generated using the latest container is that those members are not available in older versions of PySide2 (which may exist in old versions of e.g. Maya, Nuke etc). This, to me seemingly new, behaviour is not completely clear in e.g. our [project goals](https://github.com/mottosso/Qt.py/blob/51d2b6255da169e3f662f92c714763447c37eb33/README.md#project-goals).

I think it would be multiple orders of difficult to maintain a Qt.py which is somehow aware of where it is running and what limitations in terms of common members it may have. I think we should just go with the latest and greatest. This means slightly tweaking and clarifying this in the project goals, I think.


### Dealing with this in our membership test

Right now, we have a test for members. In this test, I'm excluding members based on which container is running the test. This seems to me to be the most pragmatic way to make this test run. When new members are added to Qt.py, old containers would fail and we'd have to add the new members to a "skip list" in the membership test. Right now, this was the most maintainable approach I could come up with.

### Documenting the member changes

I wouldn't call the "skip list" in the membership test a documentation of common members change ..but it kind of is, for us. I don't think this is worth dragging out from the darkness (of the tests) and into the light (of the main README.md) and present such changes there. Especially not since everyone is probably using different versions of PySide2 (without being able to pinpoint the exact version due to the lack of a version string).

### The need to update the Docker containers more frequently

It's likely that a user is running a later build of PySide2 than we have in our Docker test container, and that they believe a certain member needs to be added to Qt.py. We could potentially build a new Docker image and append this to the current set of test containers. But I'd say we postpone this issue onto the future for when (and if) it happens. I have great hopes that we won't have to update the test containers more often than once a year.


@mottosso what do you think about all of this?

mottosso:
> The need to update the Docker containers more frequently

They won't need any updates, so long as they are an accurate reflection of the versions set by VFX Reference Platform. Users using any other versions are on their own.

At most, we'll need a new image each year. I think that should answer the remaining questions too?

fredrikaverpil:
The problem lies with that PySide2 does not have a maintained version string (any commit or build has version "2.0.0.dev0") and the VFX Platform just specifies "PySide 2.0.x" which basically translates into any version (or commit) of PySide2.

So everyone (including us, vendors) probably just uses the latest source available at the time.

fredrikaverpil:
The problem lies with that PySide2 does not have a maintained version string (PySide2 built from any git commit from the "5.6" branch has version "2.0.0~alpha0") and the VFX Platform just specifies "PySide 2.0.x" which basically translates into any version (or commit) of PySide2.

> Users using any other versions are on their own.

That's everyone, including us. Everyone probably just use the latest source available at the time when they need to build PySide2. We chose to use a build around the time of SIGGRAPH.

The VFX Platform *could* specify a git commit or a git [commit date/hash](https://codereview.qt-project.org/#/c/202199/), but they aren't doing that.

This is why I posted my previous novel-like "topics for discussion"...

fredrikaverpil:
To avoid getting too deep into complex discussions over theoretical scenarios, I've instead updated the README.md and DOCKER.md to attempt to explain the issue at hand.

So, in case you understand what I'm trying to explain I think we are good :) but if you find the updated README (check the diff) hard to grok, then we probably need to clarify this further.

@mottosso To sum it up, it would be great with a review of this PR and add the following to the next release notes:

```
New common members added:
  - QtGui.QDesktopServices
  - QtMultimedia
  - QtOpenGL
```

I think we should then consider releasing a non-beta version of Qt.py at PyPi and in Anaconda cloud.

alby128:
Since you are adding modules, it would also be useful to include QtCore.QStandardPaths in the next release. Thanks.

fredrikaverpil:
@alby128 I don't see QtCore.QStandardPaths in either of the bindings/versions (Qt >5.6). Which binding are you using where this is available and which version? 

Qt.py will only contain built in support for members available throughout all bindings. You can modify this behavior and add members yourself using QtSiteConfig.py. Please refer to the main README for more information.

alby128:
QStandardPaths has been introduced in Qt 5 as (partial) replacement for (some features of) QDesktopService. See the first paragraph here: http://doc.qt.io/qt-5.6/whatsnew50.html . So it is not available on PySide1.x and PyQt4. However, it is available on PyQt5 (see [here](http://pyqt.sourceforge.net/Docs/PyQt5/api/QtCore/qstandardpaths.html?highlight=qstandardpaths##PyQt5-QtCore-QStandardPaths)) and on PySide2 (I checked in the conda binaries). 

I do not know what is your policy for modules that are only available on Qt >= 5.6. Thanks for the tip on how to add a member.

fredrikaverpil:
@alby128 thanks for clarifying. We want Qt.py to support members which exist throughout all the bindings so that software built with Qt.py has greater chances running regardless of binding. We're know this limits the Qt library compatibility but this is also one of the reasons we built Qt.py to begin with; to support both Qt4 and Qt5 bindings (but only members which exists in both).

Of course, your request is not uncommon and that's why we also enabled the possibility to add your own members in case you need to break out of "total compatibility". I hope this works for you (QtSiteConfig.py) as we're don't want to add these specific members to Qt.py as they do not exist throughout all the bindings.

mottosso:
Hey @fredrikaverpil, should we merge this?

fredrikaverpil:
@mottosso yes, we should be fine merging this. Two late changes were made to the CY2018 which isn't part of this docker build:

> - 26th November 2017 - CY2018 has had a late change to the compiler version which is now gcc 6.3.1. This successfully resolves issues that were discovered with gcc 5.3.1.
> - 28th August 2017 - Added a note for gcc 5 and updated the Qt note to include a link to the qtdeclarative modifications.

 I'll try to cover those as soon as I can, but in the meanwhile we're fine merging this as-is.

fredrikaverpil:
I'm re-running the latest checks and when done, feel free to merge!

fredrikaverpil:
All tests OK - merge away!

PS. I noticed that Qt4 won't compile with the late CY2018 gcc version change. This means we'll have to manage two gcc versions for a CY2018-Update 1 image. But for now, let's merge this.

fredrikaverpil:
How about bumping the version to 1.1.0b8 and planning for a 1.1.0 release on pip and anaconda cloud?
Do you think we should finish some other stuff prior to 1.1.0?

mottosso:
Sounds like a good idea. 👍

PR #239, opened 2017-08-23
--convert option: raise an error if the input ui file defined signals

cchampet:
* Currently it does not work: I have to define my signal in my python
code instead of my ui file.
* I did not find a good solution... So for now I think it's better to
warn the user about this unsupported feature than silently tell nothing.
Do you have a solution to support this feature? :)

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1295,6 +1295,11 @@ def _convert(lines):
         line = line.replace("from PySide2 import", "from Qt import QtCompat,")
         line = line.replace("QtWidgets.QApplication.translate",
                             "QtCompat.translate")
+        if "QtCore.SIGNAL" in line:
+            raise NotImplementedError("QtCore.SIGNAL is missing from PyQt5 "
+                                      "and so Qt.py does not support it: you "
+                                      "should avoid defining signals inside "
+                                      "your ui files.")
         return line
 
     parsed = list()


mottosso:
Thanks @cchampet. Is your .ui file generated via Qt 4 or Qt 5? Qt.py only supports .ui files generated with Qt 5.

cchampet:
My .ui file was generated via Qt 4, but I've just generated it again via Qt 5, and I have the same file:
```xml
<ui version="4.0">
...
 <connections>
  <connection>
   <sender>cancelButton</sender>
   <signal>released()</signal>
   <receiver>refererenceImportDialog</receiver>
   <slot>close()</slot>
   <hints>
    <hint type="sourcelabel">
     <x>283</x>
     <y>356</y>
    </hint>
    <hint type="destinationlabel">
     <x>343</x>
     <y>178</y>
    </hint>
   </hints>
  </connection>
 </connections>
</ui>
```
Note: the __ui version__ does not seem to define the Qt version...

mottosso:
Ah, in that case this seems reasonable. @fredrikaverpil do you have any input on this?

fredrikaverpil:
No, unfortunately not. I've only used .ui files by loading them on the fly. I'm not up to speed with the `--convert` functionality.

I'm creating .ui files using Qt Creator 4.1.0, based on Qt 5.7.0 which works great with Qt.py and the `QtCompat.loadUi` function. Never tried the `--convert` option of Qt.py.

mottosso:
Yes, it was me. :) The code itself isn't the problem, but rather whether we should ban the use of declaring signals in Qt Designer, or whether there is a way around it.

I'd say it should be supported, but for now this is a good implementation to let the user know of the problem.

@cchampet if you could have a look at the Hound warnings, I'd be happy to merge this.

cchampet:
Agreed we should ban the use of declaring signals in Qt Designer.
@mottosso I've just pushed force the commit to fix the Hound warnings.

mottosso:
Thanks @cchampet!

PR #242, opened 2017-08-29
Update README with conda info, include LICENSE in tarball

### LINKED ISSUE #202: Upload to Conda
### Goal

Expose Qt.py to users of [Conda](https://conda.io)

### Implementation

See here: https://conda.io/docs/build_tutorials/pkgs.html

fredrikaverpil:
Fixes #202 

Qt.py was today added to conda/Anaconda Cloud/conda-forge and its feedstock lives [here](https://github.com/conda-forge/qt.py-feedstock/blob/master/README.md).

I added the Anaconda Cloud badge to the README and extended the install instructions with a conda example.

What's particularly cool about this, is that PySide2 can be – as of yesterday – _also_ installed with conda:
This makes me very happy:

```bash
$ conda config --add channels conda-forge
$ conda install qt.py pyside2
```

We should make our `LICENSE` be included in the [Qt.py tarball](https://pypi.org/project/Qt.py/#files) instead of keeping a copy of it in [qt.py-feedstock](https://github.com/conda-forge/qt.py-feedstock). I've added `MANIFEST.in` to this PR, which is [supposed to handle this](https://docs.python.org/3/distutils/sourcedist.html#specifying-the-files-to-distribute) but I'm not sure how to test this. We need to trigger a PyPi release (through Travis) to know for sure?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ setup(
     url="https://github.com/mottosso/Qt",
     license="MIT",
     zip_safe=False,
+    data_files=['LICENSE'],
     py_modules=["Qt"],
     classifiers=classifiers
 )
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
     url="https://github.com/mottosso/Qt",
     license="MIT",
     zip_safe=False,
-    data_files=['LICENSE'],
+    data_files=["LICENSE"],
     py_modules=["Qt"],
     classifiers=classifiers
 )
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-include LICENSE
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/mottosso/Qt.py.svg?branch=master)](https://travis-ci.org/mottosso/Qt.py) [![PyPI version](https://badge.fury.io/py/Qt.py.svg)](https://pypi.python.org/pypi/Qt.py)
+[![Anaconda-Server Badge](https://anaconda.org/conda-forge/qt.py/badges/version.svg)](https://anaconda.org/conda-forge/qt.py)
 
 ### Qt.py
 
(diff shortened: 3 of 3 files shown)


mottosso:
> I've added MANIFEST.in to this PR, which is supposed to handle this but I'm not sure how to test this. We need to trigger a PyPi release (through Travis) to know for sure?

That is one way of doing it, and you can test it via `python setup.py sdist` *I think*.

Alternatively, there is a `data_files` argument to `setup.py` you could use to include it, then the above will work for sure and you can inspect the resulting zip file.

fredrikaverpil:
> Alternatively, there is a data_files argument to setup.py

I was thinking of that too. But wouldn't that add `LICENSE` to the wheel and cause it to get dropped inside of end-users `site-packages` directory?

fredrikaverpil:
> python setup.py sdist

That worked!

And `LICENSE` is inside.

fredrikaverpil:
Before `MANIFEST.in`: 

screen shot 2017-08-29 at 15 53 46

And after:

screen shot 2017-08-29 at 15 54 36

fredrikaverpil:
I just ran `python setup.py bdist_wheel` to check the contents of the wheel. And with `MANIFEST.in` in place, there are no additional files added to the wheel (such as the license).

fredrikaverpil:
Ok, I just also used `data_files` in `setup.py` and that resulted in this wheel:

screen shot 2017-08-29 at 16 00 04


...and this tarball:

screen shot 2017-08-29 at 16 00 10

I think that's cleaner. So, I'll vote for adding data_files to setup.py and remove MANIFEST.in.

@mottosso what would you prefer?

fredrikaverpil:
Okay, I'm ready if you are. Pull the trigger on this PR if you're happy!

setup.py
```diff
@@ -31,6 +31,7 @@
     url="https://github.com/mottosso/Qt",
     license="MIT",
     zip_safe=False,
+    data_files=['LICENSE'],
```
mottosso: Nitpick, please use consistent quotation marks `"` :)

Other than that, all good!

fredrikaverpil: Whoops. Good catch. Fixed.

mottosso:
Cool, win. :)

Issue #235, opened 2017-08-18
Python 2.6 Support


Ahuge:
Hey all

I am slightly ignorant as to the benefits of importlib, looking at it's source it looks fairly basic, and much of it's complexity seems more than we need.

Python2.6 does not appear to have importlib and thus Qt.py cannot be used in anything below 2.7

Does anyone have any problem with me creating an implementation of how we are using importlib but only using __import__ instead?

It would be something like this:
```python
def _import_sub_module(module, name):
    module = __import__(module.__name__ + "." +  name)
    for level in name.split("."):
        module = getattr(module, level)
    return module
```

With a little bit of sanity checking in there.

Our studio is planning on updating to python2.7 *soon*, but that has been the plan for a while, so if you don't think you want to add this complexity to support 2.6 I understand.

Thanks

fredrikaverpil:
@Ahuge as long as it doesn't hamper future development and cause Python 3 issues (which I don't think it'll do), I don't see a problem with this.

To me it looks like your code should work exactly as-is, so you could just set up a PR and we'll. take it from there.

mottosso:
As it happens, `importlib` is not a native citizen of Python 2.7 *either*, but is actually (partially) backported from Python 3.

Here is the entire definition of `importlib` in Python 2.7.

**`python27/lib/importlib/__init__.py`**

```python
"""Backport of importlib.import_module from 3.x."""
# While not critical (and in no way guaranteed!), it would be nice to keep this
# code compatible with Python 2.3.
import sys

def _resolve_name(name, package, level):
    """Return the absolute name of the module to be imported."""
    if not hasattr(package, 'rindex'):
        raise ValueError("'package' not set to a string")
    dot = len(package)
    for x in xrange(level, 1, -1):
        try:
            dot = package.rindex('.', 0, dot)
        except ValueError:
            raise ValueError("attempted relative import beyond top-level "
                              "package")
    return "%s.%s" % (package[:dot], name)


def import_module(name, package=None):
    """Import a module.

    The 'package' argument is required when performing a relative import. It
    specifies the package to use as the anchor point from which to resolve the
    relative import to an absolute import.

    """
    if name.startswith('.'):
        if not package:
            raise TypeError("relative imports require the 'package' argument")
        level = 0
        for character in name:
            if character != '.':
                break
            level += 1
        name = _resolve_name(name[level:], package, level)
    __import__(name)
    return sys.modules[name]
```

[...]

PR #245, opened 2017-08-29
python2.6 support

Ahuge:
Working on #235 

Don't merge yet. I haven't had a change to do it all yet or run the unittests yet.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -775,7 +775,7 @@ def _setup(module, extras):
     for name in list(_common_members) + extras:
         try:
             submodule = _import_sub_module(
-                module.__name__, name)
+                module, name)
         except ImportError:
             continue
 
--- a/Qt.py
+++ b/Qt.py
@@ -760,8 +760,8 @@ def _new_module(name):
 
 
 def _import_sub_module(module, name):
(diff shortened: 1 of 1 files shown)


Ahuge:
I don't know Travis well enough to setup 2.6 tests. Do we need/want them?

I think it would be somewhere in the docker setup that I would add support for 2.6 I think.

fredrikaverpil:
> I don't know Travis well enough to setup 2.6 tests. Do we need/want them?

The question is whether all the bindings will build on Python 2.6... I'm not sure. I would have to try that. For this PR, I suggest we skip such a test for now. I can add Python 2.6 support the next time I'm working on the #234 and see what happens.

Ahuge:
Hey so I fixed the houndci issues, I am changing the title of this to remove the "WIP"

mottosso:
Completely missed this one, sorry about that. I think this looks good. @Ahuge would you happen to have tested this (in production)?

Ahuge:
I think I've got a copy of it at work, I'll confirm in a bit

Ahuge:
Yeah confirmed the branch py26-support works in production.

fredrikaverpil:
I'd say let's just merge this then.

PR #257, opened 2017-12-02
Missing q style option view item v4

mangopipeline:
Ok sorry it took me a bit to get this pull request out...
I'm a bit swamped and this is my first time using git to do a pull request...

it also took me a a couple of hours to understand how the QtCompat class was being populated...

I"m not sure this is a good way to do this change...
Currently the uses is 
QtCompat.QStyleOptionViewItemV4.GET()

it might be more elegant to tweak the bindingMethod so that it support mapping classes directly...
it's not often that this is needed, so maybe the .GET method is ok...

maybe it needs to be standardized ?


mottosso:
Thanks @mangopipeline.

Could you provide an example of the expected use of this member? I'm not sure this is a variable, a function or class at the moment.

mangopipeline:
Do you want me to add it to the example module?
i don't really have the time to put it together...

I only use it inside delegates so i would have to put a pretty complex view model delegate example....

Here's the a delegate paint method use it in....

the Qt4 docs show it as a class

http://doc.qt.io/qt-4.8/qstyleoptionviewitem.html
http://doc.qt.io/qt-4.8/qstyleoptionviewitemv4.html

in Qt5  QStyleOptionViewItemV4 has been merged into QStyleOptionViewItem

The strange part is that it's not a simple renaming of the Class both do exist in Qt4.*

one seems to help with drawing frames, and the second is just a bucket for styling as i understand it...

 ```
   def paint(self, painter, option, index):
        options = QtComp.QStyleOptionViewItemV4(option)
        self.initStyleOption(options, index)
        
        if not options.widget:
            style = QtWidgets.QApplication.style()
        else:
            style = options.widget.style()

        data = index.data(role=QtCore.Qt.UserRole)
        document = self.getTextDocument(option, data)

        #no text label rendered.
        options.text = ''

        # Draw the element 
        # options 
        style.drawControl(QtWidgets.QStyle.CE_ItemViewItem, options, painter)

        paint_context = QtGui.QAbstractTextDocumentLayout.PaintContext()

        #mouse over change the highlight color.
        if options.state & QtWidgets.QStyle.State_Selected:
            pColor = options.palette.color(QtGui.QPalette.Active, QtGui.QPalette.HighlightedText)
            paint_context.palette.setColor(QtGui.QPalette.Text,pColor)

        # Get paint rectangle in screen coordinates
        text_rectangle = style.subElementRect(QtWidgets.QStyle.SE_ItemViewItemText, options,options.widget)
        #text_rectangle = options.rect

        # Adjust the rectangle to fix issue occurring on windows.
        text_rectangle.adjust(+5, 0, -5, 0)

        painter.save()
        painter.translate(text_rectangle.topLeft())
        painter.setClipRect(text_rectangle.translated(-text_rectangle.topLeft()))
        document.documentLayout().draw(painter, paint_context)
        painter.restore()

```

**Comment by mottosso** [comment]
To confirm, this does *not* work for you?

```python
def paint(self, painter, option, index):
       options = QtWidgets.QStyleOptionViewItem(option)
```

As I understand it, `QStyleOptionViewItemV4` is just an alias for `QStyleOptionViewItem` and should not be used at all.

- https://doc.qt.io/qt-5/qstyleoptionviewitem-obsolete.html

mangopipeline:
This works in qt5 fine but not qt4.
In Qt4 QStyleOptionViewItemV4 is not an alias for QStyleOptionViewItem they are 2 different classes used in different ways.

This is why QStyleOptionViewItemV4 might need to stay around for backwards compatibility.

Carlos Anguiano

mottosso:
I've added this member to QtCompat [here](https://github.com/mottosso/Qt.py/pull/258), could you let me know if this works for you?

mottosso:
Closed in favour is #258

PR #260, opened 2017-12-11
Testing with CY2018 Update 1

fredrikaverpil:
There was one late change to the CY2018 [VFX Platform](http://www.vfxplatform.com/):

> **26th November 2017** - CY2018 has had a late change to the compiler version which is now gcc 6.3.1. This successfully resolves issues that were discovered with gcc 5.3.1.

Also, `qtdeclarative` was added:

> **28th August 2017** - Added a note for gcc 5 and updated the Qt note to include a link to the qtdeclarative modifications.

This PR aims to address these additions.

### How to's

    # Build new image
    docker build . -t fredrikaverpil/qt.py:2018-update1 -f Dockerfile.vfxplatform2018-update1

    # Run tests locally, Unix
    docker run --rm -v $(pwd):/Qt.py -e PYTHON=2.7 fredrikaverpil/qt.py:2018-update1
    docker run --rm -v $(pwd):/Qt.py -e PYTHON=3.4 fredrikaverpil/qt.py:2018-update1
    docker run --rm -v $(pwd):/Qt.py -e PYTHON=3.5 fredrikaverpil/qt.py:2018-update1
    docker run --rm -v $(pwd):/Qt.py -e PYTHON=3.6 fredrikaverpil/qt.py:2018-update1

    # Run tests locally, Windows
    docker run --rm -v %CD%:/Qt.py -e PYTHON=2.7 fredrikaverpil/qt.py:2018-update1
    docker run --rm -v %CD%:/Qt.py -e PYTHON=3.4 fredrikaverpil/qt.py:2018-update1
    docker run --rm -v %CD%:/Qt.py -e PYTHON=3.5 fredrikaverpil/qt.py:2018-update1
    docker run --rm -v %CD%:/Qt.py -e PYTHON=3.6 fredrikaverpil/qt.py:2018-update1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,24 @@
 matrix:
   include:
 
+  # VFX Platform 2018-update1
+  - os: linux
+    env:
+    - VFXPLATFORM=2018-update1
+    - PYTHON=2.7
+  - os: linux
+    env:
+    - VFXPLATFORM=2018-update1
+    - PYTHON=3.4
+  - os: linux
+    env:
+    - VFXPLATFORM=2018-update1
+    - PYTHON=3.5
+  - os: linux
+    env:
+    - VFXPLATFORM=2018-update1
+    - PYTHON=3.6
+
   # VFX Platform 2018
   - os: linux
     env:
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -95,6 +95,7 @@ You can then run `python2.7`, `python3.4`, `python3.5`, `python3.6` and so on (d
 | Qt5 | 2016-06-16 | [commit](http://code.qt.io/cgit/qt/qt5.git/commit/?h=5.6&id=4566f0ac50e5ea143943c1251028fb01c70289ce) |
 | Adsk Qt5 `qtbase` | 2017-06-07 | [commit](https://github.com/autodesk-forks/qtbase/commit/c4e51d0162f7619c83e25e623ecd3bc549932040) |
 | Adsk Qt5 `qtx11extras` | 2017-02-24 | [commit](https://github.com/autodesk-forks/qtx11extras/commit/c6c59d5d902db8be3661cab929be85a38fda0faa) |
+| Adsk Qt5 `qtdeclarative` | 2017-05-09 | [commit](https://github.com/autodesk-forks/qtdeclarative/commit/b14cd49f891327d8fa5f5607eb19288ca137c5e2) |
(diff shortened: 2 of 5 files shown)


fredrikaverpil:
I've noticed that we cannot just change the gcc version, or Qt4 won't build. Therefore we'll have to maintain two gcc versions side-by-side so that we can build PySide and PyQt4 successfully.

Currently the build of this Docker image will fail because I just switched out the gcc version.

fredrikaverpil:
Better late than never, I guess.

I've now added the CY2018 late change as a separate testing docker container tagged `2018-update1`:

> 26th November 2017 - CY2018 has had a late change to the compiler version which is now gcc 6.3.1. This successfully resolves issues that were discovered with gcc 5.3.1. --  [vfxplatform](http://www.vfxplatform.com/)

### Changes in testing Qt.py from CY2018

To be honest, the changes this update introduced doesn't really mean much for Qt.py, as I've left all commits the same as in CY2018. Therefore tests should run identical to the CY2018 container. The really only change is that we're using a newer build environment for all building except Qt4.

### Issues tackled

The main issues with this change is that Qt4 won't build with the newer gcc 6.3.1 without errors. I tackled this by building Qt4 with gcc 5.31 and everything else with gcc 6.3.1. I wasn't sure if mixing gcc versions like this was going to work, but [tests are completing without issues](https://travis-ci.org/fredrikaverpil/Qt.py/builds/399439105?utm_source=github_status). However, I would like to stress that this is not something I would do in a production environment.

Just to elaborate for my future self when I turn back to this PR to see what the heck I did;
* devtoolset-4 is installed to offer gcc 5.3.1
* make is installed to offer make for devtoolset-4
* devtoolset-6 is installed to offer gcc 6.3.1
* devtoolset-6-make is installed to offer make for devtoolset-6
* When building, always run commands using the appropriate build environment;
  * `scl enable devtoolset-6 'gcc --version'  # gcc 6.3.1`
  * `scl enable devtoolset-4 'gcc --version'  # gcc 5.3.1`
* For qt-creator, the `qbs` submodule URL has been changed and therefore a sed command is added to update to the now active URL

### Should this be merged?

I'm not sure if it adds any value to us. Maybe if we updated the commits of each git repository to be taken from somewhere around this update (November, 2017) rather than using the very same commits as for CY2018, then it would make more sense to merge this.

I primarily wanted to solve this problem and see if it could be done and also start thinking about how to tackle CY2019. So I'm fine with us just closing this if we feel this really just adds complexity without value.

mottosso:
> Not sure what Hound is on about?

Looks like a bug. Next commit should flush it out.

PR #262, opened 2017-12-15
Release 1.1.0

fredrikaverpil:
Let's assemble what should be written in the release notes for 1.1.0.


fredrikaverpil:
@mottosso if I look at [what's been written](https://github.com/mottosso/Qt.py/releases) for all pre-releases since 1.0.0, this is quite a lot, but very informative. I'm thinking we could just copy-paste that in that chronological order and remove the pre-release numbers. What do you say?

fredrikaverpil:
Actually, we should take some time and write down what's clear in this release, as it is a big release. I've started: https://github.com/mottosso/Qt.py/releases/edit/untagged-1d82a4295d8944b66f56

fredrikaverpil:
I just added members which were added since 1.0.0 to the 1.1.0 release draft in https://github.com/mottosso/Qt.py/releases

@mottosso these remain to be documented for the release:

* Changes to QSiteConfig
* Added wrapInstance and getCppPointer
* Custom Attribute Editor Template
* Get top-level window in any binding and any application.

PR #265, opened 2018-01-11
Add from Qt.QtCompat import support

MHendricks:
This pull request adds support for `from Qt.QtCompat import X` so it's consistent with the other Qt modules.


mottosso:
Nice one. :) I can't think of any side effects to doing that, but let me have a think.

mottosso:
Nothing comes to mind, let's give it a go!

Issue #267, opened 2018-01-16
Missing QtPrintSupport


alexwidener:
Hi,

I am unable to find a reason for why this would be missing and am hoping somebody can enlighten me.

```
import Qt
print Qt.__binding__
print Qt.__binding_version__
from Qt import QtPrintSupport

PySide2
2.0.0~alpha0
# Error: ImportError: file <maya console> line 5: No module named QtPrintSupport # 
```

And in the same session:

```
import PySide2
from PySide2 import QtPrintSupport
QtPrintSupport
# Result: <module 'PySide2.QtPrintSupport' from '/[removed]/autodesk/maya/2018/maya/lib/python2.7/site-packages/PySide2/QtPrintSupport.so'> # 
```

I see the same result in PyQt4, where the member also exists.

Best,
Alex

mottosso:
Thanks @alexwidener, looks like a missing member indeed. Welcome to submit a PR with it added to the list in the Qt.py file; the automatic tests will kick in to check that everything is A-OK.

fredrikaverpil:
This seems odd to me. We have a membership comparison script which will compare all members across bindings...

Please let me verify this before merging anything.

alexwidener:
https://travis-ci.org/mottosso/Qt.py/jobs/329962171#L570

Looks like PySide is the problem.

MHendricks:
Yes, the classes in PyQt4/PySide are under QtGui, and they seem to be the same classes. It seems like all you need to do is add the correct values to the `_misplaced_members` dict. PySide in Maya 2016, and PySide2 in Maya 2017 both have the print classes. I haven't looked in other DCC's.

#268 I'm using QtPrintSupport as my example in this pull request, but that's because I noticed the bug updating some widgets that were printing using PyQt4. I agree that if QtPrintSupport can be added as default members of Qt.py it should be added.

alexwidener:
Closed #269 in favor of #268

MHendricks:
@fredrikaverpil unless you object, I'll update #268 to include QtPrintSupport in _common_members by default instead of requiring enabling using QtSiteConfig.

Without that update to `_reassign_misplaced_members` we can't add QtPrintSupport to Qt4 without modifying `_pyqt4` and `_pyside` to add the missing QtPrintSupport modules.

fredrikaverpil:
Hm. I haven't really had time to investigate this but so it kind of seems like Adsk had some custom version of PySide(1), which we don't have in our test suite.

[...]

PR #268, opened 2018-01-16
Misplaced members can't create Qt sub-modules

MHendricks:
Currently you can't use misplaced members to add new Qt modules like `Qt.QtPrintSupport` universally. It works for Qt5 because both PySide2 and PyQt5 have the QtPrintSupport module, but in Qt4, the print classes are stored on QtGui.

This patch gives `_reassign_misplaced_members` the ability to create missing modules like `Qt.QtPrintSupport`.

I modified the logic of how `_reassign_misplaced_members` rejects members a little bit from #212. If the destination method exists, but the source module was not created for Qt, it will create the module.

### DIFF
--- a/examples/QtSiteConfig/QtSiteConfig.py
+++ b/examples/QtSiteConfig/QtSiteConfig.py
@@ -8,12 +8,6 @@ def update_members(members):
     Arguments:
         members (dict): The members considered by Qt.py
     """
-    # By default Qt.py does not include QtPrintSupport, so lets add it.
-    members["QtPrintSupport"] = [
-        "QPrintPreviewWidget",
-        "QPrinter",
-    ]
-
     # Contrived example used for unit testing. This makes the QtCore module
     # not accessible. I chose this so it can be tested everywhere, for a
     # more realistic example see the README
@@ -27,16 +21,6 @@ def update_misplaced_members(members):
     Arguments:
         members (dict): The members considered by Qt.py
     """
-    # Printer support in Qt4 was stored under QtGui instead of QtPrintSupport
-    members["PyQt4"]["QtGui.QPrintPreviewWidget"] = (
-        "QtPrintSupport.QPrintPreviewWidget"
-    )
-    members["PyQt4"]["QtGui.QPrinter"] = "QtPrintSupport.QPrinter"
-    members["PySide"]["QtGui.QPrintPreviewWidget"] = (
-        "QtPrintSupport.QPrintPreviewWidget"
-    )
-    members["PySide"]["QtGui.QPrinter"] = "QtPrintSupport.QPrinter"
-
     # Create Qt.QtGui.QColorTest that points to Qtgui.QColor for unit testing.
     members["PySide2"]["QtGui.QColor"] = "QtGui.QColorTest"
     members["PyQt5"]["QtGui.QColor"] = "QtGui.QColorTest"
--- a/examples/QtSiteConfig/QtSiteConfig.py
+++ b/examples/QtSiteConfig/QtSiteConfig.py
@@ -9,9 +9,9 @@ def update_members(members):
         members (dict): The members considered by Qt.py
     """
     # By default Qt.py does not include QtPrintSupport, so lets add it.
-    members['QtPrintSupport'] = [
-        'QPrintPreviewWidget',
-        'QPrinter',
+    members["QtPrintSupport"] = [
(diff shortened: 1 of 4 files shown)


mottosso:
Mike is on a roll.

@MHendricks, happy to hand you "keys to the castle" if you're up for merging and releasing changes on your own? Code quality and feature proposals aligns with what I'd expect. Deployment to pypi is triggered automatically when making a release. I think it could be good.

MHendricks:
That was unexpected, Thanks. I'll do my best.

examples/QtSiteConfig/QtSiteConfig.py
```diff
@@ -22,6 +27,15 @@ def update_misplaced_members(members):
     Arguments:
         members (dict): The members considered by Qt.py
     """
+    # Printer support in Qt4 was stored under QtGui instead of QtPrintSupport
+    members['PyQt4']['QtGui.QPrintPreviewWidget'] = \
```
mottosso: Could we stick with double-ticks for quotation marks, and parenthesis for line-breaks?

E.g.

```python
members["PyQt4"]["QtGui.QPrintPreviewWidget"] = (
  "QtPrintSupport.QPrintPreviewWidget"
)
```

And then similarly for assertions.

```python
name = QtPrintSupport.QPrintPreviewWidget.__name__
assert name == "QPrintPreviewWidget', (
  "Qt.QtPrintSupport.QPrintPreviewWidget was not correctly mapped"
)
```

mottosso:
> That was unexpected, Thanks. I'll do my best.

Don't mention it, you've earned it. :) Invite sent. My only critique is on the cosmetics I mentioned in my other comment just now.

examples/QtSiteConfig/QtSiteConfig.py
MHendricks: Oops, I try to use double-ticks for Qt.py, but missed those.
I was trying to figure out how to do a multi-line assert as that line was extremely long but assert's don't work as functions. I had not thought of using the parenthesis like that before.

mottosso:
Eek! I'm not too sure about that last one. Did this member really exist in PySide? It's a slippery slope from here. Last thing I want is to use members of Qt.py in my program that would cause my program to break on another binding. Especially if it happened without warning or explicit consent from the developer.

MHendricks:
See #267, All of QtPrintSupport classes are part of all 4 bindings, but in Qt4 the classes are under QtGui and QtPrintSupport does not exist, so I'm using `_misplaced_members` to add the QtPrintSupport module so Qt.py is consistent with PySide2.

It's listed in all 4 bindings in the latest membership log that I could find.
https://gist.github.com/fredrikaverpil/6fb1e6a2273fcd877b4b56c91738d36f#file-run_membership_output-log-L387

[`tests.test_membership()`](https://github.com/mottosso/Qt.py/blob/master/tests.py#L698) I've verified that this will fail if `Qt.QtPrintSupport` does not exist, or does not have all the classes listed in `_common_members["QtPrintSupport"]`. Which would be the case if any of the bindings did not have all of these classes.

I can restore QtPrintSupport to requiring QtSiteConfig if needed, but it seems that its a part of all 4 bindings and could be added by default.

membership.py
```diff
@@ -51,16 +51,20 @@ def compare(dicts):
...
 
+    # When Qt4 was moved to Qt5, they split QtGui into QtGui, QtWidgets, and
+    # QtPrintSupport.
     pyside['QtWidgets'] = pyside['QtGui']
     pyqt4['QtWidgets'] = pyqt4['QtGui']
+    pyside['QtPrintSupport'] = pyside['QtGui']
+    pyqt4['QtPrintSupport'] = pyqt4['QtGui']
 
     write_json(pyside, pyside_filepath)
     print('--> Copied QtGui to QtWidgets for ' + os.path.basename(
```
fredrikaverpil: We should change this print message to also mention `QtPrintSupport`.

MHendricks: Heh, I guess I completely ignored the print statement.

MHendricks: It now outputs:
```
C:\blur\dev\Qt_fork>membership.py --copy-qtgui
--> Wrote PySide.json
--> Copied QtGui to QtWidgets and QtPrintSupport for PySide.json
--> Wrote PyQt4.json
--> Copied QtGui to QtWidgets and QtPrintSupport for PyQt4.json
```

PR #270, opened 2018-01-22
Add Syncplay to the "Projects using Qt.py" section

alby128:
We integrated Qt.py a month ago in version 1.5.1 . Thanks for all this work.


mottosso:
Thanks, and you're welcome!

PR #273, opened 2018-01-30
Adding QWebEngineView

krets:
This gives me enough to port my existing `QWebView` widgets into `QWebEngineView`.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -430,6 +430,11 @@ _common_members = {
     "QtTest": [
         "QTest"
     ],
+    "QtWebEngineWidgets": [
+        "QWebEnginePage",
+        "QWebEngineView",
+        "QWebEngineSettings"
+    ],
     "QtWidgets": [
         "QAbstractButton",
         "QAbstractGraphicsShapeItem",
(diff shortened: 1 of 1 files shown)


krets:
I committed this with the wrong author email. I'll update and re-pull.

krets:
It looks like github managed my --amend.
Please merge if this works.

fredrikaverpil:
Those modules won't be available on all bindings, as you can see from the failing tests.

I would recommend you look into [QtSiteConfig.py](https://github.com/mottosso/Qt.py#qtsiteconfigpy) to add these on your end.

mottosso:
Continuing from the [Maya Python mailing list](https://groups.google.com/forum/#!topic/python_inside_maya/vDqLRKDOSYY).

@krets Thanks for the pull request. I haven't used these web widgets myself, but my impression is that they are too different to be treated as one and the same; one wrapping WebKit and the other Chromium. 

For completeness, the members exposed through Qt.py are those that are *identical* across all bindings, such that you can write an application in any binding and know for sure it will run without problems in all other bindings without changes to the original application.

If you are confident that QWebView qualifies, then I see no problem including it in Qt.py. What I'd need is tests that exercise the most important functionality of QWebView. Functionality that isn't portable should be made an example of as [Caveats](https://github.com/mottosso/Qt.py/blob/master/CAVEATS.md).

Let me know if you have any questions!

fredrikaverpil:
@krets in case Qt decided to move members around between Qt4 and Qt5 in regards to QtWebEngineWidgets, you need to figure out how these were moved/renamed and then all we have to do is to include this here: https://github.com/mottosso/Qt.py/blob/master/membership.py#L54

By executing the `build_membersip.sh` script, a `common_members.json` file will be generated. Then all you need to do is to copy its contents into the `common_members` dictionary in Qt.py.

fredrikaverpil:
So in this case, you should make sure that you can see the following in the generated `common_members.json` file (by adding to the `copy_qtgui_to_modules()` function):

```
    "QtWebEngineWidgets": [
        "QWebEnginePage",
        "QWebEngineView",
        "QWebEngineSettings"
    ],
```

fredrikaverpil:
Edited my previous post for clarity..

PR #277, opened 2018-03-08
Add Compatibility members for #151

Ahuge:
In Qt5 QPixmap.grabWidget got moved to QWidget.grab.

This code moves the member to QtCompat.

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -10,6 +10,7 @@ There are cases where Qt.py is not handling incompatibility issues.
 - [QtWidgets.QHeaderView.setResizeMode](#qtwidgetsqheaderviewsetresizemode)
 - [QtWidgets.qApp](#qtwidgetsqapp)
 - [QtCompat.wrapInstance](#qtcompatwrapinstance)
+- [QtGui.QPixmap.grabWidget](#qtguiqpixmapgrabwidget)
 - [QtCore.qInstallMessageHandler](#qtcoreqinstallmessagehandler)
 
 <br>
@@ -63,7 +64,7 @@ True
 
 I had been using the id as an index into a list. But the unexpected return value from PyQt4 broke it by being invalid. The workaround was to always check that the returned id was between 0 and the max size I expect.  
 
-– @justinfx
+\- @justinfx
 
 
 <br>
@@ -352,3 +353,37 @@ True
 ```
 
 Note the `False` for PySide2 and `True` for PyQt5.
+
+#### QtGui.QPixmap.grabWidget
+
+The method of capturing a widget to a pixmap changed between Qt4 and Qt5.
+
+PySide and PyQt4:
+```python
+# PySide
+>>> from Qt import QtGui, QtWidgets
+>>> app = QtWidgets.QApplication(sys.argv)
+>>> button = QtWidgets.QPushButton("Hello world")
+>>> pixmap = QtGui.QPixmap.grabWidget(button)
+```
+
+PySide2 and PyQt5
+```python
+# PySide2
+>>> from Qt import QtGui, QtWidgets
+>>> app = QtWidgets.QApplication(sys.argv)
+>>> button = QtWidgets.QPushButton("Hello world")
+>>> pixmap = button.grab()
+```
+
+##### Workaround
+
+Use compatibility wrapper.
+
+```python
+# PySide2
+>>> from Qt import QtCompat, QtWidgets
+>>> app = QtWidgets.QApplication(sys.argv)
+>>> button = QtWidgets.QPushButton("Hello world")
+>>> pixmap = QtCompat.QWidget.grab(button)
+```
--- a/Qt.py
+++ b/Qt.py
@@ -43,7 +43,7 @@ import types
 import shutil
 
 
-__version__ = "1.1.0"
+__version__ = "1.2.0.b1"
 
 # Enable support for `from Qt import *`
 __all__ = []
--- a/Qt.py
+++ b/Qt.py
@@ -761,6 +761,9 @@ interface for obsolete members, and differences in binding return values.
 """
 _compatibility_members = {
     "PySide2": {
+        "QWidget": {
+            "grab": "QtWidgets.QWidget.grab",
+        },
         "QHeaderView": {
             "sectionsClickable": "QtWidgets.QHeaderView.sectionsClickable",
             "setSectionsClickable":
@@ -778,6 +781,9 @@ _compatibility_members = {
         },
     },
     "PyQt5": {
+        "QWidget": {
+            "grab": "QtWidgets.QWidget.grab",
+        },
         "QHeaderView": {
             "sectionsClickable": "QtWidgets.QHeaderView.sectionsClickable",
             "setSectionsClickable":
@@ -795,6 +801,9 @@ _compatibility_members = {
         },
     },
     "PySide": {
+        "QPixmap": {
+            "grabWidget": "QtWidgets.QWidget.grab",
+        },
         "QHeaderView": {
             "sectionsClickable": "QtWidgets.QHeaderView.isClickable",
             "setSectionsClickable": "QtWidgets.QHeaderView.setClickable",
@@ -810,6 +819,9 @@ _compatibility_members = {
         },
     },
     "PyQt4": {
+        "QPixmap": {
+            "grabWidget": "QtWidgets.QWidget.grab",
+        },
         "QHeaderView": {
             "sectionsClickable": "QtWidgets.QHeaderView.isClickable",
             "setSectionsClickable": "QtWidgets.QHeaderView.setClickable",
--- a/Qt.py
+++ b/Qt.py
@@ -801,8 +801,8 @@ _compatibility_members = {
         },
     },
     "PySide": {
-        "QPixmap": {
-            "grabWidget": "QtWidgets.QWidget.grab",
+        "QWidget": {
+            "grab": "QtWidgets.QPixmap.grabWidget",
         },
         "QHeaderView": {
             "sectionsClickable": "QtWidgets.QHeaderView.isClickable",
@@ -819,8 +819,8 @@ _compatibility_members = {
         },
     },
     "PyQt4": {
-        "QPixmap": {
-            "grabWidget": "QtWidgets.QWidget.grab",
+        "QWidget": {
+            "grab": "QtWidgets.QPixmap.grabWidget",
         },
         "QHeaderView": {
             "sectionsClickable": "QtWidgets.QHeaderView.isClickable",
--- a/tests.py
+++ b/tests.py
@@ -679,6 +679,11 @@ def test_qtcompat_base_class():
     QtCompat.QHeaderView.setSectionsMovable(header, True)
     assert QtCompat.QHeaderView.sectionsMovable(header) is True
 
+    # Verify that the grab function actually generates a non-null image
+    button = QtWidgets.QPushButton('TestImage')
+    pixmap = QtCompat.QWidget.grab(button)
+    assert not pixmap.isNull()
+
 
 def test_cli():
     """Qt.py is available from the command-line"""


Ahuge:
Any chance we can talk about merging this in?

mottosso:
Of course, sorry about that, had missed this one. Looks straightforward enough, do we also want to version this up for a release, or better wait for more features/changes?

mottosso:
Could we add an example or test for this as well?

MHendricks:
Do we also want to add support for the `QtGui.QPixmap.grabWindow` -> `QtGui.QScreen.grabWindow` option mentioned in the #151 issue? It sounds like its pretty strait forward compatibility member.

When documenting grab, you should probably mention the caveat that you should not use `QPixmap QPixmap::grabWidget(QWidget * widget, int x = 0, int y = 0, int width = -1, int height = -1)` that is available in Qt4. See https://github.com/mottosso/Qt.py/issues/151#issuecomment-371350402.

mottosso:
Ah, doesn't that mean that `grab` is too generic of a name, if we also have `grabWindow`? Should we call it `grabWidget`, despite it being called just `grab` in Qt 5 (albeit on a class calls `QWidget`)?

```python
QtCompat.grabWidget()
QtCompat.grabWindow()
```

MHendricks:
Sorry, I copied the example from before we added QtCompat namespacing. The new way to access it would be:

`Qt.QtCompat.QScreen.grabWindow(....)`
`Qt.QtCompat.QWidget.grab(...)`

`QPixmap.grabWidget` is the Qt4 version of Qt5's `QWidget.grab` command.

mottosso:
Ah, in that case, I think it makes sense to add `grabWindow` too. Would you like to add it @MHendricks? Then I think we'd better version bump to `1.2.0.b1`.

MHendricks:
Sure, I'll work on it tonight.

MHendricks:
Wow, I'm glad I used `Qt.QtCompat.QScreen.grabWindow` in all bindings. I'm not sure if we want to support it as it will require decorators for at least some of the QtCompat namespace bindings among other issues.

All testing has been done in Maya 2016 for Qt4 and Maya 2017 for Qt5 on windows.

- `PySide2.QtGui.QScreen` is not implemented. https://bugreports.qt.io/browse/PYSIDE-434. However, they have included the old obsolete method `PySide2.QtGui.QPixmap.grabWindow`. 
- VFXPLATFORM=2017 tests don't have access to `PySide2.QtGui.QPixmap.grabWindow`. I did a simple assert [test](https://travis-ci.org/blurstudio/Qt.py/builds/353182315) and Qt.py failed to import using PySide2
- `PyQt5.QtGui.QScreen.grabWindow` requires a instance of `QScreen`. See "Testing Code" below.
- `PyQt5.QtGui.QPixmap.grab` and `PyQt5.QtGui.QPixmap.grabWindow` do not exist so we can't just use `QPixmap.grabWindow`.
- PySide2, PySide and PyQt4 both do not require passing a QScreen object.

I think we have four choices:

1. Don't expose `QtCompat.QScreen.grabWindow` at this time.
2. Add a `QtCompat.grabWindow` implementation that works consistently for all bindings, but doesn't strictly conform to PySide2/PyQt5 method signatures.
3. Make `QtCompat.QScreen.grabWindow` conform to PyQt5's signature. `QtCompat.QScreen.grabWindow(screen, winId, ...)`. I think that if PySide2 ever adds support for QScreen this is the best choice, but it would force you to have to pass None, or some other object as the first argument(PyQt5 requires a type of QScreen so None is not the best option). This would require decorators for PySide2, PySide and PyQt4
4. Make `QtCompat.QScreen.grabWindow` conform to `QtGui.QPixmap.grabWindow`. This would require adding a decorator for PyQt5 that automatically gets the primaryScreen object and passes it to the `PyQt5.QtGui.QScreen.grabWindow` call.

I'm leaning towards the second option, but as a separate pull request.

Testing Code:
```python
import os
from maya import OpenMayaUI as omui
mainWindowId = long(omui.MQtUtil.mainWindow())

# Test PySide
os.environ['QT_PREFERRED_BINDING'] = 'PySide2;PySide'
import Qt
reload(Qt)
print Qt.__binding__
from Qt import QtCompat, QtWidgets

mainWindow = QtCompat.wrapInstance(mainWindowId)
label = QtWidgets.QLabel()
pixmap = QtCompat.QScreen.grabWindow(mainWindow.winId())
label.setPixmap(pixmap)
label.show()

# Test PyQt
os.environ['QT_PREFERRED_BINDING'] = 'PyQt5;PyQt4'
reload(Qt)
print Qt.__binding__
from Qt import QtCompat, QtWidgets

mainWindow = QtCompat.wrapInstance(mainWindowId)
label = QtWidgets.QLabel()
if Qt.__binding__ == 'PyQt5':
    # https://github.com/leo-editor/leo-editor/commit/d4e29c9 (Glad I found this)
    screen = QtWidgets.QApplication.instance().primaryScreen()
    pixmap = QtCompat.QScreen.grabWindow(screen, mainWindow.winId())
else:
     pixmap = QtCompat.QScreen.grabWindow(mainWindow.winId())
label.setPixmap(pixmap)
label.show()
```

MHendricks:
I didn't get a chance to look into unit tests, but we could probably set up a test where we make a QWidget and `assert not QtCompat.QWidget.grab(widget).isNull()`. That will test that the function was called and actually returned a QPixmap.

fredrikaverpil:
Sigh... what a mess. Let's assume QScreen will never get implemented in PySide2 for now. I believe too much work is needed in other places before the QtC devs get to this particular issue.

I'm leaning towards this one, at least initially:
> 2. Add a QtCompat.grabWindow implementation that works consistently for all bindings, but doesn't strictly conform to PySide2/PyQt5 method signatures.

mottosso:
Yeah, (2) is fine, it doesn't have to follow PySide2 since it's sandboxed in our own `QtCompat` module. So long as it does something useful, I have no problem with adding something like this there.

MHendricks:
I will finish this pull request by just adding the test for `QtCompat.QWidget.grab` and documentation.
`QtCompat.grabWindow`. I'll add a second pull request specifically for `QtCompat.grabWindow`.

mangopipeline:
I did some looking into this today...
to me the solutions seems rather simple but i admit i could be missing something...

the Qt5 method is simple enough to map to Qt4....

in Qt4 we just drop the screen input....

Ideally this would work via the Qt.QtGui.QScreen grab name space wich is currently missing...

the implementation could look something like this....

```python
def screenGrab(screen, win_id,  x = 0,  y = 0,  width = -1, height = -1):
    if Qt.__binding__ in  ('PyQt5','PySide2):
       return Qt._QtGui.QScreen.grabWindow(screen, win_id, x, y, width, height)
    return Qt._QtGui.QPixmap.grabWindow(win_id, x, y, width, height)
```

in the implementation above we simply drop the usage of the screen variable when in Qt4
if this method was mapped to Qt.QtGui.QScreen it would be obvious that the code just follows Qt5 standards...

Feel free to let me know if I'm missing something...

mottosso:
Isn't this resolved? If there's a problem, could you try and illustrate what it is? All I see here is a solution to an already solved problem. :S

mangopipeline:
I’ll look into this again
I apologize if the code has been marged.

This ticket was marked as open 
And the original contributor said there would be a separate pull request to fix grabWindow 
which I did not find.

I searched the latest Qt.py and could not find 
a grabWindow method.

So assumed the pull request for this particular method never happened

mangopipeline:
@mottosso 
Maybe i'm not being explicit enough, or my communications is happening in the wrong issue.
if so i apologize...

This particular ticket addressed the .grab, .grabWidget and .grabWindow method,  but ".grabWindow" was left out of the fix, and promised as a future pull request which i don't think took place.

This has me a bit confused...

i'm currently looking at Qt.py v 1.2.1 which confirms this that grabWindow was never added to the compatibility members...

from looking at this particular issue it seems @MHendricks was working with a version of PySide2 in Maya  did not have a QScreen component, i'm guessing this is why the pull request for this method was never actually made...

 PySide2 has been updated (at least in maya) and now provides the previously missing  QScreen component.

which means that @MHendricks 3rd proposed solution is now possible.

> Make QtCompat.QScreen.grabWindow conform to PyQt5's signature. QtCompat.QScreen.grabWindow(screen, winId, ...). I think that if PySide2 ever adds support for QScreen this is the best choice, but it would force you to have to pass None, or some other object as the first argument(PyQt5 requires a type of QScreen so None is not the best option). This would require decorators for PySide2, PySide and PyQt4

and sense the pull request he suggested in this issue, doesn't seem to have taken place it seem to me that maybe it could be added now?

So my question to you is...
is there no plan to support the QPixmap.grabWindow change and there for you're considering this issue as resolved? 

Or is the issue resolved because QPixmap.grabWindow is supported and i'm just a giant cluts that is missing what under my nose? :D 

Thanks for your patience :)

MHendricks:
I guess I never added that pull request.

We didn't include `QScreen.grabWindow` at the time because it was not implemented in PySide2(at least the version included in Maya 2017 on windows.) Looks like PySide2 for Maya 2019 on windows now supports QScreen.

Now that its supported in PySide2 we could add `Qt.QtCompat.QScreen.grabWindow` similar to how you suggested. For new Qt5 we could probably just expose `QtGui.QScreen.grabWindow`, but for Qt4 we would probably need to make a decorator method that takes and ignores the screen argument using the `_build_compatibility_members` feature. This should work similar to how `Qt.QtCompat.QWidget.grab` was implemented in this pull request. I'm not sure how to handle the screen argument. You can pass None, but that seems like it would require the user to write extra code to deal with that.

I don't know if we can support QScreen in Qt4. This makes me think we probably don't want to support QScreen in Qt.py. We could probably use `QPixmap.grabWindow` to make a function that works in both Qt4 and Qt5. I think this would belong on `QtCompat.grabWindow` because its not replacing a Qt5 function(and QPixmap.grabWindow is deprecated in Qt 5.0). It could possibly go in https://github.com/mottosso/Qt5.py, but as both PySide2 and PyQt5 now support QScreen it probably already supports it.

mangopipeline:
@MHendricks thanks for the reply!
> I'm not sure how to handle the screen argument. You can pass None, but that seems like it would require the user to write extra code to deal with that.

seems to me that if we were to map QtApplication.instance().primaryScreen() in Qt4 to a method that returned None, We would not really be forcing developers to write "extra" code, as much as we would be forcing them to follow Qt5 convention, which in my opinion is a practice already enforced by Qt.py through the usage of QWidgets name space regardless of of Qt version.

This is why i was thinking the QtCompat name space is a bit hacky....

if Qt.py could  map QtWidgets.QApplication.instance().primaryScreen() to a method that returned None (in Qt4)

and Qt.py QtGui.QScreen.grabWindow could be mapped to the code...

``` python
    if Qt.__binding__ in  ('PyQt5','PySide2):
       return Qt._QtGui.QScreen.grabWindow(screen, win_id, x, y, width, height)
    return Qt._QtGui.QPixmap.grabWindow(win_id, x, y, width, height)
````
Then the expectation would just be to for developers to write Qt5 complaint code, and the code would work in both Qt versions....

If this is asking to much, then QtCompat could host a pimaryScreen, and grabWindow method with the same behaviors...

``` python
   # if Qt.py can map the desired qt4 behavior 
    screen = QtWidgets.QApplication.instance().primaryScreen()
    pixmap = QtGui.QScreen.grabWindow(screen, mainWindow.winId())
```

``` python
   # if Qt.py if QtCompat is the only way this would still follow Qt5 rules
    screen = QtCompat.primaryScreen()
    pixmap = QtCompat.grabWindow(screen, mainWindow.winId())
```
 for me personally, i would rather write Qt5 compatible code, as Qt4 seems to be on it's last legs...
if we can ever fully switch PySide2, we might be able to retire Qt.py all together someday. Not rallying some heavily in QtCompat could simplify this in the future...

however, get that it might not be an option...

MHendricks:
> if Qt.py could map QtWidgets.QApplication.instance().primaryScreen() to a method that returned None (in Qt4)

Qt.py has a project goal of "Qt.py should not affect other bindings running in same interpreter session." This means that we don't directly modify methods on Qt objects like `PyQt4.QtGui.QApplication`, that means that we have to go with `QtCompat.primaryScreen()`.

This seems reasonable to me, if any other use of `QScreen` is considered outside of the scope of Qt.py. It forces you to write your code as close to Qt5 style as possible, but still allows it to work in Qt4 and should be able to work in older versions of PySide2 that don't support QScreen.

```python
    screen = QtCompat.primaryScreen()
    pixmap = QtCompat.QScreen.grabWindow(screen, mainWindow.winId())
```
I think this would be consistent with the other QtCompat implementations, `QtCompat.QHeaderView.sectionsMovable(header_instance)` for example.

mangopipeline:
@MHendricks 
fair enough thanks for educating me :)

cheers!

mottosso:
> for me personally, i would rather write Qt5 compatible code, as Qt4 seems to be on it's last legs...

If it helps, just to underline what Mike mentioned, the [Qt5.py project](https://github.com/mottosso/Qt5.py) is suitable for when you don't need Qt 4 anymore. It can also be used in conjunction with Qt.py, in places where you want/need Qt 5 functionality, such as in a module or in an if-block. It's being designed with that in mind. I expect it to pick up steam over the coming year or so, as we're finally able to relinquish ourselves from the horrors of Python 2 and Qt 4. 👍

mangopipeline:
Thanks I’ll check it out... once I can retire qt4 and PySide2 in python 3 is a bit more stable I’m hoping to retire the Qt.py layer 
And the pyqt5 binding...

I just hope I don’t have to support Qt6 by then lol

PR #286, opened 2018-05-02
Added QtCompat.delete

kgermain:
Using QtCompat.delete(qpointer) like sip.delete or shiboken.delete is very usefull when reloading UI, where widget.deleteLater() does not destroy the object right away.


kgermain:
Tested with PyQt4, pyside and pyside2

mottosso:
Thanks @kgermain, agreed this would be a useful addition.

It looks like you've forgotten to pull the latest Qt.py from your fork however, you'll need to perform that merge before we can merge on this end.

One question, how come you use a lambda as opposed to just `Qt.QtCompat.delete = sip.delete`? If some of them have different argument signature, than that would be a good reason.

mottosso:
I also wonder whether we should allow/handle a `shiboken` object being passed into sip via.

It could be preceeded with a check, e.g. `if isinstance(obj, Qt.QtCore.QObject): then OK`. @fredrikaverpil what do you think?

kgermain:
That's true @mottosso, I forgot to pull, I'll update and add this extra check tomorrow for you to review =)
For the use of lambda, you're right, I didn't think too far on this one (or quite the opposite).
And by the way, thanks for qtpy, that's awesome !

fredrikaverpil:
This touches upon https://github.com/mottosso/Qt.py/issues/53 from summer of 2016! 😄 
Would it be worth collecting the methods _not directly_ under QtCompat but instead under e.g `QtCompat.GeneratorRunner.someMethod` or perhaps `QtCompat.BindingGenerator.someMethod`?

I'm not really content with "GeneratorRunner" or "BindingGenerator" as a term to collect shiboken/sips under one roof... as those terms are fairly obfuscating. Not sure what to collectively call them, but I think it would be worth a thought. 

> I also wonder whether we should allow/handle a shiboken object being passed into sip via.

Hm. Was that sentence ended prematurely?
I presume you mean you were wondering whether we should allow/handle a shiboken object passed to sip methods and vice versa?

> It could be preceeded with a check, e.g. if isinstance(obj, Qt.QtCore.QObject): then OK.

Unfortunately, I have very little experience with both shiboken and sip. I'm not sure that it has to be a QObject you are passing into these functions...?
The docs on both shiboken and sip state that you pass C++ objects into e.g. the `delete` method. But just in general, I would argue that it may not be Qt.py's responsibility to handle this, as long as you can pass the same object into both shiboken/sip.

Sorry for not really being clear on anything here :wink:

mottosso:
> Would it be worth collecting the methods not directly under QtCompat

I see what you mean, but when it comes to global functions like these and `loadUi`, I think having them here is fine. There's little room for mistaking it for anything else, and there likely won't be another `delete` from elsewhere also needing to be in the global space.

> I'm not sure that it has to be a QObject you are passing into these functions...?

No it'd be any object wrapping a Qt object, like colors and points too. Dunno, might not be possible, but I suspect passing a `shiboken` object into `QtCompat.delete` wrapping `sip` might cause a not-so-obvious crash.

fredrikaverpil:
> I see what you mean, but ... , I think having them here is fine.

👍 

> No it'd be any object wrapping a Qt object, like colors and points too. Dunno, might not be possible, but I suspect passing a shiboken object into QtCompat.delete wrapping sip might cause a not-so-obvious crash.

True. But hm... I'm just thinking out loud that it feels like an edge case ⚡️  that you'd have both sip and shiboken imported at the same time and confusing them with each other.

We generally don't handle stuff like this (e.g. what gets passed into the loadUi function), so what I meant before was that I think that just offering the mapping would be enough, from my perspective, and since at least for me, personally, I would have to research this (which would take time), rather than just implement the mapping. It's your call :)

kgermain:
Ok I added a def _delete(object) to handle the test before calling the wrapper's delete method.
And as @fredrikaverpil says, I think the case where a user mixes sip and shiboken is most likely to never happen.
Also, should we enhance a bit the delete method to handle a list of objects ?
like `if isinstance(object, list):
for obj in object:
_delete(obj)`

mottosso:
I had completely forgotten about this PR, sorry about that! I've actually been using a slightly simpler version in production for a few weeks and am reducing this PR to that. If/when we encounter said issue about trying to delete shiboken objects with sip and vice versa, let's address that then.

Thanks!

PR #288, opened 2018-05-16
New unittests and testing my modification of the translate function

Ahuge:
I think I had found a bug in my previous implementation of QtCompat.translate.

It should now support all cases that each binding support.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -804,7 +804,7 @@ def _translate(context, sourceText, *args):
             raise TypeError(
                 "Expected 4 or 5 arguments, got {0}.".format(len(args)+4)
             )
-        else:
+        elif len(args) == 1:
             encoding = args.pop(0)
 
 
--- a/Qt.py
+++ b/Qt.py
@@ -798,7 +798,7 @@ def _translate(context, sourceText, *args):
     elif len(args):
         if isinstance(args[0], str) or args[0] is None:
             disambiguation = args.pop(0)
-        if len(args) and isinstance(args[-1], int) or args[-1] is None:
+        if len(args) and (isinstance(args[-1], int) or args[-1] is None):
             n = args.pop(-1)
         if len(args) > 1:   # There is more args than we need.
             raise TypeError(
--- a/Qt.py
+++ b/Qt.py
@@ -43,7 +43,7 @@ import types
 import shutil
 
 
-__version__ = "1.2.0.b2"
+__version__ = "1.2.0.b3"
 
 # Enable support for `from Qt import *`
 __all__ = []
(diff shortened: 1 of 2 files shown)


tests.py
```diff
@@ -372,7 +372,8 @@ def test_load_ui_invalidxml():
 
     from xml.etree import ElementTree
     from Qt import QtWidgets, QtCompat
-    app = QtWidgets.QApplication(sys.argv)
+    app = QtWidgets.QApplication.instance() \
+          or QtWidgets.QApplication(sys.argv)
```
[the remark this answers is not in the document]
Ahuge: :|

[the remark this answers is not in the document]
Ahuge: bummer yeah, I'll make it look prettier haha

Ahuge:
Cool looks like we're good now.


Thoughts on this PR?

Qt.py
```diff
@@ -783,14 +783,30 @@ def _translate(context, sourceText, *args):
...
+    # If last arg is "n", we require "disambiguation" to be passed
+    #    In Qt4, if "n" is passed, encoding must be passed.
+    #    In Qt5
+    if len(args) > 3:
         raise TypeError(
-            "Expected 4 or 5 arguments, got {0}.".format(len(args)+2))
+                "Expected 4 or 5 arguments, got {0}.".format(len(args)+2)
+        )
+    elif len(args):
+        if isinstance(args[0], (str, type(None))):
```
mottosso: `type(None)` seem a little ad-hoc, how about..

```python
if isinstance(args[0], str) or args[0] is None:
   ...
```

?

Qt.py
```diff
@@ -783,14 +783,30 @@ def _translate(context, sourceText, *args):
...
+
+    # First arg needs to be (str, NoneType)
+    # Last arg needs to be (int, NoneType)
+    # If last arg is "n", we require "disambiguation" to be passed
+    #    In Qt4, if "n" is passed, encoding must be passed.
+    #    In Qt5
+    if len(args) > 3:
         raise TypeError(
-            "Expected 4 or 5 arguments, got {0}.".format(len(args)+2))
+                "Expected 4 or 5 arguments, got {0}.".format(len(args)+2)
```
mottosso: Surprised the Hound didn't point this out, but there should be spaces around the `+` sign.

mottosso:
Nitpicking aside, I think it looks allright. Just to confirm, does this now handle arguments that conform to PySide2, and massages them to make it the other bindings happy?

Qt.py
Ahuge: Yeah that works for me, I'll update them.

Ahuge:
Uh sort of.

It allows you to pass the encoding in, but if you are on a Qt5 binding, it will not pass that parameter down into the binding.

And conversely, it will autofill the encoding to QCoreApplication.CodecForTr if it was expected, for example the underlying binding is Qt4.

mottosso:
Ok, just making sure, the same code must be able to run in all bindings. And looking at the tests, some tests only run on Qt 5 bindings. Does that mean Qt.py running on Qt 5 would act differently than on Qt 4? If so, is there a possibility of code working perfectly fine with Qt.py on Qt 4 to break on Qt 5, or vice versa?

Ahuge:
You know...they should work in both, I don't know why I did that...one sec

tests.py
```diff
@@ -292,7 +299,10 @@ def test_load_ui_signals():
     """Tests to see if the baseinstance connects signals properly"""
     import sys
     from Qt import QtWidgets, QtCompat
-    app = QtWidgets.QApplication(sys.argv)
+    app = (
```
mottosso: Question about this actually; during tests, every test should be running in its own Python interpreter, so there *should* never be any need to look for an existing instance. Have you run into problems trying to instantiate a new application? That could be a bug.

mottosso:
Hey @Ahuge it's been a while, is still still something you'd like to see merged? The hound had a lot of issues with it it seems. :S

Ahuge:
Hey, 

I totally forgot I had this open still.


Let me take a look at it today!

Ahuge:
This has been here forever and I haven't looked at it in years.
I am going to assume that this is fine to just close.

mottosso:
Yes, unclear what caused the CLA assistant to make noise. And yes, it seems no one has run into this issue, so it's probably safe to leave as-is.

PR #301, opened 2018-11-01
Update README.md to include Spin VFX

herronelou:
We've been using Qt.py extensively (and exclusively for all Qt work) for the last year, so just adding to the list of studios using Qt.py.


mottosso:
Thanks @herronelou!

PR #302, opened 2018-11-06
Update to new location of QStringListModel in PySide2.

freidrichen:
QStringListModel was moved from QtGui to QtCore in Qt5. PySide2 kept it in QtGui for a while, but have now also moved it to QtCore. (See https://bugreports.qt.io/browse/PYSIDE-614.)

This means that Qt.py also need to update to be compatible with newer versions of PySide2.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -917,7 +917,7 @@ These members from the original submodule are misplaced relative PySide2
 """
 _misplaced_members = {
     "PySide2": {
-        "QtGui.QStringListModel": "QtCore.QStringListModel",
+        "QtCore.QStringListModel": "QtCore.QStringListModel",
         "QtCore.Property": "QtCore.Property",
         "QtCore.Signal": "QtCore.Signal",
         "QtCore.Slot": "QtCore.Slot",


freidrichen:
hm... looks like the tests are running on older versions of Qt5/PySide2 than those available on PyPI?

mottosso:
Hmm, this won't work with Maya 2017-2018, where it still resides in QtGui. We don't have a strategy for compatibility across versions of the VFX Platform at the moment, so I'm not 100% sure how to proceed with this. Preferably, we'd let the user choose which version of VFX Platform to comply with; if it's 2019, then this PR should apply.

@fredrikaverpil thoughts?

> hm... looks like the tests are running on older versions of Qt5/PySide2 than those available on PyPI?

Yes, the tests are running against the 2017 and 2018 versions of the VFX Platform.

freidrichen:
> Hmm, this won't work with Maya 2017-2018, where it still resides in QtGui. We don't have a strategy for compatibility across versions of the VFX Platform at the moment, so I'm not 100% sure how to proceed with this. Preferably, we'd let the user choose which version of VFX Platform to comply with; if it's 2019, then this PR should apply.

I see. For this particular issue i guess it would be pretty easy to simply look for QStringListModel in both QtCore and QtGui and always put it in QtCore regardless of where it was found, right?

mottosso:
For this case, I'm not too sure! It would mean that code written for Maya 2018 and below would stop working in 2019 and above. For complete compatibility, it would probably need to exist in both places, but that may just delay actually having to deal with it.

freidrichen:
It seems this wasn't as easy to fix as I first thought =P

For the moment I will use a local workaround and just leave the PR as is. Feel free to close it if you decide to go for some other solution moving forward.

mottosso:
Let us know how you get along with your workaround; if it works for 2017-2019 versions of Maya (for example), then it's likely a good enough workaround for merge into Qt.py. :)

fredrikaverpil:
> Hmm, this won't work with Maya 2017-2018, where it still resides in QtGui. We don't have a strategy for compatibility across versions of the VFX Platform at the moment, so I'm not 100% sure how to proceed with this. Preferably, we'd let the user choose which version of VFX Platform to comply with; if it's 2019, then this PR should apply.
> 
> @fredrikaverpil thoughts?

Yeah, I believe there are a few changes like this (but I can't find them now when I am typing this). And like you say, if we move such attributes from `QtGui` to `QtCore` this we will break currently working code in the PySide2 version inside Maya 2016.5, 2017, 2018 and likely also upcoming Maya versions which are still based on Qt 5.6.

I assume that Maya will stick with the now ancient and pre-alpha version of PySide2 (based on Qt 5.6) which currently ships with Maya 2016.5, 2017 and 2018 until they update to the official PySide2 release which is [going to be based on Qt 5.12](http://blog.qt.io/blog/2018/06/13/qt-python-5-11-released/). When doing this, I assume Autodesk also will want to update Maya's Qt to version 5.12. This is a somewhat major undertaking given all their modifications they maintain for Qt, and could delay the adoptation of PySide2-5.12 in Maya. Any way, I personally don't expect to see this happening for the next major Maya release, in case a new major release of Maya is imminent since neither PySide2-5.12 or Qt 5.12 has been finished/released yet. Although I certainly can dream, and hope the next Maya release will come with PySide2-5.12, as this will make the lives of UI wizards so much nicer. :wink:
**Disclaimer:** I am a Maya beta tester, under NDA on these things, but I also do not know what their stand is on this whole subject and I am honestly and plainly just speculating here.

Getting back on topic, and in my view;

We cannot merge this PR as it will break currently working code which uses `QStringListModel`, just like @mottosso has previously stated.

I believe we will need to start looking into how we can support both PySide2-5.6 _and_ newer releases such as PySide2-5.12 (which is actually part of VFX platform CY2019) and not get stuck in yesteryear's tech. We could branch Qt.py into two different versions or we could make the one and the same Qt.py detect which of these PySide2 versions is currently in use. Or perhaps we should utilize `QtCompat` or `QtSiteConfig.py` for these kinds of changes. Perhaps a nice approach would be to keep one `QtSiteConfig.py` per VFX Platform-defined PySide2 release or so.

Is PyQt in the same boat, does any of you know?
I somehow feel we should prioritize PySide2 at this point in time, but I may be wrong here. I am not particularly aware of how widespread the use of PyQt is in our industry (visual effects).

mottosso:
> Or perhaps we should utilize QtCompat

That's a good idea!

We could exclude this member from being part of Qt.py, to avoid anyone using it accidentally, and expose a version compatible with this and future versions via `QtCompat`. I think that should solve it, and be real explicit about it.

> Is PyQt in the same boat, does any of you know?

As far as Qt.py is concerned, all members should conform to PySide2. If PyQt does anything differently, we'll map Qt.py to make it look like PySide2. I expect they both follow (as closely as makes sense) the original Qt organisation.

mottosso:
This should be resolved now, in 1.2.0.b3 at least. Any issues, feel free to re-open.

PR #304, opened 2018-12-05
Dual PySide2 version support

Ahuge:
This might be a bit of a brute force way of supporting both but it appears to work.

Not sure if it's really an appropriate solution for the main upstream or not.


mottosso:
Related to https://github.com/mottosso/Qt.py/pull/302

The issue with this (I think?) is that if you do manage to write an application that uses this via `QtGui`, then that could would break in this other version.

What we should do, is pick one of the two, and stick with it. Then if someone writes with Qt.py, it'll work regardless of the underlying version.

It'd be a departure from PySide2, but PySide2 is seems to already depart from itself and isn't leaving us much choice. What matters is that your code doesn't break, and that it works across bindings (and now, across versions of PySide2, sigh..).

Thoughts?

Ahuge:
This PR makes it so that if your version of Pyside2 has it in QtGui, the older Pyside2, it will map to QtCore

If you're on the newer Pyside2, it'll still map to QtCore. Behind the scenes, one of the misplaced members will fail because per Pyside2 binding it only exists in one place

mottosso:
Ah, yes now I see. Yes, I think this is the way to go. Now it can only ever be available in a single location. It may not be the cleanest implementation (as you mention, it'll work because it fails) but the result should suffice. Happy for you to merge this once you're happy. 👍

Ahuge:
Cool thanks

PR #316, opened 2019-03-30
Add QtCompat.isValid()

OEP:
I brought this up originally in #53 where wrapInstance() and friends were being discussed, and its cousin delete() was added in #286. It feels appropriate to add this now.

This checks the deleted state of the underlying C++ object, which is useful for Python programmers who are about to make calls on a QObject and have some reason to believe it may have been deleted out from under them.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -849,6 +849,13 @@ if sys.version_info <= (3, 4):
         finally:
             app.exit()
 
+    def test_isValid():
+        from Qt import QtCompat, QtCore
+        obj = QtCore.QObject()
+        assert QtCompat.isValid(obj)
+        QtCompat.delete(obj)
+        assert not QtCompat.isValid(obj)
+
 
 if binding("PyQt4"):
     def test_preferred_pyqt4():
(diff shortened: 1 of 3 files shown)


mottosso:
Looks great to me, thank you! For consistency, I'd prefer if you could sign the CLA to protect users and future contributors from legal trouble.

tests.py
```diff
@@ -849,6 +849,13 @@ def test_implicit_wrapInstance():
         finally:
             app.exit()
 
+    def test_isValid():
+        from Qt import QtCompat, QtCore
+        obj = QtCore.QObject()
+        assert QtCompat.isValid(obj)
+        QtCompat.delete(obj)
+        assert not QtCompat.isValid(obj)
```
OEP: This test is only running on Python 3.4 and before, but technically it works for all the `docker run` commands I see in your README. I just put this next to `wrapInstance()`, no idea if that was the right thing.

mottosso: I think that's right; in the comment above that block, it says shiboken1 isn't available prior to that, and this test would depend on it.

mottosso:
Sorry for the delay on this one, completely slipped my mind. Thanks @OEP

PR #320, opened 2019-07-01
Use effectively vfxplatform2018 for docker build 2018

BoboTiG:


mottosso:
Thanks @BoboTiG, this makes sense.

PR #324, opened 2019-09-18
trying to fix loadUi widget placement issue

### LINKED ISSUE #323: Dialog with loadUi placement inconsistency on show
Whenever I try to create a Dialog which has an existant active parent Window, the placement of the Dialog on show (or exec_) is dependant upon the binding in question. 

For example if have:

```python
#simple_load.py

from __future__ import print_function
import sys

from Qt.QtWidgets import (
        QWidget, QApplication, QVBoxLayout, QPushButton, QDialog)
from Qt.QtCompat import loadUi


class Dialog(QDialog):
        super(Dialog, self).__init__(parent=parent)
        loadUi("untitled.ui", self)
        self.pushButton.clicked.connect(self.close)


class Widget(QWidget):
        super(Widget, self).__init__(parent=parent)
        self.layout = QVBoxLayout(self)
        self.btn = QPushButton('Click')
        self.btn.clicked.connect(self.clicked)
        self.layout.addWidget(self.btn)

    def clicked(self):
        dialog = Dialog(self)
        dialog.exec_()


if __name__ == "__main__":
    app = QApplication(sys.argv)
    widget = Widget()
    widget.show()
    app.exec_()

```

with `untitled.ui` as:

```xml
<ui version="4.0">
 <class>Dialog</class>
 <widget class="QDialog" name="Dialog">
  <property name="geometry">
   <rect>
    <x>0</x>
    <y>0</y>
    <width>199</width>
    <height>60</height>
   </rect>
  </property>
  <property name="windowTitle">
   <string>Dialog</string>
  </property>
  <layout class="QVBoxLayout" name="verticalLayout">
   <item>
    <widget class="QLabel" name="label">
     <property name="text">
      <string>Message</string>
     </property>
    </widget>
   </item>
   <item>
     <property name="text">
      <string>OK</string>
     </property>
    </widget>
   </item>
  </layout>
 </widget>
 <resources/>
 <connections/>
</ui>
```
and I do 

```cmd
(nukenv) D:\workspace\test>set QT_PREFERRED_BINDING=PySide2;PyQt5
(nukenv) D:\workspace\test>python simple_load.py
```
I will get the second dialog at the top-left corner.

and if I do
```cmd
(nukenv) D:\workspace\test>set QT_PREFERRED_BINDING=PyQt5;PySide2
(nukenv) D:\workspace\test>python simple_load.py
```
I will get the second dialog at the center of the parent window.

if I use the `setup_ui` function as in [baseinstance1.py](https://github.com/mottosso/Qt.py/blob/master/examples/loadUi/baseinstance1.py)

I will get the second dialog in the middle of the screen regardless of the binding. This I can manipulate by overridiing `showEvent` and calling `move`. However, `setup_ui` and the override seem like just unnecessary inconvenient boiler plate in otherwise somewhat clean `Qt.py`.

Am I wrong to expect consistent behaviour out of the box?

MagicRedDeer:
This is an attempt to fix #323. Where a UI loaded in `PySide` from a `.ui` file does not show up at the top of the parent widget as expected.

This solution creates a new widget as a child or the "base instance" and then uses the new widgets layout or central widget on the base instance.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -930,12 +930,15 @@ def _loadUi(uifile, baseinstance=None):
                 if self.baseinstance:
                     # correct geometry on show
                     parent = self.baseinstance.parentWidget()
-                    if parent and hasattr(parent, 'geometry'):
+                    if parent and hasattr(parent, 'frameGeometry'):
                         geo = self.baseinstance.geometry()
                         pgeo = parent.geometry()
+                        pfgeo = parent.frameGeometry()
+                        hdiff = pfgeo.height() - pgeo.height()
                         self.baseinstance.move(
-                                pgeo.x() + pgeo.width()/2 - geo.width()/2,
-                                pgeo.y() + pgeo.height()/2 - geo.height()/2)
+                            pgeo.x() + pgeo.width()/2 - geo.width()/2,
+                            pgeo.y() + pgeo.height()/2 - geo.height()/2 -
+                            hdiff)
 
                 # Workaround for PySide 1.0.9, see issue #208
                 widget.parentWidget()
--- a/Qt.py
+++ b/Qt.py
@@ -927,6 +927,23 @@ def _loadUi(uifile, baseinstance=None):
                 widget = Qt._QtUiTools.QUiLoader.load(
                     self, uifile, *args, **kwargs)
 
+                if self.baseinstance:
+
+                    # copy over all the widgets and layouts from attributes
+                    for member in dir(widget):
+                        value = getattr(widget, member)
+                        if isinstance(value, Qt.QtCore.QObject):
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks for this. Could you implement a test for this as well? That way it'll get run during CI and ensure it works for all bindings on all platforms. Also if you could post a reproducible we could run locally as well, to confirm it works visually that would be great.

mottosso:
Hi @MagicRedDeer, thanks again for submitting this PR. I know writing tests are a drag, but if you could paste something even in a comment, then we could run it on our end and confirm it works and doesn't break anything else. That would help this PR getting merged.

Issue #333, opened 2020-02-03
Qt_py unable to import sip in newer versions


maxnbk:
We have a PyQt5 build where, for compatibility reasons, we've built SIP to provide a private module (`--modulename=PyQt5.sip`).
We're getting a:
```
Traceback (most recent call last):
  File "/sw/packages/internal/mhandoff/5.12.2/bin/mhandoff", line 5, in <module>
    from Qt_py.Qt import QtWidgets
  File "/sw/dev/maxnbk/packages/Qt_py/1.2.3.m0/python/Qt_py/Qt.py", line 1874, in <module>
    _install()
  File "/sw/dev/maxnbk/packages/Qt_py/1.2.3.m0/python/Qt_py/Qt.py", line 1810, in _install
    available[name]()
  File "/sw/dev/maxnbk/packages/Qt_py/1.2.3.m0/python/Qt_py/Qt.py", line 1522, in _pyqt5
    _setup(module, extras)
  File "/sw/dev/maxnbk/packages/Qt_py/1.2.3.m0/python/Qt_py/Qt.py", line 1257, in _setup
    module, name)
  File "/sw/dev/maxnbk/packages/Qt_py/1.2.3.m0/python/Qt_py/Qt.py", line 1245, in _import_sub_module
    module = getattr(module, level)
AttributeError: 'module' object has no attribute 'sip'
```
The following fixes it (switching the order of attempted imports), and it continues to work, for both PyQt-5.9 as well as the newer PyQt-5.12.
```
diff --git a/Qt.py b/Qt.py
index fe4b45f..343d665 100644
--- a/Qt.py
+++ b/Qt.py
@@ -1508,13 +1508,12 @@ def _pyqt5():
     extras = ["uic"]

     try:
-        import sip
+        # Relevant to PyQt5 5.11 and above
+        from PyQt5 import sip
         extras += ["sip"]
     except ImportError:
-
-        # Relevant to PyQt5 5.11 and above
         try:
-            from PyQt5 import sip
+            import sip
             extras += ["sip"]
         except ImportError:
             sip = None
```

We might have done something wrong somewhere along the chain, but I wanted to better understand the issues at play, so posting this issue to see if it either helps anything/anyone else, or whether I should be taking another approach to solve the issue, or if indeed it really is a bug. Insight appreciated.

mottosso:
[...]

PR #334, opened 2020-02-05
fix: swap sip import order for pyqt5 to account for compat w private …

maxnbk:
…module names in 5.11+

for issue #333

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1508,13 +1508,13 @@ def _pyqt5():
     extras = ["uic"]
 
     try:
-        import sip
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks @maxnbk, this looks solid. Would you mind bumping [the version](https://github.com/mottosso/Qt.py/blob/90207a6de95a795e6333c428652f6bd05e7d26ea/Qt.py#L47) to 1.2.4? That way I can make a release with this fix.

maxnbk:
version bumped.

mottosso:
Thanks!

PR #335, opened 2020-03-06
Add QtSiteConfig.preferred_binding

MHendricks:
I need to use PyQt5 in Maya/3ds Max/Houdini, etc. It works great until we need to install a third party script that was written expecting to be using PySide/PySide2. For the most part it works, but eventually you end up with a argument order difference or something else causing the PySide2 code to error out.

This merge request allows you to easily specify the preferred binding for each individual Qt.py import based on the full module path.

I can use the alternative workflow if we decide this isn't worth adding, but it makes the developing and testing process easier if `QtSiteConfig.preferred_binding` manages things for you.

##### More details from the QtSiteConfig readme:

If you need to use both PyQt and PySide in the same application you may run into problems if you just set the QT_PREFERRED_BINDING environment variable.

For example, in Maya you set QT_PREFERRED_BINDING to PyQt5 to load your PyQt5 code. You then try to use the https://www.studiolibrary.com/ tool, it was written
expecting to only use PySide or PySide2, a reasonable assumption for a Maya tool. This causes errors when using it.

To work around this, you can make sure that your PyQt setup code imports studio library, then sets the QT_PREFERRED_BINDING env variable. This is nice if you want to be able to
reload your Qt.py class for debugging your QtSiteConfig setup as importing Qt will still use PyQt. You are not able to reload the studiolibrary vendored Qt.py. This also does not work in cases where you are unable to import the vendored Qt.py at this point. This also means that any time your users want to use a new tool, you have to add support for it.

Alternatively you can make your PyQt setup code set QT_PREFERRED_BINDING to PyQt5, import your Qt.py, and clear the QT_PREFERRED_BINDING environment variable
so when studio library is imported it will use PySide2. You would need to make sure to re-set the QT_PREFERRED_BINDING environment variable to allow you to reload your Qt.py.

If you implement the `preferred_binding(module_name, preferred_order)` function in your QtSiteConfig module, you can choose the desired behavior for each Qt.py when it is imported without any of the above drawbacks.

```python
def preferred_binding(module_name, preferred_order):
    """ If this is processing the global Qt.py import, use PyQt, otherwise
    use the default preferred_order.

    Args:
        module_name (str): The full module name of the Qt.py import.
        preferred_order (list): This list should be modified to control
            the order Qt bindings are loaded. This respects the
            QT_PREFERRED_BINDING environment variable.
    """
    if module_name == 'Qt':
        # The Qt module was imported directly from a path in sys.path
        # I want PyQt5 to be the first package loaded.
        preferred_order.insert(0, 'PyQt5')
    # The Qt module was imported from a vendored package for example
    # `mypackage.vendor.Qt` Don't modify the order.
```

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -42,6 +42,7 @@ import sys
 import types
 import shutil
 import importlib
+import json
 
 
 __version__ = "1.2.4"
@@ -51,6 +52,7 @@ __all__ = []
 
 # Flags from environment variables
 QT_VERBOSE = bool(os.getenv("QT_VERBOSE"))
+QT_PREFERRED_BINDING_JSON = os.getenv("QT_PREFERRED_BINDING_JSON", "")
 QT_PREFERRED_BINDING = os.getenv("QT_PREFERRED_BINDING", "")
 QT_SIP_API_HINT = os.getenv("QT_SIP_API_HINT")
 
@@ -1781,11 +1783,34 @@ class MissingMember(object):
 
 
 def _install():
-    # Default order (customise order and content via QT_PREFERRED_BINDING)
+    # Default order (customize order and content via QT_PREFERRED_BINDING)
     default_order = ("PySide2", "PyQt5", "PySide", "PyQt4")
-    preferred_order = list(
-        b for b in QT_PREFERRED_BINDING.split(os.pathsep) if b
-    )
+    preferred_order = None
+    if QT_PREFERRED_BINDING_JSON:
+        # A per-vendor preferred binding customization was defined
+        # This should be a dictionary of the full Qt.py module namespace to
+        # apply binding settings to. The "default" key can be used to apply
+        # custom bindings to all modules not explicitly defined. If the json
+        # data is invalid this will raise a exception.
+        # Example:
+        #   {"mylibrary.vendor.Qt": ["PySide2"], "default":["PyQt5","PyQt4"]}
+        try:
+            preferred_bindings = json.loads(QT_PREFERRED_BINDING_JSON)
+        except ValueError as e:
+            # Python 2 raises ValueError, Python 3 raises json.JSONDecodeError
+            # a subclass of ValueError
+            _log("JSONDecodeError: {}".format(e))
+        else:
+            preferred_order = preferred_bindings.get(__name__)
+            # If no matching binding was used, optionally apply a default.
+            if preferred_order is None:
+                preferred_order = preferred_bindings.get("default", None)
+    if preferred_order is None:
+        # If a json preferred binding was not used use, respect the
+        # QT_PREFERRED_BINDING environment variable if defined.
+        preferred_order = list(
+            b for b in QT_PREFERRED_BINDING.split(os.pathsep) if b
+        )
 
     order = preferred_order or default_order
 
--- a/README.md
+++ b/README.md
@@ -184,11 +184,12 @@ This also covers inconsistencies between bindings. For example PyQt4's QFileDial
 
 These are the publicly facing environment variables that in one way or another affect the way Qt.py is run.
 
-| Variable             | Type  | Description
-|:---------------------|:------|:----------
-| QT_PREFERRED_BINDING | str   | Override order and content of binding to try.
-| QT_VERBOSE           | bool  | Be a little more chatty about what's going on with Qt.py
-| QT_SIP_API_HINT      | int   | Sets the preferred SIP api version that will be attempted to set.
+| Variable                  | Type  | Description
+|:--------------------------|:------|:----------
+| QT_PREFERRED_BINDING_JSON | str   | Override order and content of binding to try. This can apply per Qt.py namespace.
+| QT_PREFERRED_BINDING      | str   | Override order and content of binding to try. Used if QT_PREFERRED_BINDING_JSON does not apply.
+| QT_VERBOSE                | bool  | Be a little more chatty about what's going on with Qt.py
+| QT_SIP_API_HINT           | int   | Sets the preferred SIP api version that will be attempted to set.
 
 <br>
 
@@ -227,11 +228,30 @@ PyQt5
 Constrain available choices and order of discovery by supplying multiple values.
 
 ```bash
-# Try PyQt first and then PySide, but nothing else.
-$ export QT_PREFERRED_BINDING=PyQt:PySide
+# Try PyQt4 first and then PySide, but nothing else.
+$ export QT_PREFERRED_BINDING=PyQt4:PySide
 ```
 
 Using the OS path separator (`os.pathsep`) which is `:` on Unix systems and `;` on Windows.
+
+If you need to control the preferred choice of a specific vendored Qt.py you can use the `QT_PREFERRED_BINDING_JSON` environment variable instead.
+
+```json
+{
+    "Qt":["PyQt5"],
+    "myproject.vendor.Qt":["PyQt5"],
+    "default":["PySide2"]
+}
+```
+
+This json data forces any code that uses `import Qt` or `import myproject.vendor.Qt` to use PyQt5(`from x import Qt` etc works too, this is based on `__name__` of the Qt.py being imported). Any other imports of a Qt module will use the "default" PySide2 only. If `"default"` is not provided or a Qt.py being used does not support `QT_PREFERRED_BINDING_JSON`, `QT_PREFERRED_BINDING` will be respected.
+
+```bash
+# Try PyQt5 first and then PyQt4 for the Qt module name space.
+$ export QT_PREFERRED_BINDING_JSON="{"Qt":["PyQt5","PyQt4"]}"
+# Use PyQt4 for any other Qt module name spaces.
+$ export QT_PREFERRED_BINDING=PySide2
+```
 
 <br>
 
--- a/tests.py
+++ b/tests.py
@@ -550,6 +550,10 @@ def test_vendoring():
     shutil.copy(os.path.join(os.path.dirname(__file__), "Qt.py"),
                 os.path.join(vendor, "Qt.py"))
 
+    # Copy real Qt.py into the root folder
+    shutil.copy(os.path.join(os.path.dirname(__file__), "Qt.py"),
+                os.path.join(self.tempdir, "Qt.py"))
+
     print("Testing relative import..")
     assert subprocess.call(
         [sys.executable, "-c", "import myproject"],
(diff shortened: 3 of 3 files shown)


mottosso:
> For example, in Maya you set QT_PREFERRED_BINDING to PyQt5 to load your PyQt5 code. You then try to use the https://www.studiolibrary.com/ tool, it was written
expecting to only use PySide or PySide2, a reasonable assumption for a Maya tool. This causes errors when using it.

Trying to wrap my head around how this works; that tool seems to reference Qt.py only, and should thus work with PyQt5 as well. What errors are you expecting? :O

There's one thing that doesn't make sense to me. How would changing the preferred order *after* Qt.py has already been imported into the running Python session have any effect? Once Qt.py has been imported, it would get stored amongst other imported modules in `sys.modules`, to be referenced the next time anyone tries to import it.

Is the goal to allow tools like studiolibrary to have its own copy of `sys.modules["Qt"]` that references PySide2, and for your tool to have another `sys.modules["Qt"]` that references PyQt5? The test doesn't take into account what should happen during a second import, or at least I think not?

And if its this specific, why not just `import PyQt5`? Qt.py works just fine alongside an explicit import.

MHendricks:
> There's one thing that doesn't make sense to me. How would changing the preferred order _after_ Qt.py has already been imported into the running Python session have any effect? Once Qt.py has been imported, it would get stored amongst other imported modules in `sys.modules`, to be referenced the next time anyone tries to import it.

I'm not changing the preferred order after import of a given Qt.py module. I'm controlling what what the preferred order is the first time a given module is imported(or has reload called on it). `import Qt` and `import studiovendor.Qt`. are two unique instances of Qt.py and can even have different versions of the Qt.py code. They respect the QT_PREFERRED_BINDING environment variable only at the first time they are imported.

> Is the goal to allow tools like studiolibrary to have its own copy of sys.modules["Qt"] that references PySide2, and for your tool to have another sys.modules["Qt"] that references PyQt5?

Tools like studiolibrary already have their own copy of `sys.modules['Qt']`, though for studio library its actually `sys.modules['studiovendor.Qt']`, golaem uses `sys.modules['glm.Qtpy.Qt']`, BroTools uses `sys.modules['BroTools.core.broqt.vendor.Qt']`, but that is only accessible after the BroTools plugin is loaded. Each of these are their own unique instance of the Qt.py code respecting the QT_PREFERRED_BINDING environment variable when each of them was imported.

If your site-packages looks like this.
```
|  Qt.py
|  PyQt5
|  PySide2
\---myproject
    |   __init__.py
    \---vendor
            Qt.py
            __init__.py
```
Running this code should output: `Qt: PyQt5 myproject: PySide2`.
```python
import os
os.environ['QT_PREFERRED_BINDING'] = 'PyQt5'
import Qt
del(os.environ['QT_PREFERRED_BINDING'])
import myproject.vendor.Qt
print('Qt: {} myproject: {}'.format(Qt.__binding__, myproject.vendor.Qt.__binding__))
```
This setup works and I can use it without the need for this pull request, but if I wanted to call `reload(Qt)`, it ends up re-initialized using PySide2 as its binding. This requires me teaching my fellow developers how to work around this. The pull requests makes it so it just works automagically once I implement `QtSitePackages.preferred_binding`.

<br>

> The test doesn't take into account what should happen during a second import, or at least I think not?

The test code is checking that the `__binding__` is set correctly in the subprocess asserts, and I check that the `QtSiteConfig.preferred_binding` function was run for both the `import myproject.vendor.Qt` and `import Qt` calls by checking the subprocess output for prints. I should probably add more comments to better explain what each part of the test is doing.

> And if its this specific, why not just import PyQt5? Qt.py works just fine alongside an explicit import.

All three of these tools are using a copy of Qt.py to transition from PySide to PySide2. I'm using Qt.py to transition from PyQt4 to PyQt5. My code is specific to PyQt not PyQt5. I am also using compiled libraries that don't have PySide bindings.


I don't have easy access to all of the errors I've run into but for studio library, the [.ui](https://github.com/krathjen/studiolibrary/blob/master/src/studiolibrary/widgets/PlaceholderWidget.ui#L4) file says its a QFrame class, but the [widget](https://github.com/krathjen/studiolibrary/blob/master/src/studiolibrary/widgets/placeholderwidget.py#L21) is using QWidget.  PySide2 is fine with this change, but PyQt raises this error.

```python
  File "c:\blur\dev\blurdev\blurdev\gui\toolbars\toolstoolbar.py", line 242, in runAction
    action.exec_()
  File "c:\blur\dev\blurdev\blurdev\gui\toolbars\toolstoolbar.py", line 36, in exec_
    blurdev.runTool(self._toolID)
  File "c:\blur\dev\blurdev\blurdev\__init__.py", line 384, in runTool
    tool.exec_(macro)
  File "c:\blur\dev\blurdev\blurdev\tools\tool.py", line 94, in exec_
    architecture=self.architecture()
  File "c:\blur\dev\blurdev\blurdev\cores\core.py", line 1144, in runScript
    execfile(filename, scope)
  File "C:\blur\dev\local/code/python/tools/studiolibrary/main.pyw", line 17, in <module>
    studiolibraryExecute()
  File "C:\blur\dev\local/code/python/tools/studiolibrary/main.pyw", line 13, in studiolibraryExecute
    studiolibrary.main()
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\main.py", line 45, in main
    libraryWindow = mayalibrarywindow.MayaLibraryWindow.instance(*args, **kwargs)
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\librarywindow.py", line 172, in instance
    libraryWindow = cls(name=name)
  File "C:\Program Files\Autodesk\Maya2019\Python\lib\site-packages\maya\app\general\mayaMixin.py", line 78, in __init__
    super(MayaQWidgetBaseMixin, self).__init__(parent=parent, *args, **kwargs) # Init all baseclasses (including QWidget) of the main class
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\librarywindow.py", line 401, in __init__
    self.updatePreviewWidget()
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\librarywindow.py", line 1645, in updatePreviewWidget
    self.setPreviewWidgetFromItem(self._currentItem, force=True)
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\librarywindow.py", line 1667, in setPreviewWidgetFromItem
    self.clearPreviewWidget()
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\librarywindow.py", line 1640, in clearPreviewWidget
    widget = studiolibrary.widgets.PlaceholderWidget()
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiolibrary\widgets\placeholderwidget.py", line 22, in __init__
    studioqt.loadUi(self)
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studioqt\utils.py", line 115, in loadUi
    widget.ui = QtCompat.loadUi(path, widget)
  File "C:\blur\dev\local\workgroups\maya\3rdParty\scripts\studiovendor\Qt.py", line 836, in _loadUi
    return Qt._uic.loadUi(uifile, baseinstance)
  File "C:\blur\lib\msvc2015_64_qt5\19493\python\Lib\site-packages\PyQt5\uic\__init__.py", line 226, in loadUi
    return DynamicUILoader(package).loadUi(uifile, baseinstance, resource_suffix)
  File "C:\blur\lib\msvc2015_64_qt5\19493\python\Lib\site-packages\PyQt5\uic\Loader\loader.py", line 72, in loadUi
    return self.parse(filename, resource_suffix, basedir)
  File "C:\blur\lib\msvc2015_64_qt5\19493\python\Lib\site-packages\PyQt5\uic\uiparser.py", line 1011, in parse
    actor(elem)
  File "C:\blur\lib\msvc2015_64_qt5\19493\python\Lib\site-packages\PyQt5\uic\uiparser.py", line 812, in createUserInterface
    self.toplevelWidget = self.createToplevelWidget(cname, wname)
  File "C:\blur\lib\msvc2015_64_qt5\19493\python\Lib\site-packages\PyQt5\uic\Loader\loader.py", line 57, in createToplevelWidget
    (type(self.toplevelInst), classname)))
TypeError: ('Wrong base class of toplevel widget', (<class 'studiolibrary.widgets.placeholderwidget.PlaceholderWidget'>, 'QFrame'))
```

This is a easy fix, but as one of the only places that use PyQt5 inside of Maya, I've run into similar issues with all 3 of the 3rd party Maya tools I've had to add support for that use a vendored Qt.py.

mottosso:
> `sys.modules['studiovendor.Qt']`

Aaaah, doh. Yes, you're right. That's genius.

Spontaneously, rather than providing a function, I would have gone for a dictionary.

```py
QtSiteConfig.preferred_binding = {
  "studiovendor.Qt": ["PySide2"],
  "mylibrary.vendor.Qt": ["PyQt4"],
  "Qt": ["PyQt5", "PyQt4"],
}
```

A function could return all manner of things, and throw all manner of exceptions that need handling and reporting to the end user. A dictionary would reference something that either exists or does not, with a list of bindings that are either valid or invalid.

A function is at least to me somewhat unexpected for something called `preferred_binding`. That to me reads as a list, like `preferred_binding = ["PyQt4", "PySide"]`, similar to the environment variable. It could then optionally be a dictionary, for when you need per-library preferences, and we could check which it is via e.g. `if isinstance(preferred_binding, dict)`.

With that kind of use, we could potentially even expand on the environment varible, to be either a comma-separated list of bindings, or a JSON-encoded dictionary like the one above. Rez does this and it's visually ugly, but it works.

If we go for a function, how about `preferred_binding_callback` or `preferred_binding_function`?

MHendricks:
I've changed the code to support using a json dict for configuration. I need to work on the tests, but wanted to run it by you first.

This is not ready to merge because it is not backwards compatible and would require all vendored qt's to be updated to work. We could use a separate env var to support json data? Something like `QT_PREFERRED_JSON`, or `QT_PREFERRED_BINDING_JSON`. If the new env var is set its used and expected to be a json formatted string. Otherwise QT_PREFERRED_BINDING is used.

Would we want to make it so we can configure other info? I don't know if we should put too much info into an environment variable. I know in the past windows has had issues with really long environment variables. I've had to use the registry to work around the setx command's value limit https://superuser.com/questions/387619/overcoming-the-1024-character-limit-with-setx. We could make the env var support pointing to a file on disk, but that is probably overkill. I set QT_PREFERRED_BINDING in my startup scripts so I don't need setx, but I think it should be supported.

That being said, I would expect that only one or two vendor module paths will ever need modified. Most users are going to be using the expected PySide, PyQt for the application they need to support. In my case I should only ever need `{"Qt":["PyQt5","PyQt4"]}` or `{"blur.Qt":["PyQt5","PyQt4"]}` if I vendor the Qt we use in our code.

The original reason I went with a function was because I thought I was going to have to do some sort of inspect magic of the stack to figure out how what Qt module I would need, when I realized that `__name__` would work, I didn't change design.

Also, looks like I need to fix the tests, and I need to add a few more edge case tests I've thought of too.

mottosso:
> This is not ready to merge because it is not backwards compatible

Sorry, where does it break backwards compatibility? The check you've got for whether the variable is a JSON document or not seems to handle cases of backwards compatible values.

> and would require all vendored qt's to be updated to work. 

Hm, iIs there a way around that I wonder? Even with the prior approach, didn't we still need `studiolibrary` and friends to update their versions?

> We could use a separate env var to support json data?

That is an option, Rez does this too. I'm torn about whether it is the better option. On the one hand, it would simplify both the implementation and documentation. On the other it would introduce another environment variable. One the other other, it would make it clear which environment variables support JSON formatting.

If so, my preference is `QT_PREFERRED_BINDING_JSON` such that `_JSON` can be a consistent suffix to any other variables we choose to make this possible with.

> I know in the past windows has had issues with really long environment variables

That past is still with us, but in this case I think we're safe. It's problematic for 2,000 characters or more, which would have been relevant if we provided file paths rather than module paths I think.

And even still, the problem lies with cmd.exe, and today PowerShell is an option and supports far greater numbers, and finally adding through Python would support any amount, given it happens within the same process that accesses it.

Overall, I'm quite happy with this. What do you think?

The immediate issue would be how to handle the transition. It must be backwards compatible, that's for sure. But on having every instance of Qt.py make use of the feature; I suppose it could be up to vendored Qt.py's to upgrade if they have a specific need of only running on a subset of bindings, like studiolibrary. Those that aren't upgraded would simply not pay attention to our new environment variable. And worst case, a user could always set the `QT_PREFERRED_BINDING` prior to importing any third-party library.

MHendricks:
> Hm, iIs there a way around that I wonder? Even with the prior approach, didn't we still need `studiolibrary` and friends to update their versions?

The break in backwards compatibility is using `QT_PREFERRED_BINDING` to store json data. Older versions of Qt.py would not know that they need to parse json data and would raise the [`ImportError("No Qt binding were found.")`](https://github.com/mottosso/Qt.py/blob/master/Qt.py#L1822) error when they try to import the json string as a python module.

Using `QT_PREFERRED_BINDING_JSON ` to store the json data will be backwards compatible, as it will just be ignored by the older versions of Qt.py.

> The immediate issue would be how to handle the transition.

I'll update the merge request to use `QT_PREFERRED_BINDING_JSON ` if it is not a empty string, raising json errors if it can't parse the string. Otherwise it will respect `QT_PREFERRED_BINDING` as it currently does. This allows you to use both env variables if you have the need to customize the older vendored Qt.py modules. It allows the vendored Qt.py maintainers to upgrade Qt.py at their leisure. You can also still set `QT_PREFERRED_BINDING` prior to importing any third-party library.

I agree with your environment variable points.

mottosso:
This sounds good to me.

MHendricks:
Cool, the tests passed, I still need to update the readme info.

It looks like the example for ["Constrain available choices and order of discovery by supplying multiple values."](https://github.com/mottosso/Qt.py#override-preferred-choice) is referencing `PyQt` instead of `PyQt4` or `PyQt5`. Is this intentional, to show that it will fallback to `PySide`? 

##### Current
```bash
# Try PyQt first and then PySide, but nothing else.
$ export QT_PREFERRED_BINDING=PyQt:PySide
```

##### What I would expect:
```bash
# Try PyQt4 first and then PySide, but nothing else.
$ export QT_PREFERRED_BINDING=PyQt4:PySide
```

mottosso:
> Is this intentional, to show that it will fallback to PySide?

Looks like a typo, welcome to fix.

This looks good to me, just one thing stands out. The `json.loads` is able to throw an exeption, which means that `import Qt` is where that exception would get thrown. That's probably no good.

Could we add a try/except around that and print something loud when it fails? Ideally mentioning that it tried parsing that variable for JSON text but couldn't, such that someone who didn't set things up initially can figure out what's going on. Then we can fallback to default values if that happens.

MHendricks:
Heh, that took a lot of trial and error. My local testing didn't replicate the error so I had to test on Travis. Some day I may try to setup a proper local test system for stuff like this.

It now captures the json error and if using `QT_VERBOSE`, print `JSONDecodeError: ` and the error message I capture. This should give you the info to debug why your json is invalid. It will fallback to `QT_PREFERRED_BINDING` if invalid json is used. I think I have tests for all of this.

I think this is ready to merge unless I need to make any more changes. I hope the readme changes makes sense.

mottosso:
Looks good. The error message is a little hard to interpret unless you know where to look, but I'll take a stab at that in a separate PR.

Nice work @MHendricks!

PR #336, opened 2020-03-13
Update error message for failure to parse preferred JSON

mottosso:
Qt.py can print an error on import that looks like..

```cmd
> $ python -c "import Qt"
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

But it doesn't tell you why, and if you aren't familiar with the new `QT_PREFERRED_BINDING_JSON` variable then I expect we'll see issues pop up here.

I've changed that message to:

```cmd
> $ python -c "import Qt"
Qt.py [warning]: Failed to parse QT_PREFERRED_BINDING_JSON='blabla;bla'
Qt.py [warning]: Falling back to default preferred order
```

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -581,6 +581,7 @@ def test_vendoring():
     #
     # Test invalid json data
     #
+    env = os.environ.copy()
     env["QT_PREFERRED_BINDING_JSON"] = '{"Qt":["PyQt5","PyQt4"],}'
 
     cmd = "import myproject.vendor.Qt;"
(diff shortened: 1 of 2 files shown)


tests.py
```diff
@@ -578,72 +579,43 @@ def test_vendoring():
...
         assert popen.returncode == 0, msg
 
+    error_check = b"Qt.py [warning]:"
+    assert err.startswith(error_check), err
+
     print('out------------------')
     print(out)
 
     print('err ------------------')
     print(err)
```
MHendricks: Wow, I left a lot of debug prints from trying to debug my tests. These can probably be removed. I don't think they are useful anymore.

mottosso: It's fine. Let's blame it on corona.

Issue #348, opened 2020-09-04
wrapInstance: Search inheritance hierarchy for a suitable class


mitchellsimmons:
Hi, I would like to suggest a change to the implementation of QtCompat.wrapInstance.

## Issue

In the current implementation, wrapInstance only checks if the direct superclass is a QtWidgets class. Therefore if the class has a deeper inheritance hierarchy you end up getting a QObject.

In the below example, a pointer to an instance of class B will return a QObject instead of a QDialog.

```python
class A(QtWidgets.QDialog):
    pass
    
class B(A):
    pass
```

## Implementation

I've made changes for when `base is None`. Here the loop will eventually find QObject if nothing more significant is found. Alternatively you could use `while meta_object is not None` and add an `else` clause to check if the loop broke but I don't think it is necessary.

```python
def _wrapinstance(ptr, base=None):
    assert isinstance(ptr, long), "Argument 'ptr' must be of type <long>"
    assert (base is None) or issubclass(base, Qt.QtCore.QObject), (
        "Argument 'base' must be of type <QObject>")

    if Qt.IsPyQt4 or Qt.IsPyQt5:
        func = getattr(Qt, "_sip").wrapinstance
    elif Qt.IsPySide2:
        func = getattr(Qt, "_shiboken2").wrapInstance
    elif Qt.IsPySide:
        func = getattr(Qt, "_shiboken").wrapInstance
    else:
        raise AttributeError("'module' has no attribute 'wrapInstance'")

    if base is None:
        q_object = func(long(ptr), Qt.QtCore.QObject)
        meta_object = q_object.metaObject()

        while True:
            class_name = meta_object.className()

            try:
                base = getattr(Qt.QtWidgets, class_name)
            except AttributeError:
                try:
                    base = getattr(Qt.QtCore, class_name)
                except AttributeError:
                    meta_object = meta_object.superClass()
                    continue

            break

    return func(long(ptr), base)
```

mottosso:
Yes, this is fantastic. Could you transform this into a PR? That way we can get tests running and hunt down any potential issue. The idea is sound, and from a glance it looks solid.

mitchellsimmons:
Hey Marcus,
I've added the PR. I'm a bit new to this so let me know if I there's anything wrong.
Thanks

PR #349, opened 2020-09-04
wrapInstance: Search inheritance hierarchy for a suitable class

mitchellsimmons:
This is in reference to #348.

Adds support to QtCompat.wrapInstance for classes which have a deeper inheritance hierarchy.

## Example

Calling QtCompat.wrapInstance with a pointer to an instance of class B will now return a QDialog instead of a QObject.

```python
class A(QtWidgets.QDialog):
    pass
    
class B(A):
    pass
```

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -897,7 +897,7 @@ if sys.version_info < (3, 5):
     # https://github.com/PySide/shiboken-setup/issues/3
 
     def test_wrapInstance():
-        """Tests .wrapInstance cast of pointer to explicit class.
+        """Tests .wrapInstance cast of pointer to explicit class
 
         Note:
             sip.wrapInstance will ignore the explicit class if there is a more
@@ -933,11 +933,12 @@ if sys.version_info < (3, 5):
             app.exit()
 
     def test_implicit_wrapInstance_for_base_types():
-        """Tests implicit `Foo` -> `Foo` cast of `Foo` pointer to `Foo` object,
-        produced by .wrapInstance when:
+        """Tests .wrapInstance implicit cast of `Foo` pointer to `Foo` object
 
-        #. The `base` argument has a default value.
-        #. `Foo` is a standard Qt class.
+        Testing is based upon the following parameters:
+
+        1. The `base` argument has a default value.
+        2. `Foo` is a standard Qt class.
 
         """
         from Qt import QtCompat, QtWidgets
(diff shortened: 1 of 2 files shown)


mottosso:
I think this looks great, thanks @mitchellsimmons.

From the test suite, it looks like nothing existing broke. That's great. The remaining bit is converting your example into a test, [here](https://github.com/mottosso/Qt.py/blob/cfd1b1d3155485d7b855b9578fc847cf7f5a5153/tests.py#L899). Something that exercises this feature and covers as many corner cases as you can think of.

Up for it?

mitchellsimmons:
Thanks @mottosso.

I ran into an issue when trying to run the docker image.
The entrypoint.sh file appears to be using CRLF line endings.
I had to convert to LF to get the tests to run.

Once I've made my changes what would be the best way for me to post them here?

mitchellsimmons:
Sorry you can ignore the line endings issue, this was my fault. I'm using windows so I needed to update `git config --global core.autocrlf` to `false` before cloning the repository.  Maybe it could be useful to add a `.gitattributes` file to force LF line endings for the entrypoint.sh file.

mottosso:
> Once I've made my changes what would be the best way for me to post them here?

Whenever you push to your own repository, this PR will automatically update and re-run the tests with your latest changes. So go right ahead and push. And don't worry about pushing too often and creating too many commits. If it becomes a problem, we can "squash" all of your commits into a single one for a simplified Git history. That way you also don't necessarily need to run the tests on your own machine/Docker, although that is probably the faster option since it can take a minute or two for GitHub to get started.

mitchellsimmons:
Thanks for explaining this @mottosso.

Let me know what you think of the tests, I can make some changes if needed.

Regarding the minor change to `wrapInstance`  [here](https://github.com/mitchellsimmons/Qt.py/blob/9fdd65aec8efd9ef466e08f00e80d096c057e305/Qt.py#L782): after running my tests with sip I found that it always returns the most suitable class regardless of what the user provides so i decided it isn't necessary to run the search when sip is being used.

Regarding changes to testing [here](https://github.com/mitchellsimmons/Qt.py/blob/9fdd65aec8efd9ef466e08f00e80d096c057e305/tests.py#L308) and [here](https://github.com/mitchellsimmons/Qt.py/blob/9fdd65aec8efd9ef466e08f00e80d096c057e305/tests.py#L894): tests are not currently running for Python 3.4 because of the micro version.

tests.py
```diff
@@ -920,8 +932,14 @@ def test_wrapInstance():
         finally:
             app.exit()
 
-    def test_implicit_wrapInstance():
-        """.wrapInstance doesn't need the `base` argument"""
+    def test_implicit_wrapInstance_for_base_types():
+        """Tests implicit `Foo` -> `Foo` cast of `Foo` pointer to `Foo` object,
```
mottosso, line 936: This summary is a little long, and has a line-break.

The first line of the docstring is what shows up as [the name of the test](https://travis-ci.org/github/mottosso/Qt.py/jobs/724576499#L576) so we'd better keep it short and sweet. You can move the elaboration to the next line and beyond. Looks pretty good already perhaps, without the comma at the end there.

tests.py
```diff
@@ -920,8 +932,14 @@ def test_wrapInstance():
...
+        produced by .wrapInstance when:
+
+        #. The `base` argument has a default value.
```
mottosso, line 939: Are these bullet lists? Would rather make them `1.` and `2.` in that case to keep focus on the code rather than the formatting.

mottosso (changes requested):
Looks good to me, great work. Other than a few style-issues, this is good to merge.

tests.py
mitchellsimmons: Yep it's just rst auto-enumeration, I've changed it to explicit.
If any of the elaboration is too much I'm happy to remove it.
Cheers

mottosso: Ooo, learnt something new. Thanks.

mottosso (approved):

mottosso:
Thanks @mitchellsimmons, it's a wrap!

PR #350, opened 2020-10-01
Include the module name that failed to import in warning

MHendricks:
On windows, if a required dll is missing you get the exception message `DLL load failed: The specified module could not be found.`.
This makes means that you have to test each import manually so you can debug the problem.

This pull request adds the name of the module that failed to import to the warning.

Old: `Qt.py [warning]: ImportError: DLL load failed: The specified module could not be found.`
New: `Qt.py [warning]: ImportError(QtXmlPatterns): DLL load failed: The specified module could not be found.`

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -45,7 +45,7 @@ import importlib
 import json
 
 
-__version__ = "1.3.1"
+__version__ = "1.3.2"
 
 # Enable support for `from Qt import *`
 __all__ = []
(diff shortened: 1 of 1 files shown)


mottosso:
Nice one. :) Could you version this to 1.3.2 and I'll make a new release?

MHendricks:
Version updated

mottosso:
Released, thanks @MHendricks!

PR #352, opened 2020-11-18
PyQt enums patch

j0yu:
Patch `PyQt*` Qt namespace enums to behave like `PySide*`.

i.e. Fixes :negative_squared_cross_mark:  in `QtCore.Qt`
Attributes | `PyQt*` | `PySide*`
---|---|---
`Qt.Checked` value | :heavy_check_mark:  | :heavy_check_mark: 
`Qt.CheckState` class   | :heavy_check_mark:  | :heavy_check_mark: 
`Qt.CheckState.Checked` value | :negative_squared_cross_mark:  | :heavy_check_mark: 
`Qt.CheckState.values` dict | :negative_squared_cross_mark:  | :heavy_check_mark: 

**UPDATE** Clarification: Above is just 1 example, the intention is to patch for all flags/enums in the [Qt Namespace](https://doc.qt.io/qt-5/qt.html)

Would this fall under a similar re-name/re-expose `pyqtSignal` as `Signal`? Just curious if this will actually fall under the umbrella of [No wrappers](https://github.com/mottosso/Qt.py/blob/8a8953bc672b75a157a6c2a9b49e48985bb3e8a7/CONTRIBUTING.md#no-wrappers)

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1813,16 +1813,21 @@ def _pyqt_enums_patch():
         if not name.startswith("_")
     }
 
-    known_enums = set()
-    for enum_class in qt_attrs.values():
+    known_enums = {}
+    for class_name, enum_class in qt_attrs.items():
         if inspect.isclass(enum_class) and issubclass(enum_class, int):
-            known_enums.add(enum_class)
-            enum_class.values = {}
+            class EnumFlags(enum_class):
+                values = {}
+
+            known_enums[enum_class] = class_name
+            setattr(Qt.QtCompat, class_name, EnumFlags)
 
     for attr_name, attr in qt_attrs.items():
-        if isinstance(attr, int) and attr.__class__ in known_enums:
-            attr.__class__.values[attr_name] = attr
-            setattr(attr.__class__, attr_name, attr)
+        class_name = known_enums.get(attr.__class__)
+        if isinstance(attr, int) and class_name is not None:
+            enum_class = getattr(Qt.QtCompat, class_name)
+            enum_class.values[attr_name] = attr
+            setattr(enum_class, attr_name, attr)
 
 
 def _install():
--- a/Qt.py
+++ b/Qt.py
@@ -37,6 +37,7 @@ LICENSE
 
 """
 
+import inspect
 import os
 import sys
 import types
@@ -1813,7 +1814,7 @@ def _pyqt_enums_patch():
     }
 
     known_enums = set()
-    for attr_name, enum_class in qt_attrs.items():
+    for enum_class in qt_attrs.values():
         if inspect.isclass(enum_class) and issubclass(enum_class, int):
             known_enums.add(enum_class)
             enum_class.values = {}
--- a/Qt.py
+++ b/Qt.py
@@ -1800,6 +1800,30 @@ class MissingMember(object):
         raise NotImplementedError(self.__err)
 
 
+def _pyqt_enums_patch():
+    """Patch ``PyQt*`` Qt namespace enums to behave like ``PySide*``.
+
+    i.e. ``PyQt*`` only has Qt.Checked and Qt.CheckState, but not ``PySide*``
+    Qt.CheckState.Checked attribute or Qt.CheckState.values dictionary.
+    """
+    qt_attrs = {
+        name: getattr(Qt.QtCore.Qt, name)
+        for name in dir(Qt.QtCore.Qt)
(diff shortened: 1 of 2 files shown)


Qt.py
```diff
@@ -1800,6 +1800,30 @@ def __call__(self, *a, **kw):
...
+    """Patch ``PyQt*`` Qt namespace enums to behave like ``PySide*``.
+
+    i.e. ``PyQt*`` only has Qt.Checked and Qt.CheckState, but not ``PySide*``
+    Qt.CheckState.Checked attribute or Qt.CheckState.values dictionary.
+    """
+    qt_attrs = {
+        name: getattr(Qt.QtCore.Qt, name)
+        for name in dir(Qt.QtCore.Qt)
+        if not name.startswith("_")
+    }
```
mottosso, line 1814: This one for example could (and likely is) pick up enums other than the 2 being fixed by this PR.

tests.py
```diff
@@ -1159,3 +1159,44 @@ def test_multiple_preferred():
...
+        class_attributes = {  # Version agnostic flags and enums
+            "CheckState": ["Checked", "Unchecked", "PartiallyChecked"],
+            "Orientation": ["Horizontal", "Vertical"],
+        }
+        for class_name, attributes in class_attributes.items():
+            qt_class = getattr(QtCore.Qt, class_name)
+
+            values_path = "QtCore.Qt.{0}.{1}".format(class_name, "values")
+            assert hasattr(qt_class, "values"), (
+                "{0} should exist, but is missing".format(values_path))
```
mottosso, line 1176: This is very hard to read

mottosso (changes requested):
Thanks for this, good job with the little table there demonstrating the changes. Nice touch. 👍 

However, this is too magical/automatic with too much formatting going on to understand what's happening. See comments in code.

If the goal is adding these two members, it would be much preferred just adding them directly, although it can't modify the original class so it'll need to go in `QtCompat`

```py
# Bad
from PyQt4 import QtCore
QtCore.Qt.CheckState.Checked

# How did you get there?
# Ah.. someone has imported Qt.py somewhere
```

```py
# Good
from Qt import QtCompat
QtCompat.CheckState.Checked
```

Qt.py
j0yu: yeah, it was the original intention to be able to capture all cases found, not just `CheckState` only. The description in the PR could probably do with a better wording to highlight it.

j0yu:
> `Ah.. someone has imported Qt.py somewhere`
Great point, I think the `QtCompat` way is more tidy. But it will be great for the future if there is a wizardry to retain the `QtCore.Qt.CheckState.Checked` syntax without ruining the original `PyQt5.QtCore` module

j0yu:
0fa580a3d81f9ab14fbfa9034737721b5d83354e:  it's starting to gets messy imo, but I want to get an opinion on this.

## QtCompat

image

As per https://github.com/mottosso/Qt.py/pull/352#discussion_r526186885, the above implementation will work, but it does add the enums directly and if not careful, can conflict with existing attributes already inside `QtCompat`

## EnumFlags class

To replicate the `PySide*` behaviour in `QtCompat`, it requires a class `EnumFlags`. This might go against simplicity and is effectively a new _feature_ as it is a construct that's made in `Qt.py` to _mock_ the `PySide*` behaviour.

----

Bottom line, _it works_, but at what cost?

tests.py
j0yu: I switched to `%s` notations in newest commits

mottosso:
Thanks for the update, but this is still to magical. Or implicit, is the word.

Let's just make it real explicit instead, e.g. here:

https://github.com/mottosso/Qt.py/blob/8a8953bc672b75a157a6c2a9b49e48985bb3e8a7/Qt.py#L1547-L1553

One could put..

```py
CheckState = type("CheckState", (object,), {})
CheckState.Checked = Qt.Checked
CheckState.values = ...  # etc

QtCompat.CheckState = CheckState
QtCompat.Enum2 = Qt.Enum2
QtCompat.Enum3 = Qt.Enum3
...  # etc
```

And so forth. That way, we control which enums are actually patched, and also document what they are. That implicit method is not only hard to read, but doesn't tell you which enums are patched. It also runs the risk of patching different enums on different builds of PySide2/PyQt5/etc.

mottosso:
Also keep an eye on the tests, they're unhappy about something.

PR #353, opened 2020-11-20
Add QtCore.QStandardPaths for PySide2/Qt5

AdricEpic:
Discovered this module wasn't available in a PySide2 environment, and would like to have as part of the official release so we don't have to maintain an internally-modified version.


mottosso:
Sorry, this one is only available to Qt 5 and would break anything written with Qt.py that runs on Qt 4 such as PyQt4 and PySide(1).

If you only need PyQt5 and PySide2, I would recommend [Qt5.py](https://github.com/mottosso/Qt5.py)

- https://doc.qt.io/qt-5/qstandardpaths.html

image

PR #355, opened 2021-01-23
Fixed vendorised copies of Qt.py clashing

AWhetter:
We are using a tool that vendorises Qt.py (https://github.com/krathjen/studiolibrary). But two copies of Qt.py currently write to the same location in sys.modules, meaning that they can clash with each other.
This change follows other usage of sys.modules in Qt.py and writes to the module location of the current instance of Qt.py, rather than to the global location.


mottosso:
Nice catch, this looks solid. Thanks @AWhetter.

PR #358, opened 2021-07-13
Bugfix for #357

friedererdmann:
In #357 I proposed a solution to capture non-ascii characters in python 2 unicode strings before passing them to str().

An oversight meant that Python 2 was still unable to run the code. I propose the bugfix in this code and made the small test case to prove that the code now works in Python 2 and Python 3 and will go through all code tests.

```python
# coding=utf-8
# works in python 2 and python 3
import sys


def _warn_import_error(exc):
    if sys.version_info < (3, 0):
        if type(exc).__name__ == 'unicode':
            exc = exc.encode('ascii', 'replace')
    msg = str(exc)
    if "No module named" in msg:
        return
    _warn("ImportError(%s): %s" % (module, msg))


def _warn(text):
    sys.stderr.write("Qt.py [warning]: %s\n" % text)


message = u"DLL load failed : le module spécifié est introuvable."
_warn_import_error(message)
```

Bugfix for cast to ascii
Python logic: in previous version, the declaration of unicode = str in the Python 3 exclusive path made the Python 2 code fail, as unicode was now seen as an undeclared variable (as it had been declared in the Python 3 code, so Python assumes it is no longer the build-in type).

This version uses string comparison to work around the issue of the type.__name__, which seems to be the only way to work around this issue.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1260,18 +1260,13 @@ def _setup(module, extras):
     Qt.__binding__ = module.__name__
 
     def _warn_import_error(exc, module):
-        if sys.version_info > (3, 0):
-            unicode = str
-        else:
-            try:
-                if isinstance(exc, unicode):
-                    exc = exc.encode('ascii', 'replace')
-            except (UnboundLocalError, NameError):
-                pass
+        if sys.version_info < (3, 0):
+            if type(exc).__name__ == 'unicode':
+                exc = exc.encode('ascii', 'replace')
         msg = str(exc)
         if "No module named" in msg:
             return
-        _warn("ImportError(%s): %s" % (module, msg))
+        _warn("ImportError: %s" % msg)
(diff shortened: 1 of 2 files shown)


friedererdmann:
Apologies for the slip up in the previous Pull request, my test case was flawed.

mottosso:
No problem. Can you add the test to [`tests.py`](https://github.com/mottosso/Qt.py/blob/master/tests.py) too? That way it'll get run on all Python's and won't break again.

friedererdmann:
encountered one more hickup - please do not merge yet

friedererdmann:
My repro case is still failing on an actual French PC, so I need to re-work this. I'll re-post the PR later when I can figure out what Windows and 3Ds Max (our test case) are doing.

PR #359, opened 2021-07-16
Actual fix for #346 - unroll of #357

### LINKED ISSUE #346: Problem in _setup() on non-English OS
There's a problem in that code in 1.2.6:

```
 def _warn_import_error(exc):
        [...]
        _warn("ImportError: %s" % msg)
```
The _warn() call will throw a unicode-decode-error when run on a french OS, for instance:
```<type 'exceptions.UnicodeDecodeError'> 'ascii' codec can't decode byte 0xe9 in position 59: ordinal not in range(128)```

Possibly a missing .encode('utf-8') on msg?

friedererdmann:
After much back and forth trying to pin point the issue for one of our French speaking users, I got around getting a French Windows PC and rolled back through all the changes made to fix #346 .

This resets `_setup` and `_warn_import_errors` which I thought were the original culprits. However, the issue is specific to how 3Ds Max (and other Autodesk DCCs) use unicode strings in stdout.

This change moves catching the unicode error to `_warn` and encodes the message using the preferred locale, resulting in more readable code and a more readable error message.

I updated the test to test this without the need for a full setup.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1260,18 +1260,13 @@ def _setup(module, extras):
     Qt.__binding__ = module.__name__
 
     def _warn_import_error(exc, module):
-        if sys.version_info > (3, 0):
-            unicode = str
-        else:
-            try:
-                if isinstance(exc, unicode):
-                    exc = exc.encode('ascii', 'replace')
-            except (UnboundLocalError, NameError):
-                pass
+        if sys.version_info < (3, 0):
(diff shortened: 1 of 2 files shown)


mottosso:
Brilliant, great work @friedererdmann. Happy for me to merge and release this?

friedererdmann:
Yes, this should be safe to release

PR #364, opened 2021-09-13
Update Qt.py

mangopipeline:
adding support for QtGui.QGuiApplication


mangopipeline:
bummer I see the ci failed because the member is not in Qt4.
feel free to ignore this :(

mottosso:
Indeed, such is life. :( Thanks for the effort though!

Issue #369, opened 2022-05-24
QtCompat.isValid() Not Available for All Qt Objects


andryjong:
`QtCompat.isValid()` starts with the following `assert` statement:
https://github.com/mottosso/Qt.py/blob/242ec430ff390cf24051819802b9eec1e665d3b2/Qt.py#L816

However, not all objects in Qt inherits from `QObject` (e.g. `QGraphicsItem`, `QListWidgetItem`, `QTreeWidgetItem`, etc.). 
This `assert` statement makes `QtCompat.isValid()` not available for such objects.

How can I best check for object validity in this case?

andryjong:
@mottosso -- Unfortunately, just like `QObject`, we cannot guarantee that an item is still in memory.

Casting `QGraphicsItem` into `QGraphicsObject` would work if the item itself is still valid.
Otherwise, we'll get into the same situation where it would say something like "Internal C++ object already deleted". 😭 

I've gotten around the `AssertionError` by having all of my items also inherit from `QObject`, like follows:
```python
class GroupItem(QtWidgets.QGraphicsItemGroup, QtCore.QObject)
```
... but this feels rather like a "hacky" way to get around the `assert` statement.

mottosso:
Oh, is the problem only that there is an assert in Qt.py? Then it is surely safe to remove. I can't remember why this was put there. What kind of error should we expect shiboken to throw if an object does not support it? Should probably be fine? Something the user could catch?

If that's the only problem, then a PR removing this assert is more than OK.

andryjong:
@mottosso -- Is there a time frame and/or process to get the latest version to PyPI?

mottosso:
Sorry, I'm working on it. It's supposed to be automatic, but the build chain (Travis, or PyPI) has changed, apparently. If it doesn't work in the next 20 mins, then it'll be tomorrow.

andryjong:
Understood. Thank you!! 🙏

mottosso:
All done! 🥳

andryjong:
Thank you so much, sir!

PR #370, opened 2022-05-26
Remove assert Statement from QtCompat.isValid()

andryjong:
In reference to issue #369: 

- Remove `assert` statement from `QtCompat.isValid()` to ensure the function is available for any objects (not just ones that inherit from `QObject`).
- Update the `test_isValid()` unittest.


mottosso:
Fantastic, thanks @andryjong

Issue #395, opened 2024-02-25
The imp module is removed with Python 3.12


dvzrv:
Hi! :wave: 

We're currently cleaning up packages for the Python 3.12 rebuild on Arch Linux.

I noticed that the tests in this repository are still using the imp module
https://github.com/mottosso/Qt.py/blob/2c0f4dc596204670f4597b841821cff8a171c3a2/tests.py#L6

This module has been removed with Python 3.12:
https://docs.python.org/dev/whatsnew/3.12.html#imp

MHendricks:
This is being addressed by #463. I've added testing of Python 3.12, 3.13 and 3.14.

Yeah, it's being maintained but a we often have to wait years for the programs we use this in to get newer versions of python or Qt. When I was working on the 2.0 refactor recently the newer versions were not working and I didn't have time to debug why. Though now that I have imp seems to be the last issue we had.

PR #396, opened 2024-02-25
Use importlib on Python > 3.11 in tests

dvzrv:
As the imp module is removed with Python 3.12, use importlib instead.

https://docs.python.org/dev/whatsnew/3.12.html#imp

Fixes: https://github.com/mottosso/Qt.py/issues/395

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -3,7 +3,6 @@
 import io
 import os
 import sys
-import imp
 import shutil
 import tempfile
 import subprocess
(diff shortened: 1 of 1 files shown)


dvzrv:
Sorry, but is this a joke? A CLA?

dvzrv:
Feel free to reuse this code, but this is unfortunately not something I will continue contributing to further. :wave:

Issue #401, opened 2024-03-12
headerToModule() is destructive when CustomWidget path is a Python . path


jasonbrackman:
In the following function:

```
def headerToModule(header):
                    """
                    Translate a header file to python module path
                    foo/bar.h => foo.bar
                    """
                    # Remove header extension
                    module = os.path.splitext(header)[0]

```

This will break imports if they were written in Python as the structure of the custom widget is "path.to.module"

Since the function always expects a path and a .h extension -- the above is destructive and removes the `.module` in my example above.  Custom widgets are not possible if using Python to author the custom widget with Qt.py

Possible Fix?
Patch the Qt.py: 
    module = os.path.splitext(header)[0] if header.endswith(".h") else header

The issue here is that if its a python module you simply can't use '.h' as a module name.  Its a `gotcha` with the above.

mottosso:
Happy with this change, so long as there's a test for these two cases in https://github.com/mottosso/Qt.py/blob/master/tests.py

jasonbrackman:
The following test cases are what I believe are appropriate inputs/results based on current code expectation and also addressed the bug:

```
    path_tests = {
        # Input: Expected
        
        # Valid paths; .h paths need to be updated to work with Qt.py, Python . paths should be untouched.
        "path.to.module": "path.to.module",
        "path\\to\\module.h": "path.to.module",
        "path/to/module.h": "path.to.module",
        "module.h": "module",
        "module": "module",

        # malformed; leaving the bad input QtDesigner input.
        "path.to.module.py": "path.to.module.py",   # unexpected .py
        "path\\to\\module.py": "path\\to\\module.py",  # python paths are always . paths.
        "path/to/module.py": "path/to/module.py",   # ^
        "path\\to\\module": "path\\to\\module",   #  This is malformed.  It should either be a .path for Python or have a .h ext.
        "path/to/module": "path/to/module",  # ^
        "module.py": "module.py",  # Python modules should have have a .py extension
        }
```

Should Qt.py attempt to correct malformed input?  If the end user entered the malformed data, I'd expect it to fail in PyQt -- so expecting that Qt.py would behave the same?  I'm surprised that Qt.py has to do any .h manipulation as well -- but I don't have an example to test so have kept the 'functionality'.

mottosso:
[...]

PR #402, opened 2024-03-13
Updating the `headerToModule` def to only manipulate '.h' files.

jasonbrackman:
https://github.com/mottosso/Qt.py/issues/401

This is an attempt to have the least amount of impact on the existing functionality, and allow Python custom widgets paths to work.   

The original `header` authored by the user will always be attempted first.  If it fails to import for the PySide implementation, a second attempt is made using the original Qt.py `headerToModule` function.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -930,6 +930,11 @@ def _loadUi(uifile, baseinstance=None):
                     Translate a header file to python module path
                     foo/bar.h => foo.bar
                     """
+
+                    if header.endswith(".h") is False:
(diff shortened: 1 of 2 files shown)


jasonbrackman:
Will be update after installing docker / testing locally.

mottosso:
Looks good to me. Let me know when you're happy and we can merge this.

jasonbrackman:
I won't make any other changes if all is good.  Thanks!

PR #403, opened 2024-05-02
Update Qt.py

mottosso:
Updating README, tests and version

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -47,7 +47,7 @@ import importlib
 import json
 
 
-__version__ = "1.3.10"
+__version__ = "1.4.0"
 
 # Enable support for `from Qt import *`
 __all__ = []


mottosso:
Tests for the PySide6 transition is not working. 😭 Help needed, we need to ensure that code written for PySide6 also runs on PySide2 and vice versa, and these two points are the start of that:

- https://github.com/mottosso/Qt.py/pull/403/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R605

mottosso:
These tests should work for every binding.

```py
def test_midbutton_qt6():
    """QtCore.MidButton was renamed QtCore.MiddleButton in Qt 6"""
    from Qt import QtCore, QtCompat

    if binding("PySide6"):
        assert QtCompat.Qt.MidButton == QtCore.Qt.MiddleButton
    else:
        assert QtCompat.Qt.MidButton == QtCore.Qt.MidButton


def test_set_font_weight():
    """Qt 6 changed font weights from integers to enums"""
    from Qt import QtGui, QtCompat
    font = QtGui.QFont()
    QtCompat.QFont.setWeight(font, 400)
```

PR #405, opened 2024-05-09
Updated _misplaced_members for PySide6

johnsonthomassonttf:
Long time listener, first time caller from [The Third Floor](https://thethirdfloorinc.com/)

I located these 4 members in PySide6 and based this modification on similar commits. Please let me know if there's a better way to do this. We're actively testing in PySide6 and PySide2, so I may discover more misplaced members.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -931,6 +931,11 @@ NOTE: For bindings where a member is not replaced, they still
 """
 _misplaced_members = {
     "PySide6": {
+        "QtGui.QUndoCommand": "QtWidgets.QUndoCommand",
+        "QtGui.QUndoGroup": "QtWidgets.QUndoGroup",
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks! Yes, this is the way, except we also need them for PyQt4 and PyQt5. It's hard to miss right now.. And tests won't pick this up.. But if they aren't there, then they simply would not get added to Qt.py when using those.

Could you add these to PyQt4 and 5 as well?

johnsonthomassonttf:
Added. Also added to PySide. I couldn't find API docs for PyQt4, but I believe it should be a match for PySide 1, so I referenced those API docs. Interesting that they flip-flopped on the location of these members.

mottosso:
Looks good, thanks @johnsonthomassonttf!

PR #406, opened 2024-05-10
add QProxyStyle

robertjoosten:
Previously unable to access QProxyStyle as it was not mapped.

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -482,6 +482,7 @@ _common_members = {
         "QPlainTextEdit",
         "QProgressBar",
         "QProgressDialog",
+        "QProxyStyle",
         "QPushButton",
         "QRadioButton",
         "QRubberBand",


robertjoosten:
i needed to test to see if it worked for the other bindings as well
which I see it doesn't. Does this go against the commonality of all Qt versions or is there a way to slot it in?

mottosso:
> Does this go against the commonality of all Qt versions or is there a way to slot it in?

Yes unfortunately this won't work. It would mean you could write a UI with Qt.py, and sometimes it would work, sometimes it would not. Based on what bindings the user would have available. It's important that if it works at all, it should work on all bindings, and PySide2 is still the source of truth for what exists and where.

mottosso:
For completeness, the recommended workflow for when you have binding-specific requirements is to explicitly include these, e.g.

```py
import Qt

try:
   from PySide6.QtCore import QProxyStyle
   CAN_PROXY_STYLE = True
except ImportError:
   CAN_PROXY_STYLE = False
```

And thus you can design your code to take this into account.

PR #408, opened 2024-05-15
Pyqt5 more misplaced

tbttfox:
Sorry for the PR twofer. I got pulled away before I could update my last one with this fix.

Looks like 80dc006094fab4e4a950b2cf0fbb74ad130d5dbd removed a chunk of misplaced members that shouldn't have been removed. Possibly because the key equals the value? In the other backends, it seems they got re-added, but not for PyQt5. I've added those back in.

Also `QItemSelection` and `QItemSelectionRange` are handled by the `_misplaced_members` mechanism, so they shouldn't be in `_common_members`

Finally, there were two remaps of QStringListModel in the PySide6 section, which look like they were copied from the PySide2 section.  PySide6 doesn't have the issue that required the double remap, so I removed the unnecessary one.


mottosso:
Thanks for this. Looks like many things vanished in the upgrade to PySide6, good that they are still here.

PR #409, opened 2024-05-16
Update reference to release 1.4.1 in README.md

pylipp:


mottosso:
Thanks, better late than never. :)

PR #411, opened 2024-05-24
Support "QtSvgWidgets.QSvgWidget"

ababak:


MHendricks:
Thanks for the pull request. Sorry for the delay, I'm working my way through the backlog now that I'm starting to maintain Qt.py.

It looks like you moved the binding to `QtSvgWidgets` which is correct for PySide6 but for the current [Qt4, Qt5 and Qt6](https://github.com/mottosso/Qt.py?tab=readme-ov-file#subset-or-common-members) support design of Qt.py it probably should be put on `QtSvg` instead.

> Members of Qt.py is a subset of PySide2. Which means for a member to be made accessible via Qt.py, it will need to (1) be accessible via PySide2 and (2) each of the other supported bindings.

However once we drop Qt4 support we may change the reference from PySide2 to PySide6, which would mean the correct location is back to `QtSvgWidgets`. https://github.com/mottosso/Qt.py/issues/413#issuecomment-2810279903

I think for now we should add it back to `QtSvg`, but depending on the responses to https://github.com/mottosso/Qt.py/issues/413#issuecomment-2810279903 it would eventually be added to the new `QtSvgWidgets`.

Also, it looks like we lost another svg member `QGraphicsSvgItem`. It was also moved into `QtSvgWidgets` and should probably be restored.

MHendricks:
I'm going to close this pull request and address the issue with #434 once #431 is merged.

PR #419, opened 2024-09-08
little fix QT 6 Transition documentation

miquelcampos:
Just a little fix on the readme documentation :)

| `QtCore.Qt.MidButton`  | `QtCompat.QtCore.Qt.MidButton` >> | `QtCore.Qt.MidButton`  | `QtCompat.Qt.MidButton`


mottosso:
Good catch, thanks :)

PR #428, opened 2025-02-03
Add some missing stubs to QtCompat

jdrese:


mottosso:
Thanks!

nrusch:
Thanks for this update. We just ran into a situation where we need to `# type: ignore` our way around the missing symbols in the `QtCompat` stubs, so I'm wondering if it might be possible to get a new point release pushed that incorporates these changes.

mottosso:
I've pushed 1.4.2 now.

nrusch:
Fantastic, thank you!

PR #437, opened 2025-05-09
convert function handle files generated using uic 5.15.2

stevenjking:
This updates the convert function to work with files generated by pyside2-uic version 5.15.2


MHendricks (approved):
Thanks, this looks good, I'll plan to release this on Monday.

When I discovered there was a Qt.py cli recently I wondered why it didn't seem to work but I always just directly load the .ui files so I didn't take the time to look into it.

PR #443, opened 2025-10-09
Add __binding__ to type stubs

### LINKED ISSUE #442: __binding__ missing from Qt-stubs
I'm typing checking the following code with mypy

```python
import Qt
Qt.__binding__
```

mypy raises the following error: `Module has no attribute "__binding__"  [attr-defined]`
It looks like `__binding__` is missing from Qt-stubs.

AWhetter:
Fixes #442

I could have used `Literal` here to specify exactly which strings are valid, but I couldn't think of a good use case for the extra complexity. So I've kept is simple and defined it as `str`.

### DIFF
--- a/Qt-stubs/__init__.pyi
+++ b/Qt-stubs/__init__.pyi
@@ -1,4 +1,4 @@
-
+__binding__: str
 __qt_version__: str
 IsPyQt5: bool
 IsPyQt4: bool


MHendricks:
Thanks, I haven't really adopted using type stubs beyond having mypy installed in my IDE. Mypy was complaining about this one. I'm going to add the new isPy*6 as well.

Issue #413, opened 2024-05-27
Should we remove support for PySide/PyQt4?


mottosso:
Question for the community and users:

With the introduction of PySide6, should we remove support for PySide and/or PyQt4? If so, we could start to include members of Qt.py that exist in PySide2/PySide6, but not PySide/PyQt4, [such as QGuiApplication](https://github.com/mottosso/Qt.py/pull/412).

Vote with a 👍  for "Yes" and 👎 for "No".

dekekincaid:
Yes, major version up and drop qt4 support.  If people really still want qt4 support they just use v1.x.

MHendricks:
The dropping of classes like `QTextCodec` in 1.4 has already caused issues for us and I've had to resort to using the pip requirement of `Qt.py==1.3.*`.

mottosso:
Looks like QTextCodec is not included in Qt 6 anyway, so even without PySide/PyQt4 this would still be missing.

- https://doc.qt.io/qt-6/qtextcodec.html

matthewlow-dwa:
For potential guidance, the VFX Reference Platform recommends that:
[https://vfxplatform.com/](https://vfxplatform.com/#:~:text=Providers%20of%20software%20libraries%20focused%20on%20VFX%20and%20animation%20content%20creation%20should%20aim%20to%20support%20their%20releases%20for%20the%20current%20calendar%20year%20and%20the%20three%20preceding%20years%20with%20compatible%20updates.%20Studios%20and%20end%20users%20should%20have%20no%20expectation%20of%20support%20for%20older%20libraries.)
> Providers of software libraries focused on VFX and animation content creation should aim to support their releases for the current calendar year and the three preceding years with compatible updates. Studios and end users should have no expectation of support for older libraries.

One way to interpret that recommendation is that Qt4 hasn't been supported since VFX Platform CY2015, well outside the current + 3 year window.

dekekincaid:
Many libraries way outside the scope of vfx use Qt.py. The vfx platform has minimal to do with this project so it needs a longer tail than 3 years.  That being said Centos/RHEL 7 is probably the last Linux distribution that still supports/ships with qt4. It is EOL at the end of June and it had a 10 year support period.

mottosso:
Looks like this is the way to go. If anyone is up for submitting a PR with PySide and PyQt4 support dropped that would get the ball rolling. It'd consist of two primary steps.

1. Remove all PySide/PyQt4 code
[...]

PR #446, opened 2025-12-12
Adding QWindow

Jerakin:
Imports fails.

`from Qt.QtGui import QWindow`

Adding `QtGui.QWindow` to _misplaced_members. _common_members will not work due to QWindow not being in Qt4

### DIFF
--- a/Qt.py
+++ b/Qt.py
@@ -1079,6 +1079,7 @@ NOTE: For bindings where a member is not replaced, they still
 """
 _misplaced_members = {
     "PySide6": {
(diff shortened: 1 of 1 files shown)


MHendricks:
This is being addressed by the v2.0 releases that are in progress. It doesn't belong in the v1.4 releases as its not available in [Qt4(PySide/PyQt4)](https://github.com/mottosso/Qt.py/wiki/Membership-between-Qt4,Qt5,Qt6).

See https://github.com/mottosso/Qt.py/issues/413#issuecomment-3711871613 for the current progress on it.

PR #466, opened 2026-06-01
Extend QtCompat to handle pos method updates similar to QMouseEvent

MHendricks:
Adds QtCompat class bindings for position method changes similar to the bindings already defined for `QMouseEvent`.

`QDragEnterEvent`, `QDragMoveEvent`, `QDropEvent`, `QEnterEvent`, `QHoverEvent`, `QMouseEvent`, `QNativeGestureEvent`, `QTabletEvent`, `QWheelEvent`

Also adds testing to ensure that all bindings in `_compatibility_members` have consistent QtCompat class definitions. This removes the PyQt6 specific `QtCompat.Qt.WindowMinMaxButtonsHint` binding that I should not have added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -630,27 +630,70 @@ To make a new release onto PyPI, you'll need to have the correct permissions and
 
 ### Qt 6 Transition Guide
 
+<!-- Tooltips used by the table -->
+[tt-event-pos]: ## "In Qt6 many but not all classes had their pos methods re-worked and simplified. If you are working with a instance of the a given class replace the old call with a QtCompat replacement class. The `.toPoint()` calls are only needed if you want to preserve the older int type instead of a float."
+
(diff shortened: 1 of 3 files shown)


MHendricks:
Code used to find these items.

```py
import Qt
import re

print(Qt.__binding__)
for cn in dir(Qt.QtGui):
    if not re.match("Q\w+Event", cn):
        continue
    found = False
    for prop in ["globalPos", "globalPosF", "globalX", "globalY", "localPos", "pos", "posF", "screenPos", "windowPos", "x", "y", "globalPosition", "position", "scenePosition"]:
        if hasattr(getattr(Qt.QtGui, cn), prop):
            if not found:
                found = True
                print(f"# {cn}")
            print(f"    {cn}.{prop}")
```

